Issue #1, opened 2019-10-23
HEC Port is not exposed by the operator


matthewmodestino:
In testing deployments with Splunk Operator, it appears the default HEC port `8088` is not exposed by the Splunk containers. 

Currently, I am working around this by patching the StatefulSet that deployed, which technically, the Operator should likely detect and reconcile or revert. Will open another issue on that if required, as it may be a indicator of some work to be done on reconciliation.

mikedickey:
https://github.com/splunk/splunk-operator/pull/2

PR #2, opened 2019-10-23
Expose port 8088 on indexers for HEC

mikedickey:
Expose port 8088 on indexers for HEC

Also, only expose ports for DFC (DFS component) on search heads

### DIFF
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -270,7 +270,7 @@ func GetSplunkStatefulSet(cr *v1alpha1.SplunkEnterprise, instanceType SplunkInst
 							Image:           GetSplunkImage(cr),
 							ImagePullPolicy: corev1.PullPolicy(cr.Spec.ImagePullPolicy),
 							Name:            "splunk",
-							Ports:           getSplunkContainerPorts(),
+							Ports:           getSplunkContainerPorts(instanceType),
 							Env:             envVariables,
 							VolumeMounts:    getSplunkVolumeMounts(),
 						},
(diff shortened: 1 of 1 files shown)


mikedickey:
I tested this by adding a few defaults parameters in a SplunkEnterprise deployment:

```defaults: |-
    splunk:
      hec_disabled: 0
      hec_enableSSL: 0
      hec_token: "abcd-1234-efgh-5678"
```

Note that you can expose HEC by linking ingress rules to port 8088 of the splunk-*-indexer-headless service:

```
- host: splunk-hec.eksdemo.splenti.io
    http:
      paths:
      - backend:
          serviceName: splunk-dfscluster-indexer-headless
          servicePort: 8088
```

```
curl https://.../services/collector -H "Authorization: Splunk abcd-1234-efgh-5678" -d '{"event": "Hello, world!", "sourcetype": "manual"}'
{"text":"Success","code":0}%
```

mikedickey:
Fixes #1

aegershman:
This might be an obvious answer, but in the Ingress example...

```yml
- host: ...
    http:
      paths:
      - backend:
          serviceName: splunk-dfscluster-indexer-headless
          servicePort: 8088
```

Would the actual value of the `path` be `/services/collector`? e.g.,

```yml
- host: ...
    http:
      paths:
      - backend:
          serviceName: splunk-dfscluster-indexer-headless
          servicePort: 8088
        path: /services/collector
```

mikedickey:
@aegershman For the example, adding `path: /services/collector` would be optional and I believe it would just eliminate having to include that path in your HTTP requests to the endpoint. Not sure if there's any pros/cons to being able to hit other endpoints on 8088.

Yes, I think it's best to use ingress to terminate TLS because of all the benefits one can get out of cert-manager. I imagine you could also encrypt the channel between the nginx proxies and the indexer pods by setting hec_enableSSL to 1.

PR #5, opened 2019-10-31
Various code cleanup using go fmt and lint tools

mikedickey:
Refactored StatefulSet comparison code to fix a bug causing spark-master deployment to always be updated even when there were no material changes.

Added new Makefile targets for "make lint" and "make fmt" that run "golint" and "go fmt" on all *.go source files.

Added "make local" target to just build a local binary without building the splunk/splunk-operator container image.

Renamed "make splunk-operator" target to "make image" for clarity

Added a "make clean" target that just prunes the binary output directory and removes the splunk/splunk-operator image

Removed SplunkServiceType since it just made code more complicated

Lots of various updates for formatting and to fix lint warnings.

Updated all copyright headers to match Apache license style

Added Frequently Asked Questions document

Added version number in startup log


pkg/apis/enterprise/v1alpha1/splunkenterprise_types.go
```diff
@@ -14,20 +24,22 @@ import (
 // NOTE: Do not make struct fields private (i.e. make sure all struct fields start with an uppercase) otherwise the
 //		 serializer will not be able to access those fields and they will be left blank
 
+// SplunkResourcesSpec defines the resource requirements for a SplunkEnterprise deployment
 type SplunkResourcesSpec struct {
-	SplunkCpuRequest     string `json:"splunkCpuRequest"`
-	SparkCpuRequest      string `json:"sparkCpuRequest"`
+	SplunkCPURequest     string `json:"splunkCpuRequest"`
```
nwang92: This is definitely a future-improvement thing, but one thing I've seen as a standard on most helm charts is a completely empty block for `resources` which users can add their own map of limits/requests as you would see on a regular `Pod` spec - might be something we want to consider instead of managing 4 variables like this. Example: https://github.com/helm/charts/blob/master/stable/prometheus/values.yaml#L239

I know this isn't helm, but it may be a more flexible system - especially when future requests come in for tolerations/nodeSelector/affinity/annotations/etc.

mikedickey: Yes, plan is to ultimately break up this CRD into more specific component CRDs and re-work all these resource configs at that time.

Issue #6, opened 2019-11-01
Operator should support the automatic creation of at minimum one HEC token and configuration of HEC


rfaircloth-splunk:
Operator should support the configuration and storage of the first HEC token as a secret similar to the default admin password this is valuable for automation of splunk connect for kubernettes

mikedickey:
It does automatically generate a `hec_token` secret and plug it into config, but not sure if the format is ideal.

```
$ kubectl get secret splunk-dfscluster-secrets -o jsonpath='{.data.hec_token}' | base64 --decode
~j?vX+9szdgU)2S?hL~lZZ^M%
```

mikedickey:
v0.0.6 uses UUID style secret for hec_token

PR #19, opened 2019-12-03
Various fixes

mikedickey:
Fixed deadlock condition between deployer and search heads when apps_location is specified (note this this also requires a patch in splunk-ansible splunk/splunk-ansible#312

Updated auto-generated secrets to use alpha-numeric characters only to prevent escaping errors. This was after running into yet more circumstances where ansible plays would fail due to random bytes selected.

Updated auto-generated hec_token secret to use UUID like format

Updates to ensure HEC is always enabled by default

Updates to Ingress examples to include S2S and consolidate HEC

Added examples for app installation and custom config

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,11 +4,11 @@
 
 all: image
 
-image: deploy/all-in-one.yaml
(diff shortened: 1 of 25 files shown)


mikedickey:
Fixes #6

PR #25, opened 2020-01-24
Added splunk.pass4SymmKey to default secrets

### LINKED ISSUE #10: pass4symkey error reported in ui
pass4SymmKey setting in the clustering or general stanza of server.conf is set to empty or the default value. You must change it to a different value.

mikedickey:
Fixes #10 when combined with a splunk/splunk image that includes this splunk-ansible fix:
https://github.com/splunk/splunk-ansible/pull/356

Should work ok for splunk-operator > 0.0.6 and splunk/splunk > 8.0.1

### DIFF
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -400,10 +400,11 @@ func GetSplunkDefaults(cr *v1alpha1.SplunkEnterprise) *corev1.ConfigMap {
 func GetSplunkSecrets(cr *v1alpha1.SplunkEnterprise) *corev1.Secret {
 	// generate some default secret values to share across the cluster
 	secretData := map[string][]byte{
-		"hec_token":  generateHECToken(),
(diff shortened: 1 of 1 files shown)


pkg/splunk/enterprise/configuration.go
```diff
@@ -411,13 +412,15 @@ splunk:
     hec_enableSSL: 0
     hec_token: "%s"
     password: "%s"
+    pass4SymmKey: "%s"
     idc:
         secret: "%s"
```
nwang92: I think idc needs this too?

mikedickey: I was leaving it as `splunk.idc.secret` so that it works with both old and new containers (backwards compat).

mikedickey:
Just for the record, this also changed `splunk.idc` to `splunk.idxc` -- this was a long-standing bug that only recently was exposed via other changes made in splunk-ansible repo.

PR #28, opened 2020-02-04
Do not require licenseURL for clusters

mikedickey:
Previously, the operator only allowed creation of clusters if a licenseURL was provided.

A license is not actually required to create clusters. Without providing a license, a time-limited trial will be created. This is especially desirable for testing.

This changes the behavior so that clusters may be created without a license. When a licenseURL is not provided, the license master will not be created and the pods in the deployment will not be configured to use a license master.


rfaircloth-splunk:
Thank you!

Issue #31, opened 2020-02-10
Allow Services to be created using a Template


kurktchiev:
I am running into a scenario which requires me to modify the way the services get deployed for a cluster. In my case I would like to set all the services to be headless, but there are usecases where you want those services to use LoadBalancer or NodePort, depending on the usage.

mikedickey:
Merged to develop for inclusion in 0.1.0

Issue #14, opened 2019-11-05
enterprise.splunk.com/v1alpha1 resources spec is too generic


nickperry:
```apiVersion: enterprise.splunk.com/v1alpha1
kind: SplunkEnterprise
metadata:
  name: example
spec:
  resources:
    ...
```

This spec is mostly generic and not per instance type. I need different sizing for indexers, search-heads, spark workers and ICMs.

For example, I would like

```splunkCpuRequest```

to be split up into ```
indexerCpuRequest
searchHeadCpuRequest```
etc.

mikedickey:
Planning to address this as part of #17

mikedickey:
Merged to develop for inclusion in 0.1.0. See new v1alpha2 CRDs.

Issue #32, opened 2020-02-11
Non ASCII character


kurktchiev:
Seems that the `,` in the Dockerfile label ends up getting reported as non ASCII.

https://github.com/splunk/splunk-operator/blob/eace63e14b84c3926d8450f8dc0a6c123bce6ced/build/Dockerfile#L13

kalinon:
This is present in the following image:
```
docker.io/splunk/splunk-operator                                 latest                       fe27e19956d8
```

You can verify with:
```
docker inspect docker.io/splunk/splunk-operator | LC_ALL=C grep --color='auto' -P -n "[\x80-\xFF]"
```
![Screen Shot 2020-02-11 at 11 27 00 AM](https://user-images.githubusercontent.com/5629164/74256494-6d0a1d80-4cc1-11ea-830a-c05cada88bd5.png)

mikedickey:
Merged to develop for inclusion in 0.1.0

Issue #17, opened 2019-11-05
Separate SplunkEnterprise into multiple CRDs


mikedickey:
**CSPL-37**

Currently, there is only a single SplunkEnterprise CRD and this is used to run Splunk Enterprise, DFS components, and Spark cluster all within a K8s environment.

We would like to separate this out into multiple, more component-focused CRDs for a few reasons:

**Customers have expressed a desire to use the operator to deploy components that work with existing Splunk Enterprise deployments**

For example, most customers already have existing indexers or indexer clusters and would not want to migrate these to Kubernetes on the first pass, or possibly not for a long time. But they may want to experiment with running individual search heads or search head clusters on Kubernetes that use their existing indexers as peers. For this, we may want to have a SplunkSearchHead and SplunkSearchHeadCluster CRD that can reference external indexers.

**Customers would like to use the operator to deploy and manage new components that augment existing Splunk Enterprise deployments**

For example, an existing customer may want to deploy DFS, DSP and/or MLTK TensorFlow clusters using Kubernetes that work in conjunction with their existing Splunk Enterprise deployments. For this, they may want to deploy a DFS-enabled SplunkSearchHead or SplunkSearchHeadClusters, or potentially just deploy the spark cluster as SplunkDFSWorkers? Or they may just want to create a SplunkDSP object that they can use to manage ingest into existing Splunk Enterprise deployments.

**Customers would like to use the operator to deploy more flexible validated architectures**

For example, a customer may want to deploy both a single standalone SplunkSearchHead with ITSI installed that is separate from a SplunkSearchHeadCluster that is used for more general investigation purposes.

**Customers would like to provide more specific resource limits (CPU/memory/storage) for different components**

For example, a customer may want to have use different StorageClasses and different CPU and memory requests for search heads and indexers.

The scope of this epic is just to break down & refactor the existing functionality provided by the SplunkEnterprise CRD into:

-     SplunkSearchHead (standalone)
-     SplunkSearchHeadCluster
-     SplunkIndexer
-     SplunkIndexerCluster
-     SplunkDFSWorkers (not sure about this name; standalone spark cluster that could be used by various existing search heads/clusters)

Note that each of these should allow references either to other Kubernetes objects (e.g. a SplunkSearchHead referencing a SplunkIndexerCluster by id) or by hostname/credentials for reference to external components not hosted by the splunk-operator.

[...]

PR #33, opened 2020-02-13
Break down crds

mikedickey:
Added the following new CRDs to enterprise.splunk.com/v1alpha2, which provide all functionality of SplunkEnterprise but with more flexibility:

    * Standalone
    * LicenseMaster
    * Indexer
    * SearchHead
    * Spark

SplunkEnterprise is now just a convenience wrapper, that manages the other v1alph2 CRDs

Lots of corresponding code refactoring to adapt to the new CRD designs

All CRDS now include OpenAPIv3 Schema for validation (auto-generated from code)

Renamed many of the deploy module functions from Deploy* to Apply* since these are also now used for updates

Fixed bug where 9997 port was not being opened for standalone pod instances

Added `make generate` command, which is now used to run the K8s code generate, CRD generator, and update all the installation YAML files.

Flipped `make all` back to being alias for `make image`, building locally versus using builder container

Updated Install.md to include docs on CLUSTER_DOMAIN environment variable

Added repository overview to top-level README.md


mikedickey:
Fixes #14 and #17

mikedickey:
Fixes #31 and #32

PR #40, opened 2020-03-26
Removing a bunch of auto-generated files from deploy directory

mikedickey:
Updated "make generate" so that it only applies to CRDs YAML files and golang code that is autogenerated from pkg/api source code

Moved YAML install file generate to the build/package.sh script. This can now be used to generate a release-<VERSION> directory with all the relevant files for a new release of Splunk Operator.


kelvinatsplunk:
I think we can remove the role_binding.yaml file too ?

mikedickey:
> I think we can remove the role_binding.yaml file too ?

Yes, it is removed by this PR

Issue #50, opened 2020-03-29
Cluster peers do not require persitance


rfaircloth-splunk:
In the current generated deployment both etc and var create pvc for etc and var, however, replication means this is not required. Provide a means to disable generation of pv mounts to use ephemeral storage instead.

mikedickey:
I think persistent volumes are required regardless of replication. I suspect what you may be looking for is [Local Persistent Volumes](https://kubernetes.io/docs/concepts/storage/volumes/#local).

themrkeys:
emptydir would be just fine for indexers protected by smartstore
https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

mikedickey:
Maybe fine for experimentation, but I wouldn't feel comfortable with anyone using emptydir in production. Even with SmartStore, Splunk Enterprise has a long way to go before I think we could get that to work well.

Caldas:
I agree with @mikedickey

Caldas:
Hey @rfaircloth-splunk can we close this issue?

PR #52, opened 2020-03-30
Check cr volumes changes

### LINKED ISSUE #39: Adding new volumes to CRs yields broken StatefulSet
Create a Standalone, IndexerCluster, SearchHeadCluster or LicenseMaster CR. Modify the spec the include additional `volumes`. Operator updates StatefulSet to include the new container VolumeMount, but fails to include the Volume spec in the template.

Operator next will recognize that the pods require updates, shutdown and terminate the highest-numbered pod. The pod will fail to restart because the spec for the container references a Volume that it doesn't recognize. It will remain stuck in this state until you remove the `volumes` from the corresponding CR, at which point it fixes itself.

kelvinatsplunk:
Fixed #39.  Compare volumes for Standalone, IndexerCluster, SearchHeadCluster or LicenseMaster CRs.

### DIFF
--- a/pkg/splunk/resources/util.go
+++ b/pkg/splunk/resources/util.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/splunk/resources/util.go
```diff
@@ -197,8 +198,86 @@ func CompareEnvs(a []corev1.EnvVar, b []corev1.EnvVar) bool {
...
+
+	// make sorted copies of a and b
+	aSorted := sortVolumes(a)
+	bSorted := sortVolumes(b)
+
+	// iterate elements, checking for differences
+	for n := range aSorted {
+
+		// Compare secret volume separately because defaultMode is optional
+		if aSorted[n].VolumeSource.Secret != nil {
```
mikedickey: Does reflect.DeepEqual() not include these in the comparisons? Not clear to me why just calling that isn't sufficient.

pkg/splunk/resources/util.go
```diff
@@ -197,8 +198,86 @@ func CompareEnvs(a []corev1.EnvVar, b []corev1.EnvVar) bool {
...
+	}
+	return false
+}
+
+func compareConfigMapVolume(a, b *corev1.ConfigMapVolumeSource) bool {
+	if a.Name != b.Name {
+		return true
+	}
+
+	if a.DefaultMode != nil && b.DefaultMode != nil {
```
mikedickey: Assuming these comparison methods are necessary, I think these pointer comparisons need to be updated to account for one being nil while the other is not.

pkg/splunk/resources/util.go
```diff
@@ -197,8 +198,86 @@ func CompareEnvs(a []corev1.EnvVar, b []corev1.EnvVar) bool {
...
+			return compareSecretVolume(aSorted[n].VolumeSource.Secret, bSorted[n].VolumeSource.Secret)
+		}
+
+		// Compare config volume separately because defaultMode is optional
+		if aSorted[n].ConfigMap != nil {
+			return compareConfigMapVolume(aSorted[n].VolumeSource.ConfigMap, bSorted[n].VolumeSource.ConfigMap)
+		}
+
+		// Must use DeepEqual() because there are pointers inside.
+		if !reflect.DeepEqual(aSorted[n], bSorted[n]) {
```
gjothikumar-splunk: I think this comment has already been added above, with deep equals already present here why is checking of nested attributes necessary above ?

pkg/splunk/resources/util.go
kelvinatsplunk: The issue is the optional field "DefaultMode" . k8s initialize this field if nil (https://github.com/kubernetes/kubernetes/blob/94837daab20ce5b3fee00f26483a4d64c4d7b9e4/pkg/apis/core/v1/defaults.go#L237). So every time we compare, it will not be equal. Another option is we initialize this field ourselves when we create the volumes internally AND when the user specifies it via CR. I have another PR for this but I think this might be cleaner since we are doing it in 1 place here.

pkg/splunk/resources/util.go
kelvinatsplunk: k8s always initializes this field per https://github.com/kubernetes/kubernetes/blob/94837daab20ce5b3fee00f26483a4d64c4d7b9e4/pkg/apis/core/v1/defaults.go#L237 . So if one is nil (ie "spec"), there is no need to check for change.

pkg/splunk/resources/util.go
kelvinatsplunk: See above comment about the "DefaultMode" field being initialized by k8s.

pkg/splunk/resources/util.go
mikedickey: If we make the assumption that this is always only used to compare K8s-updated version to one generated by operator, it effectively means it is dead code because we know b.DefaultMode is always nil. In this case, I think it would be better to just not have anything. My bigger concern is that these aren't really performing the comparisons one would expect them to be performing, which means there is a high likelihood they would eventually be misused in such a way that introduces bugs.

pkg/splunk/resources/util.go
mikedickey: Another thing we need to consider is we may want to move to having volumeClaimTemplates similar to serviceTemplate, which would allow for more flexibility as contemplated in #50. This would break the assumptions you're making in here.

My preference would be to initialize these fields and keep the comparison methods strict. Wonder if would make sense to call these SetDefaults methods ourselves to ensure they stay in sync?

pkg/splunk/resources/util.go
kelvinatsplunk: Not sure I followed. The k8s-updated version always set DefaultMode to 420 if nil. But the operator version can be nil or non-nil (ie user can specified it in the CR).

pkg/splunk/resources/util.go
kelvinatsplunk: That was my thought initially to keep the comparison strict. Let me bring up the PR for this case.

PR #55, opened 2020-03-31
initial fix to cr volume

### LINKED ISSUE #39: Adding new volumes to CRs yields broken StatefulSet

kelvinatsplunk:
Fixed #39. Added cr.Volumes check and initialize optional properties in cr.Volumes to default values.


pkg/splunk/enterprise/configuration.go
```diff
@@ -196,6 +196,33 @@ func GetSplunkService(cr enterprisev1.MetaObject, spec enterprisev1.CommonSpec,
 	return service
 }
 
+// validateVolumes initializes optional properties in Volumes to default values
+func validateVolumes(spec *enterprisev1.CommonSplunkSpec) {
```
gjothikumar-splunk: The name "validateVolumes" seems to be misleading as we are setting default values to the spec objects. Something like setDefaults or initDefaults ?

pkg/splunk/enterprise/configuration.go
```diff
@@ -556,20 +583,28 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
 
+	// Explicitly set the default value here so we can compare for changes correctly with current statefulset.
+	secretVolDefaultMode := int32(corev1.SecretVolumeSourceDefaultMode)
+
 	// add defaults secrets to all splunk containers
 	addSplunkVolumeToTemplate(podTemplateSpec, "secrets", corev1.VolumeSource{
 		Secret: &corev1.SecretVolumeSource{
-			SecretName: GetSplunkSecretsName(cr.GetIdentifier(), instanceType),
+			SecretName:  GetSplunkSecretsName(cr.GetIdentifier(), instanceType),
+			DefaultMode: &secretVolDefaultMode,
```
gjothikumar-splunk: Since the default mode is already set above (validateVolumes) can't it be retrieved from the CR spec here instead of initializing with corev1.SecretVolumeSourceDefaultMode.

pkg/splunk/enterprise/configuration.go
```diff
@@ -556,20 +583,28 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
+	configMapVolDefaultMode := int32(corev1.ConfigMapVolumeSourceDefaultMode)
+
 	// add inline defaults to all splunk containers
 	if spec.Defaults != "" {
 		addSplunkVolumeToTemplate(podTemplateSpec, "defaults", corev1.VolumeSource{
 			ConfigMap: &corev1.ConfigMapVolumeSource{
 				LocalObjectReference: corev1.LocalObjectReference{
 					Name: GetSplunkDefaultsName(cr.GetIdentifier(), instanceType),
 				},
+				DefaultMode: &configMapVolDefaultMode,
```
gjothikumar-splunk: Since the default mode is already set above (validateVolumes) can't it be retrieved from the CR spec here instead of initializing with corev1.ConfigMapVolumeSourceDefaultMode.

gjothikumar-splunk:
Just confirming, since this is a bug on existing 0.1.0 release this will be merged to master and develop. Is that correct

pkg/splunk/enterprise/configuration.go
kelvinatsplunk: I will change it to setVolumeDefaults()

pkg/splunk/enterprise/configuration.go
kelvinatsplunk: validateVolumes() validates/set what the user specified in the CR. The addSplunkVolumeToTemplate() adds **additional** volumes after what the user adds in the CR.

pkg/splunk/enterprise/configuration.go
kelvinatsplunk: Comment above

Issue #56, opened 2020-03-31
unable to enable ssl on indexer cluster


rfaircloth-splunk:
While it might generally be ok to accept s2s communication among nodes in a deployment within k8s because of network protections such as weave. External exposure requires SSL. Enable TLS on additional port and provide the ability to reference a certificate from a k8s secret as the input

mikedickey:
Did you try the [Ingress examples in docs](https://github.com/splunk/splunk-operator/blob/master/docs/Ingress.md)?

Caldas:
@rfaircloth-splunk did the doc solved your problem?

Caldas:
I used cert-manager and it's worked fine @rfaircloth-splunk.

Do you still need help on this or @mikedickey could close?

PR #59, opened 2020-04-02
Updated docs on using Istio to route S2S traffic

mikedickey:


mikedickey:
Fixes #56

Caldas:
@mikedickey this is the best/recommended way to set the communication between nodes?

mikedickey:
> @mikedickey this is the best/recommended way to set the communication between nodes? If yes, I can test on my new cluster

We use Istio at Splunk, and it seems to have most traction. Doesn't mean you can't use something else, but we will likely continue to focus on it for examples, testing, etc.

Issue #47, opened 2020-03-29
Optionally expose additional ports on service


rfaircloth-splunk:
Certain apps such as webhooks may require additional customer ports provide a way to indicate that should be exposed on the service definitions

mikedickey:
Seems like this should be pretty easy to do using `serviceTemplate` but currently the code in `GetSplunkService()` overwrites any values for ports specified by user. It probably should use `append()` instead. Also, you may need to modify the PodTemplate for any corresponding StatefulSet for this to work as expected. We'll need to see if there is a clean way to do this without having to add more config parameters to CRDs..

rfaircloth-splunk:
Yes I looked at ServiceTemplate but with no example I did not grok it

PR #60, opened 2020-04-03
Allow svc ports

kelvinatsplunk:
Support #47. Update service and statefulset objects with user-specified ports in CR.


pkg/splunk/enterprise/configuration.go
```diff
@@ -223,10 +225,26 @@ func setVolumeDefaults(spec *enterprisev1.CommonSplunkSpec) {
 	}
 }
 
+func setServiceTemplateDefaults(spec *enterprisev1.CommonSplunkSpec) {
+	if spec.CommonSpec.ServiceTemplate.Spec.Ports != nil {
```
mikedickey, line 229: seems like we should also check for ServiceTemplate != nil and ServiceTemplate.Spec != nil

pkg/splunk/enterprise/configuration.go
```diff
@@ -570,6 +590,20 @@ func getSplunkStatefulSet(cr enterprisev1.MetaObject, spec *enterprisev1.CommonS
 // updateSplunkPodTemplateWithConfig modifies the podTemplateSpec object based on configuration of the Splunk Enterprise resource.
 func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec, cr enterprisev1.MetaObject, spec *enterprisev1.CommonSplunkSpec, instanceType InstanceType, extraEnv []corev1.EnvVar) {
 
+	// Add custom ports to splunk containers
+	if spec.ServiceTemplate.Spec.Ports != nil {
```
mikedickey, line 594: check for ServiceTemplate != nil and ServiceTemplate.Spec != nil ?

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -80,19 +85,23 @@ func TestGetSearchHeadStatefulSet(t *testing.T) {
...
...app.kubernetes.io/instance","operator":"In","values":["splunk-stack1-search-head"]}]},"topologyKey":"kubernetes.io/hostname"}}]}},"schedulerName":"default-scheduler"}},"volumeClaimTemplates":[{"metadata":{"name":"pvc-etc","namespace":"test","creationTimestamp":null,"labels":{"app.kubernetes.io/component":"search-head","app.kubernetes.io/instance":"splunk-stack1-search-head","app.kubernetes.io/managed-by":"splunk-operator","app.kubernetes.io/name":"search-head","app.kubernetes.io/part-of":"splunk-stack1-search-head"}},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"10Gi"}}},"status":{}},{"metadata":{"name":"pvc-var","namespace":"test","creationTimestamp":null,"labels":{"app.kubernetes.io/component":"search-head","app.kubernetes.io/instance":"splunk-stack1-search-head","app.kubernetes.io/managed-by":"splunk-operator","app.kubernetes.io/name":"search-head","app.kubernetes.io/part-of":"splunk-stack1-search-head"}},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"100Gi"}}},"status":{}}],"serviceName":"splunk-stack1-search-head-headless","podManagementPolicy":"Parallel","updateStrategy":{"type":"OnDelete"}},"status":{"replicas":0}}`)
```
mikedickey, line 88: painful that a spelling error fix trickles out like this.. not sure if it's worth tackling now, but maybe we could split all these checks across multiple lines so it's easier to see and compare diffs? or some other way? maybe we should add a backlog item to try and tackle this better in the future?

pkg/splunk/reconcile/service.go
```diff
@@ -22,21 +22,33 @@ import (
...
-		scopedLog.Info("Found existing Service")
-	} else {
-		err = CreateResource(client, service)
+	if err != nil {
+		return CreateResource(client, revised)
 	}
 
-	return err
+	// found existing Service: do nothing
+	scopedLog.Info("Found existing Service")
```
mikedickey, line 39: this comment could probably just be removed..

I don't recall why it wasn't applying updates for services. i imagine partly because we didn't expect them to change, but also when i tried flipping it, it was *always* making updates to services, even when there were no differences. looks like you tackled this below with MergeServiceSpecUpdates, but just wanted to make sure you tested that it's not updating services every time it reconciles.

pkg/splunk/reconcile/util.go
```diff
@@ -204,3 +204,20 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name
 
 	return result
 }
+
+// MergeServiceSpecUpdates merges the current and revised spec of the service object
+func MergeServiceSpecUpdates(current *corev1.ServiceSpec, revised *corev1.ServiceSpec, name string) bool {
+	scopedLog := log.WithName("MergeServiceSpecUpdates").WithValues("name", name)
+	result := false
+
+	// check for changes in Ports
```
mikedickey, line 213: Seems like we should be checking more than ports.. If a user modifies an IndexerCluster CR for example to include a serviceTemplate that previously was nil, there could be many additional changes beyond just ports. For example, type may change from ClusterIP to NodePort.

pkg/splunk/enterprise/configuration.go
kelvinatsplunk: These are inner structs.

pkg/splunk/enterprise/configuration.go
kelvinatsplunk: These are inner structs.

pkg/splunk/enterprise/configuration_test.go
kelvinatsplunk: Agree. I had to use jsondiff to see the change myself :-)

pkg/splunk/reconcile/service.go
kelvinatsplunk: Will remove the comment. I tested it wasn't making unnecessary updates. I had to initialize the targetPort field since its optional.

pkg/splunk/reconcile/util.go
kelvinatsplunk: Let me look closer here.

kelvinatsplunk: Added more checks for changes in service spec specifically on ports, external ips, type, external name.

pkg/splunk/enterprise/configuration_test.go
kelvinatsplunk: I looked a bit here. I think we should tackle this together with the integration test framework. For example, I was looking at Ginkgo and there is a matcher package called Gomega. It has a matcher for dealing with complex struct comparison. I think, for unit-testing, if we change a specified field say the replica field in the CR, we should verified just that field in the statefulset object (or other k8s objects) vs verifying the entire object.

kelvinatsplunk: I guess I am saying we refactor unit test to use those "assert" or "matcher" package.

terciodemelo:
Hello @kelvinatsplunk , would you kindly clarify something to me? Does just appending handles correctly conflicts? I found this PR while trying to have a fixed 'nodePort' mapping, instead of the default random assignment of ports. For example:

``` yaml
      - name: splunkweb
        nodePort: 32000
        port: 8000
        protocol: TCP
        targetPort: 8000
```

Caldas:
Hey @mikedickey, @terciodemelo is doing the project I told you with me. Can you anwser us?

kelvinatsplunk:
Hi @terciodemelo @Caldas 
Sorry for the delay in response. Yes, this change will handle the add/update in the serviceTemplate.Ports (ServicePorts). I will be merging this asap.

kelvinatsplunk:
Hi @terciodemelo @Caldas ,
FYI. I have merged the service port change in.

terciodemelo:
Thank you very much, @kelvinatsplunk!

Issue #64, opened 2020-04-09
Select image via shared configuration


rfaircloth-splunk:
The image version should be pulled from a shared config map or new KIND . In the current implementation on a node failure, the new node pulled a new image for one SHC peer which was newer than the CM image causing a SHC cluster failure. To avoid this "latest/edge" should not be used a specific numbered version should be required and that version should be shared by all related components

kelvinatsplunk:
Right now, the enterprise image version is specified in the operator yaml.  I agree we should put a specific version instead of "latest/edge". I don't think we need a shared config. When we support upgrade, a new KIND with version field may make sense.

mikedickey:
I'm working on a PR that includes pinning it to 8.0.3. Note that 8.0.2 will not work with 0.1.0.

Issue #24, opened 2020-01-23
Proposal to include SOK in OperatorHub


dmesser:
Hi Splunk team, since you are using the Operator-SDK for the Splunk you are not that far away to generate CSV and publish your Operator at [OperatorHub.io](https://operatorhub.io). The SDK supports creating most of the CSV skeleton for you via the `gen-csv` subcommand or you can use a graphical UI at https://operatorhub.io/bundle drop in your k8s manifest and wrap up the rest of the metadata.

WDYT?

mikedickey:
We are working on this and hope to have it included soon.

dmesser:
Awesome, looking forward to your PR.

mikedickey:
PR for OperatorHub submission is at https://github.com/operator-framework/community-operators/pull/1027

mikedickey:
https://operatorhub.io/operator/splunk

PR #65, opened 2020-04-10
Support for OLM and other things

mikedickey:
Fixed an old bug that resurfaced itself, causing the operator to fail in cluster-scope. Default client tries to cache all objects from all namespaces when WATCH_NAMESPACE is empty, causing auth failures. The fix bypasses this (broken) cache, allowing permissions to be approrpiately scoped to namespaces.

Added build/make_bundle.sh which uses the "operator-sdk generate csv" command to generate a splunk OLM bundle in the deploy/olm-catalog/splunk directory. This is executed as part of the "make generate" command.

You can now use OLM to install the splunk operator bundle to your Kubernetes cluster. If you do not have OLM installed, you can use "operator-sdk olm install" to install it (see "operator-sdk olm --help" for more). Once you have OLM installed, ensure that your ~/.kube/config context points to a clean namespace and then run "operator-sdk run --olm --manifests=deploy/olm-catalog/splunk --operator-version=0.1.0" Just delete the namespace when you are done testing.

All the deploy/*.yaml files are now sources used to construct the OLM bundle and/or standalone YAML installation files (which are generated via build/package.sh).

Removed deploy/cluster_operator.yaml and deploy/namespace_operator.yaml since these have become redundant files. They are no longer necessary because the build/package.sh script now constructs all the proper install/release YAML files from sources located in deploy/*.yaml.

Changed splunk:operator:namespace-manager from a ClusterRole to a Role, since this is easier for reuse across tools and it is not required to be a ClusterRole.

Removed splunk:operator:resource-manager from the YAML used for namespace-scoped installs, since it only is required for cluster-scoped.

These changes also made the "splunk-operator-rbac.yaml" install resource unnecessary, since CRDS are now the only cluster-admin resources that need to be installed for namespace-scoped deployments.

Note that these changes involve more YAML processing than what we previously did, so the "yq" utilitity is now a required development tool. Updated README for install instructions.

Added build/run_scorecard.sh script which can be used to run the operator-sdk scorecard test suite. You can also use "make scorecard" to run this. Note that this requires having OLM installed in your Kubernetes cluster and that your ~/.kube/config context points to a clean namespace. Also note that the "statusdescriptorstest" test currently fails due to missing descriptors for some
of the status fields. I didn't fix this because it would require a CRD update.


mikedickey:
Fixes #64 by pinning a specific release tag to all the YAML install files

mikedickey:
Fixes #66 and #24

kelvinatsplunk:
Does this change means we cannot install splunk operator outside of OLM ?

mikedickey:
> Does this change means we cannot install splunk operator outside of OLM ?

No, it's purely optional. You only need to have OLM installed to run the scorecard tests

kelvinatsplunk:
I think we need to add "image" dependency to "package" target. I was running make package and it failed looking for splunk/splunk-operator:latest to tag.

kelvinatsplunk:
In the release-xxx folder generated by make package, it seems like the files "splunk-operator-crds.yaml and splunk-operator-noadmin.yaml" are transient ? User only need to apply splunk-operator-cluster.yaml or -install.yaml. Can we remove them to avoid confusion ?

mikedickey:
> In the release-xxx folder generated by make package, it seems like the files "splunk-operator-crds.yaml and splunk-operator-noadmin.yaml" are transient ? User only need to apply splunk-operator-cluster.yaml or -install.yaml. Can we remove them to avoid confusion ?

See https://github.com/splunk/splunk-operator/blob/olm-support/docs/Install.md#installation-by-a-regular-user

mikedickey:
> I think we need to add "image" dependency to "package" target. I was running make package and it failed looking for splunk/splunk-operator:latest to tag.

I'll make sure I include this in my next push

Issue #53, opened 2020-03-30
operator 0.1.0 does not work when password is configured using defaults


mikedickey:
Previous releases allowed you to specify secrets like `splunk.password` via `defaultsUrl` or `defaults` spec parameters. 0.1.0 assumes that the auto-generated admin passwords it creates in kubernetes are valid, so if you change it using defaults, upgrades, updates or scaling operations may fail.

One possible approach to fix this is adding a `secretsRef` to the `IndexerCluster`, `SearchHeadCluster`, `LicenseMaster` and `Standalone` CRDs. This would refer to a Kubernetes secret that includes one or more of the secrets used by the operator:

* hec_token
* password
* pass4SymmKey
* idxc.secret
* shc.secret

Any which are not defined would be auto-generated. This would provide a mechanism for users to provide their own secrets, but it would still fail when different ones are set via defaults.yml.

Another, more direct approach may be for operator to parse defaults. Currently it does not understand defaults and just passes these through to the container. Parsing the YAML and using it to override settings like secrets may be a more general purpose solution that can be reused to avoid other potential mismatch issues between these things.

Whatever approach we use, I think we should support the storage of these secrets in Vault, given its popularity.

mikedickey:
Relates to https://github.com/splunk/splunk-ansible/issues/511

akondur:
[Closing the issue with new implementation of ](https://github.com/splunk/splunk-operator/blob/master/docs/PasswordManagement.md)password management.

PR #68, opened 2020-04-17
[CSPL-236] Implementation of Splunk Secrets override

gjothikumar-splunk:
Addresses issue https://github.com/splunk/splunk-operator/issues/53

- Implementation of Splunk secrets override through k8s secret
- An improvement to the current security model where secrets to be overridden are provided through defaultsUrl. Moving the override to k8s secrets is more secure
- The Splunk operator and Splunk Enterprise system all use the same secrets.

### DIFF
--- a/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
@@ -760,6 +760,43 @@ spec:
               description: Name of Scheduler to use for pod placement (defaults to
                 “default-scheduler”)
               type: string
+            secretsRef:
+              description: SecretsRef refers to kubernetes secret that holds Splunk
+                secrets that are overriden by
+              properties:
+                apiVersion:
+                  description: API version of the referent.
+                  type: string
+                fieldPath:
+                  description: 'If referring to a piece of an object instead of an
+                    entire object, this string should contain a valid JSON/Go field
+                    access statement, such as desiredState.manifest.containers[2].
+                    For example, if the object reference is to a container within
+                    a pod, this would take on a value like: "spec.containers{name}"
+                    (where "name" refers to the name of the container that triggered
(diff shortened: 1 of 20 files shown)


Caldas:
Great @gjothikumar-splunk, it looks like fine .. just decreased too much the code coverage so more tests are required 🚀

gjothikumar-splunk:
> Great @gjothikumar-splunk, it looks like fine .. just decreased too much the code coverage so more tests are required 🚀

Thanks for the review @Caldas. I am still working on tests and will make sure coverage is maintained.

gjothikumar-splunk:
We had a discussion on this and want to simply the whole secrets management and override in the operator. This is what we are thinking currently

- We will have a common set of secrets across all Splunk Enterprise components / CRs (indexer, sh, license master)
- If customer wants to override any secret and make all components use that the step would be to create a k8s secret called "splunk-secrets" and add the secrets to be overridden (pass4symmkey, password etc..) into that.
- The first CR that gets bootstrapped will look for "splunk-secrets", use the secrets form there and auto generate anything that does not exist there. All components will use the same set of secrets.

This will make secret management simpler, more predictable and also reduce unnecessary dependencies between components. This is a backward incompatible change and it will break the current deployments. At this early stages of the product, the cleaner way would be the following

- Before upgrade, create "splunk-secrets" with secrets to be used across the cluster.
- After upgrade the operator will use all secrets mentioned in "splunk-secrets" and generate the others. These should be used from thereon when interactig with Splunk Enterprise.

@mikedickey @kelvinatsplunk

Caldas:
It's a great Idea !!

Em ter, 21 de abr de 2020 16:18, Ganesh Jothikumar <notifications@github.com>
escreveu:

smohan-splunk:
Closing this PR as it is outdated. Splunk Secrets updated implementation has been merged to develop 
cc: @gjothikumar-splunk

Issue #67, opened 2020-04-12
Add support for Taint Tolerations


terciodemelo:
It would be very nice if we had support for `tolerations` in [common spec parameters](https://github.com/splunk/splunk-operator/blob/master/docs/CustomResources.md#common-spec-parameters-for-all-resources) just like `affinity`, this would enable users to leverage [Taints & Toleration](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) feature in their Splunk clusters as well.

Caldas:
Hey, @mikedickey this is something that we would like to test, so we can group node to SearchHead and Indexers better .. 

Did you like the approach?

mikedickey:
I think it would be quite easy to add pass-through for tolerations, as suggestion.

kelvinatsplunk:
Hi @Caldas @terciodemelo,
Appreciate if you can take a quick look at https://github.com/splunk/splunk-operator/pull/73.

Caldas:
Great @kelvinatsplunk ... I believe you can close this issue 😉

PR #73, opened 2020-04-22
Support taint toleration

kelvinatsplunk:
#67 
- Add tolerations field to CommonSpec to pass thru to StatefulSet and Deployment (Spark)
- Refactored CompareXXX() routines to use sortAndCompareSlices()


kelvinatsplunk:
@mikedickey Not sure why code coverage drop but looking to add more test.

PR #81, opened 2020-05-22
Ephemeral storage

mikedickey:
Added support for ephemeralStorage Splunk Common Spec parameter

Renamed CompareIPLists to CompareSortedStrings since it is a more generic function

Added some more unit tests to increase code coverage

Bump version to 0.2.0 and CRD version to v1alpha3


mikedickey:
Fixes #50

Issue #34, opened 2020-02-14
Provide Example for using Local Persistent Volumes with SOK


crawdaddy18:
Please provide an example of how to configure the Splunk on Kubernetes to work with Local Persistent Volumes.

Through personal testing, I have set up a "local-storage" StorageClass:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: local-storage
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: WaitForFirstConsumer
```
However, any attempts to use this SC leaves the Pods in a "Pending" status which may be related to no claim actually being made.

Multiple attempts have been made to modify the splunk-operator by downloading the YAML from the tiny.cc URL and the SplunkEnterprise YAML, but so far no success.

mikedickey:
Hi Chris, I haven't tried using local volumes yet but agree it's something we need to do and document. I do know that just adding a StorageClass is not sufficient. You'd need to manually create a bunch of PVs on your servers for it to use, unless you are using something else to do it for you. Have you tried this? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner

crawdaddy18:
@mikedickey I did try the link and did get it set up, but I don't think that was sufficient for what I was doing because I had only 2 EC2 instances and I believe that sidecar creates only 1 PV for each machine.

Here is what I did to get things working (EC2 instance with 1 500GB drive)

On each host I created directories in /mnt/disk by running this script:
```
#!/bin/bash

for i in {1..10}
do
  DIRNAME="vol${i}"
  mkdir -p /mnt/disk/$DIRNAME 
  chcon -Rt svirt_sandbox_file_t /mnt/disk/$DIRNAME
  chmod 777 /mnt/disk/$DIRNAME
done
```

In addition to the StorageClass above, here are my PV definitions:
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: my-local-pv10
  labels:
    type: local
spec:
  capacity:
    storage: 50Gi
  accessModes:
    - ReadWriteOnce
  persistentVolumeReclaimPolicy: Retain
  storageClassName: local-storage
  local:
    path: /mnt/disk/vol10
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - ip-10-20-1-108.us-west-2.compute.internal

---
```
[...]

PR #84, opened 2020-05-28
Updated storage docs to include new ephemeralStorage feature

mikedickey:
and pointers for using Local Persistent Volumes.


mikedickey:
Fixes #50 and #34

Issue #21, opened 2020-01-12
Standalone does not deploy a service


matthewmodestino:
```
docker-splunk image version: splunk/splunk:8.0.1
Kubernetes variant: MicroK8s
Issue: My standalone does not deploy a service that I can point my nginx ingress at. 
Workaround: Created it manually. 
```

Hey Team, 

Is it expected that a single standalone Splunk instance will not deploy a service? 

I am following the ingress docs, but am running a single standalone instance, rather than a cluster, and it looks like it doesn't trigger a service when i check `kubectl -n splunk get services -o name`

Pretty easy for me to deploy one so I can expose it to my nginx ingress, just making sure I didn't miss something when trying to follow: https://github.com/splunk/splunk-operator/blob/master/docs/Ingress.md#configuring-ingress

```
kubectl -n splunk describe splunkenterprise mattymo
Name:         mattymo
Namespace:    splunk
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"enterprise.splunk.com/v1alpha1","kind":"SplunkEnterprise","metadata":{"annotations":{},"app":"splunk","finalizers":["enterp...
API Version:  enterprise.splunk.com/v1alpha1
Kind:         SplunkEnterprise
Metadata:
  Creation Timestamp:  2020-01-12T15:26:54Z
  Finalizers:
    enterprise.splunk.com/delete-pvc
  Generation:        1
  Resource Version:  2107333
  Self Link:         /apis/enterprise.splunk.com/v1alpha1/namespaces/splunk/splunkenterprises/mattymo
  UID:               1ef9dd3d-f8e5-4508-a427-fc3a30eb9c5b
Spec:
  Defaults:  splunk:
  password: helloworld456
  hec_token: 00000000-0000-0000-0000-000000000000
  apps_location: "/tmp/ta_container_workshop.tgz,/tmp/splunk-app-for-infrastructure_201.tgz,/tmp/splunk-add-on-for-infrastructure_201.tgz,/tmp/splunk-business-flow_116.tgz" 
  License URL:  /mnt/licenses/Personal-50GB-Splunk.License
  Resources:
    Splunk Etc Storage:  5Gi
    Splunk Var Storage:  300Gi
  Splunk Image:          matthewmodestino/container_workshop:8.0.1
  Splunk Volumes:
    Config Map:
      Name:  splunk-licenses
    Name:    licenses
Events:      <none>
```

[...]

Issue #93, opened 2020-06-23
selector does not match template labels errors


mikedickey:
Creating a custom resource that includes certain labels will causing provisioning to fail. For example, the following resource:

```
apiVersion: enterprise.splunk.com/v1alpha2
kind: Standalone
metadata:
  name: test
  labels:
    app.kubernetes.io/managed-by: anyone
  finalizers:
    - enterprise.splunk.com/delete-pvc
spec: {}
```

Will yield the following errors in the splunk operator logs:

```
{"level":"error","ts":1592936956.9408255,"logger":"splunk.reconcile.UpdateResource","msg":"Failed to update resource","name":"splunk-test-standalone","namespace":"user-mdickey","error":"StatefulSet.apps \"splunk-test-standalone\" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{\"app.kubernetes.io/component\":\"standalone\", \"app.kubernetes.io/instance\":\"splunk-test-standalone\", \"app.kubernetes.io/managed-by\":\"anyone\", \"app.kubernetes.io/name\":\"standalone\", \"app.kubernetes.io/part-of\":\"splunk-test-standalone\"}: `selector` does not match template `labels`"...
```

This is because operator attempts to copy labels from custom resources into any service, deployment or statefulset objects it creates that are associated with them. The `app.kubernetes.io/managed-by` is one of the labels that operator applies itself (along with a few others) to statefulsets, and the `selector` field of the spec has to match the pod template labels. The copy from the custom resource spec clobbers the labels being set by operator, causing a mismatch and resulting in these errors from Kubernetes.

The labels that will cause this error include:

* app.kubernetes.io/managed-by
* app.kubernetes.io/component
* app.kubernetes.io/name
* app.kubernetes.io/part-of
* app.kubernetes.io/instance

PR #94, opened 2020-06-23
Added regular service for Standalone custom resource

mikedickey:
Also fixed bug with labels and annotations set by operator being clobbered by labels and annotations set on custom resources.


mikedickey:
Fixes #21 and #93

PR #99, opened 2020-07-29
Multisite / multipart indexer cluster

### LINKED ISSUE #44: Support multi site cluster of indexers
In a cluster with multiple availability zones (1.17+) provide the ability to define a indexer cluster where indexers are added to a Splunk Site and meta for affinity to the zone is added to each indexer pod so that all indexer pods of the same site are in the same zone.

romain-bellanger:
Closes #44 

This PR is structured into multiple commits which can be easier to review individually.

The purpose of this change is to support the connection of multiple IndexerCluster resources together, which for instance allows to:
* Deploy a cluster-master with different spec from the indexers (storage size, StorageClass, affinity)
* Deploy a multisite indexer cluster by connecting multiple IndexerCluster resources together (to the same cluster-master). The IndexerCluster parts managing the indexer can be constrained to run in a specific zone using affinity and assigned a Splunk site using ansible defaults (solution discussed in [this issue](https://github.com/splunk/splunk-operator/issues/44))

In order to cause minimum disruption, most of the code changes are only activated when the indexerClusterRef parameter is defined. This parameter was already present in the common spec of all Splunk CRDs, but not used so far for IndexerCluster.

However, so some impacts on the existing functionality have been identified:
* The `validateIndexerClusterSpec()` method was ensuring IndexerCluster had at least one replica (0 automatically changed to 1). This was removed to authorize the deployment of a cluster-master only with an IndexerCluster resource. Existing clusters deployed without specifying replicas (or already specifying 0 replicas) will be downscaled from 1 to 0 indexers.
* A new variable `SPLUNK_CLUSTER_MASTER_URL=localhost` is passed to the cluster master pod. This makes splunk-ansible configure indexer discovery on the cluster master, instead of using a static list of indexers. The other servers (standalone instances, search head clusters) are already configured with indexer discovery when connected to an IndexerCluster. This is seen as a general improvement for autoscaling.
* The label `app.kubernetes.io/instance` has been removed from the selector of the indexer services created by IndexerClusters (unless they specify indexerClusterRef). The label `app.kubernetes.io/part-of` is currently already set to the same value, so the load balancing shouldn't be impacted for single site clusters. For multisite, this allows the indexer services created by the cluster-master IndexerCluster part to select the indexer of all the parts (sharing the same `app.kubernetes.io/part-of` label), while the indexer service created by each part only selects the indexers which it manages using the `app.kubernetes.io/instance`, in order to allow client pods to send traffic to the indexers located in the same availability zone if needed.
* Pods created by Standalone resources get the role `SPLUNK_ROLE=splunk_search_head` when connected to a cluster master, as the [setup_multisite.yml task of the splunk_standalone role](https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_standalone/tasks/setup_multisite.yml) is empty and apparently not meant to be implemented.


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -80,46 +81,65 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
...
-	// create or update statefulset for the indexers
-	statefulSet, err = getIndexerStatefulSet(cr)
-	if err != nil {
-		return result, err
-	}
-	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: secrets, newSplunkClient: splclient.NewSplunkClient}
-	phase, err = mgr.Update(client, statefulSet, cr.Spec.Replicas)
-	if err != nil {
-		return result, err
+	if cr.Spec.Replicas > 0 {
```
gaurav-splunk, line 128: Probably we can use similar approach to skip creation of indexer services(headless and regular) i.e. only create indexer services if cr.Spec.Replicas >0 ? What do you think?

gaurav-splunk, line 128: Probably we can use similar approach to skip creation of indexer services(headless and regular) i.e. only create indexer services if cr.Spec.Replicas >0 ? What do you think?

romain-bellanger: Hi Gaurav, thanks a lot for the review :-)
Actually with the proposed solution, the indexer services (both headless and regular) created by the IndexerCluster part containing the cluster master don't select the `app.kubernetes.io/instance` label (see 3rd identified impact in PR description), so they can be used to load balance (regular) or resolve (headless) the indexers of all the other parts, which share the same `app.kubernetes.io/part-of` label.
The services created by each part/site only select the indexers of that specific part/site, which can be used to prevent cross-zone traffic (charged by some cloud providers). This provides a solution similar to [per site indexer discovery](https://docs.splunk.com/Documentation/Splunk/8.0.5/Indexer/indexerdiscovery#Use_indexer_discovery_in_a_multisite_cluster) usable with HEC inputs for instance.
The only limitation is that there is no service selecting only the indexers deployed in the IndexerCluster containing the cluster-master. This was discussed in conference call together with @mikedickey who answered that this limitation is acceptable, and seemed to approve the solution.

docs/MultisiteExamples.md
```diff
@@ -0,0 +1,274 @@
...
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: failure-domain.beta.kubernetes.io/zone
+            operator: In
+            values:
+            - zone-1a
+EOF
+```
+
+## Single IndexerCluster resource and Downward API
```
mikedickey, line 107: The Downward API approach seems a lot more complicated to me. I question whether it's a good idea to document here versus having just one recommended & supported way to setup multisite?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -226,7 +246,12 @@ func (mgr *indexerClusterPodManager) getClient(n int32) *splclient.SplunkClient
 
 // getClusterMasterClient for indexerClusterPodManager returns a SplunkClient for cluster master
 func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkClient {
-	fqdnName := splcommon.GetServiceFQDN(mgr.cr.GetNamespace(), GetSplunkServiceName(SplunkClusterMaster, mgr.cr.GetName(), false))
+	masterIdxcName := mgr.cr.GetName()
+	// For multisite / multipart cluster, use the cluster-master API of the referenced IndexerCluster
+	if len(mgr.cr.Spec.IndexerClusterRef.Name) > 0 {
+		masterIdxcName = mgr.cr.Spec.IndexerClusterRef.Name
```
mikedickey, line 252: Any reason not to allow cluster master to reside in a different namespace from indexer cluster CRs?

pkg/splunk/enterprise/indexercluster.go
mikedickey: I think this makes sense, but we should revisit with possible introduction of having a separate `ClusterMaster` CRD

pkg/splunk/enterprise/util.go
```diff
@@ -64,10 +64,21 @@ func ApplySplunkConfig(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
 	// create or retrieve splunk secrets
-	secrets := getSplunkSecrets(cr, instanceType, idxcSecret, pass4SymmKey)
-	secrets.SetOwnerReferences(append(secrets.GetOwnerReferences(), splcommon.AsOwner(cr)))
-	if secrets, err = splctrl.ApplySecret(client, secrets); err != nil {
-		return nil, err
+	var secrets *corev1.Secret
+	if instanceType == SplunkIndexer && len(spec.IndexerClusterRef.Name) > 0 {
+		secrets = &corev1.Secret{}
+		secretRefName := GetSplunkSecretsName(spec.IndexerClusterRef.Name, SplunkIndexer)
+		namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: secretRefName}
```
mikedickey, line 71: Anywhere we reference IndexerClusterRef, we may want to allow that to also specify namespace (which is not the same). at least in other cases for all the refs this is the pattern I tried to follow.

pkg/splunk/enterprise/indexercluster.go
romain-bellanger: HI Mike, thanks a lot for the review and for this finding. I'm rebasing the PR and fixing this issue.

romain-bellanger: I actually found a possible reason not to allow this:
- The indexer service on the cluster master part would not select the indexers from the other parts (I didn't find any clear documentation, but for obvious security reasons I guess the selector of a Service only applies within a namespace)
- It would also not be possible to mount the secret from the cluster master part to the other IndexerCluster resources (documented [here](https://kubernetes.io/docs/concepts/configuration/secret/#restrictions)). This could be addressed by copying and synchronizing the secret, but because of the first limitation I'd like to get your feedback before working on this.

Then if referencing a cluster master residing in a different namespace is not supported, passing a namespace in IndexerClusterRef.Namespace in an IndexerCluster resource should probably be rejected with a clear error message.

mikedickey: Makes sense. I think it's fine to just reject and say that crossing namespaces is not supported.

docs/MultisiteExamples.md
romain-bellanger: I agree that it is more complicated, I just wanted to provide something as complete as possible and throught some user might prefer to operate a single resource, but I don't see any problem to remove it. You could also remove it after merging the PR if you want to keep it in the history.

This type of setup could be integrated within the operator, automatically configuring the downward API, directly using the node name as a variable instead of doing two API calls with ansible lookups. Then a new parameter could be passed to ansible to provide the mapping between zones and sites, and the retrieval of the site could be integrated within splunk-ansible. That could reduce a lot the complexity.

This would have required heavier changes and been more intrusive, maybe adding a new parameter to the CRD, adding a module in the docker image, modifying ansible-splunk, so I didn't feel comfortable to propose this as an external contributor. This also wouldn't have solved my problem of defining different specs for cluster-master and indexers (local SSD storage).

The only difficulty which I am currently experiencing with the dedicated IndexerCluster resource per site is related to autoscaling. When adding indexers, the number of indexers doesn't necessarily remain balanced across sites. When scaling down, it currently removes indexers from several sites in parallel. I didn't expect issues as this is a graceful offline, but I'm actually experiencing troubles with buckets not searchable (which I plan to discuss together with Splunk PS). A setup using a single resource and downward API would of course solve this, but there might be other options in the operator to control this.

romain-bellanger:
PR rebased, error added when referencing a cluster master located in a different namespace. CI tests passed on a GKS cluster, and the operator instances used to operate our development clusters have been updated to run this code.

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -72,6 +75,64 @@ func TestApplyIndexerCluster(t *testing.T) {
 	splunkDeletionTester(t, revised, deleteFunc)
 }
 
+func TestApplyIndexerClusterMultipart(t *testing.T) {
```
gaurav-splunk, line 78: Maybe we can have another test just to test creation of CM part with `Replicas:0` ?

test/testenv/deployment.go
```diff
@@ -87,6 +96,52 @@ func (d *Deployment) GetInstance(name string, instance runtime.Object) error {
 	return nil
 }
 
+// PodExecCommand execute a shell command in the specified pod
+func (d *Deployment) PodExecCommand(podName string, cmd []string, stdin string, tty bool) (string, string, error) {
```
gaurav-splunk, line 100: This is nice!!

pkg/splunk/enterprise/indexercluster_test.go
romain-bellanger: Thanks for your feedback :-) Added unit test TestApplyIndexerClusterMasterOnly

docs/MultisiteExamples.md
```diff
@@ -0,0 +1,274 @@
...
+
+Please refer to the [Configuring Splunk Enterprise Deployments Guide](Example.md)
+for more information and examples about deploying the different Splunk resources
+in single site mode.
+
+## Multisite Indexer Clusters in Kubernetes
+
+[Multisite indexer cluster architecture](https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Multisitearchitecture)
+can be used for various purposes. However, Kubernetes cluster are generally deployed at the scale of
+a region, and the main purpose of this documentation if to cover the topic of building high-available
```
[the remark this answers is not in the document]
romain-bellanger: Fixed in last update

docs/MultisiteExamples.md
```diff
@@ -0,0 +1,274 @@
...
+indexer clusters ensuring that two replicas of the same bucket are never located in the same
```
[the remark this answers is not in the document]
romain-bellanger: Thanks for spotting this, I've modified this sentence in the last update.

PR #105, opened 2020-08-14
CSPL-320: Smartstore Config

sgontla:
- First cut changes just support Smartstore config.
- More changes to come, to support the overal feature


deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
```diff
@@ -2309,13 +2364,37 @@ spec:
...
+                  type: array
+              type: object
           type: object
       type: object
   version: v1alpha3
   versions:
   - name: v1alpha3
     served: true
     storage: true
-  - name: v1alpha2
```
mikedickey, line 2319: Why remove v1alpha2? I think this prevents someone from upgrading v2 to v3.

pkg/apis/enterprise/v1alpha3/common_types.go
```diff
@@ -67,3 +67,25 @@ type CommonSplunkSpec struct {
...
 	IndexerClusterRef corev1.ObjectReference `json:"indexerClusterRef"`
 }
+
+// SmartStoreSpec defines Splunk indexes and remote storage volume configuration
+type SmartStoreSpec struct {
+	// List to accommodate multiple volumes.
+	// For now, a hard check is added such that only one volume is supported to accommodate one common secrete object
+	S3VolList []S3VolumeSpec `json:"volumes,ommitempty"`
+
+	IndexList []IndexSpec `json:"indexes,ommitempty"`
```
mikedickey, line 77: please comment each public field. I'm also surprised/concerned that "make lint" isn't failing this?

pkg/apis/enterprise/v1alpha3/common_types.go
```diff
@@ -67,3 +67,25 @@ type CommonSplunkSpec struct {
...
+}
+
+// S3VolumeSpec defines the remote storage volume configuration
+type S3VolumeSpec struct {
+	RemoteName string `json:"volumeName"`
```
mikedickey, line 82: field naming is inconsistent across these. I think since it's in the context of a "S3VolumeSpec" that just "Name"/"name" is sufficient.

pkg/apis/enterprise/v1alpha3/common_types.go
```diff
@@ -67,3 +67,25 @@ type CommonSplunkSpec struct {
...
+	S3Endpoint string `json:"endpoint"`
+	// Some optional config like API version etc..?
+}
+
+// IndexSpec defines Splunk indexes configuration
+type IndexSpec struct {
+	Indexname      string `json:"indexName"`
```
mikedickey, line 89: Same here -- I think just "Name"/"name" would be better

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
```diff
@@ -88,6 +94,9 @@ type IndexerClusterStatus struct {
 
 	// status of each indexer cluster peer
 	Peers []IndexerClusterMemberStatus `json:"peers"`
+
+	// Splunk Smart store configuration
+	SmartStore SmartStoreSpec `json:"smartstore,ommitempty"`
```
mikedickey, line 99: think you have an extra "m" here: "omitempty"

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
```diff
@@ -55,6 +58,9 @@ type IndexerClusterMemberStatus struct {
 
 	// Flag indicating if this peer belongs to the current committed generation and is searchable.
 	Searchable bool `json:"is_searchable"`
+
+	// Splunk Smart store configuration
+	SmartStore SmartStoreSpec `json:"smartstore,ommitempty"`
```
mikedickey, line 63: why do we need this in member status?

pkg/splunk/enterprise/configuration.go
```diff
@@ -606,3 +609,63 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
 		podTemplateSpec.Spec.Containers[idx].Env = env
 	}
 }
+
+// ListSmartStoreVolumes lists smartstore volumes
```
mikedickey, line 613: I think "Log*" may be a better name for these vs more generic "List*", to help clarify the expected outcome

pkg/splunk/enterprise/configuration.go
```diff
@@ -606,3 +609,63 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
+func ListSmartStoreIndexes(indexList []enterprisev1.IndexSpec) {
+	scopedLog := logC.WithName("ListSmartStoreIndexes")
+	for _, index := range indexList {
+		scopedLog.Info("Index: ", "IndexName: ", index.Indexname, "RemoteLocation: ", index.RemoteLocation)
+	}
+}
+
+// isSmartstoreEnabled checks and returns true if smartstore is configured
+func isSmartstoreConfigured(smartstore *enterprisev1.SmartStoreSpec) bool {
+	return !(smartstore.IndexList == nil && smartstore.S3VolList == nil)
```
mikedickey, line 632: what if smartstore == nil?

pkg/splunk/enterprise/configuration.go
```diff
@@ -606,3 +609,63 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
+}
+
+// ValidateSplunkSmartstoreSpec checks and validates the smartstore config
+func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error {
+	if !isSmartstoreConfigured(smartstore) {
+		return nil
+	}
+	numVolumes := len(smartstore.S3VolList)
+	if numVolumes <= 0 {
+		return fmt.Errorf("Smartstore Volume configueration is missing")
```
mikedickey, line 642: typo configuration

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
sgontla: Thanks Mike. Yesterday, we were discussing this.  Actually, I didn't do this change, and it was auto-generated when i generated the CRDs with the smartstore config. I will revert it.

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
sgontla: Thanks for this nice catch Mike. I am thinking if this is the reason for the CircleCI failure. (Wondering, if it is not supposed to be caught by the compiler)

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
sgontla: I thought we can use this to check if there is a change in the smartstore config from last reconcile iteration to the now, so that, we can create a new config map. 
Please correct me, if my understanding is wrong in using this way, and if there needs to be a better alternative way.

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
mikedickey: Understood. Thought I added something to build/package.sh so that it would add this back into each of the CRD yamls, but looks like I did not. Would you be able to look into that? I think it'll just take a few extra "yq" commands, and then we don't have to worry about it getting clobbered again (we'll need to update it for each version bump, though).

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
mikedickey: Ahh ok, that makes sense. Generally, it's better to compare to the state of the output. For example, most things we compare to the current spec of the corresponding StatefulSet objects that are produced. In this case, I imagine the "output" would be the ConfigMap that is the app for indexes.conf.

Comparing to indexes.conf seems overly cumbersome. An alternative may be to store this as an extra field in the ConfigMap. The approach of storing it in status is probably fine, though.

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
sgontla: Sure Mike, i will take a look at it

pkg/apis/enterprise/v1alpha3/common_types.go
sgontla: Sorry, i missed to reply earlier. No, make lint is not failing.

pkg/splunk/enterprise/configuration.go
sgontla: > what if smartstore == nil?

I wanted to verify through debugger once to be sure before I reply. I just checked, and since SmartStore is a struct object, eve if the smartstore spec is not configured, it will not become nil, but the values it holds can be nil.

sgontla: @mikedickey  Verified through the debugger, and smartstore never becomes nil even if we don't specify in the CR. SmartStore is a struct object(not dynamically allocated) in the  StandaloneSpec and IndexerClusterSpec data structures.

pkg/apis/enterprise/v1alpha3/indexercluster_types.go
sgontla: Will evaluate and take a direction as part of CSPL-348, which is in progress for createing the configmap.

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
```diff
@@ -1002,6 +1002,38 @@ spec:
...
+                        type: string
+                      name:
+                        description: Splunk data index name
+                        type: string
+                    type: object
+                  type: array
+                volumes:
+                  description: List to accommodate multiple volumes. For now, a hard
+                    check is added such that only one volume is supported to accommodate
+                    one common secrete object
```
smohan-splunk, line 1024: Nitpick - typo with secret

pkg/splunk/enterprise/configuration.go
```diff
@@ -639,3 +642,69 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
+	}
+
+	return !(smartstore.IndexList == nil && smartstore.VolList == nil)
+}
+
+// ValidateSplunkSmartstoreSpec checks and validates the smartstore config
+func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error {
+	// Smartstore is an optional config (at least) for now
+	if !isSmartstoreConfigured(smartstore) {
+		return nil
```
gaurav-splunk, line 676: Its better to log an error msg saying something like  - "Smartstore is not configured" or something along those lines.

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -322,12 +329,23 @@ func getClusterMasterStatefulSet(cr *enterprisev1.IndexerCluster) (*appsv1.State
...
+
 func validateIndexerClusterSpec(cr *enterprisev1.IndexerCluster) error {
 	// IndexerCluster must support 0 replicas for multisite / multipart clusters
 	// Multisite / multipart clusters: can't reference a cluster master located in another namespace because of Service and Secret limitations
 	if len(cr.Spec.IndexerClusterRef.Namespace) > 0 && cr.Spec.IndexerClusterRef.Namespace != cr.GetNamespace() {
 		return fmt.Errorf("Multisite cluster does not support cluster master to be located in a different namespace")
 	}
+
+	if cr.Spec.IndexerClusterRef.Name == "" {
+		err := ValidateSplunkSmartstoreSpec(&cr.Spec.SmartStore)
```
gaurav-splunk, line 341: if `smartstore==null`, `ValidateSplunkSmartstoreSpec` should return an error and we should return from here, right?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -322,12 +329,23 @@ func getClusterMasterStatefulSet(cr *enterprisev1.IndexerCluster) (*appsv1.State
...
+		if err != nil {
+			return err
+		}
+	} else if isSmartstoreConfigured(&cr.Spec.SmartStore) {
+		return (fmt.Errorf("Smartstore configuration is not supporeted on this site. Smartstore can be configured only on site with CM"))
```
gaurav-splunk, line 346: **Comment 1.** 
Nit: s/supporeted/supported
**Comment 2.**  
I don't think so we should mention the word "site" here as it might confuse people with multi-site feature. Maybe use node or pod or resource?

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
 		}
 	}
 }
+
+func TestSmartstoreApplyIndexerClusterFailsOnInvalidSmartStoreConfi(t *testing.T) {
```
gaurav-splunk, line 240: nit: s/TestSmartstoreApplyIndexerClusterFailsOnInvalidSmartStoreConfi/TestSmartstoreApplyIndexerClusterFailsOnInvalidSmartStoreConfig

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+	cr := enterprisev1.IndexerCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "standalone",
```
gaurav-splunk, line 243: Do we want to change the name from "standalone" to something else, since this is not exactly testing smartstore on standalone instance, right?

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+			Namespace: "test",
+		},
+		Spec: enterprisev1.IndexerClusterSpec{
+			Replicas: 1,
+			SmartStore: enterprisev1.SmartStoreSpec{
+				VolList: []enterprisev1.VolumeSpec{
+					{Name: "msos_s2s3_vol", S3Endpoint: ""},
```
gaurav-splunk, line 250: So this test case should fail because `S3Endpoint: ""` ?

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+
+	var client splcommon.ControllerClient
+
+	_, err := ApplyIndexerCluster(client, &cr)
+	if err == nil {
+		t.Errorf("ApplyIndexerCluster should fail on invalid smartstore config")
+	}
+}
+
+func TestSmartstoreApplyStandaloneFailsOnInvalidSmartStoreConfi(t *testing.T) {
```
gaurav-splunk, line 270: nit: s/TestSmartstoreApplyStandaloneFailsOnInvalidSmartStoreConfi/TestSmartstoreApplyStandaloneFailsOnInvalidSmartStoreConfig

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+
+	var client splcommon.ControllerClient
+
+	_, err := ApplyStandalone(client, &cr)
+	if err == nil {
+		t.Errorf("ApplyStandalone should fail on invalid smartstore config")
+	}
+}
+
+func TestSmartStoreConfigDoNotFailsOnSitesWithCM(t *testing.T) {
```
gaurav-splunk, line 300: Please do not use "site" if its not related to multi-site env.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+	cr := enterprisev1.IndexerCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "idx_CM_site",
```
gaurav-splunk, line 303: ditto, same as above!

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+	}
+
+	err := validateIndexerClusterSpec(&cr)
+
+	if err != nil {
+		t.Errorf("Smartstore configuration is Invalid: %v", err)
+	}
+}
+
+func TestSmartStoreConfigFailsOnSitesWithNoCM(t *testing.T) {
```
gaurav-splunk, line 329: ditto: don't use "site" or "sites"

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+	cr := enterprisev1.IndexerCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "idexer_site_1",
```
gaurav-splunk, line 332: ditto:same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+				IndexList: []enterprisev1.IndexSpec{
+					{Name: "salesdata1", RemoteLocation: "testbucket-rs-london/$_index_name"},
+					{Name: "salesdata2", RemoteLocation: "testbucket-rs-london/$_index_name"},
+					{Name: "salesdata3", RemoteLocation: "testbucket-rs-london/$_index_name"},
+				},
+			},
+		},
+	}
+
+	cr.Spec.IndexerClusterRef.Name = "testCMSiteName"
```
gaurav-splunk, line 351: ditto:same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+
+	err := validateIndexerClusterSpec(&cr)
+
+	if err == nil {
+		t.Errorf("Non CM site should not allow Smartstore configuration")
```
gaurav-splunk, line 356: ditto:same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -236,3 +236,229 @@ func TestSetVolumeDefault(t *testing.T) {
...
+	}
+}
+
+func TestValidateSplunkSmartstoreSpec(t *testing.T) {
```
gaurav-splunk, line 360: You might want to also test if `smartstore=null`.

pkg/apis/enterprise/v1alpha3/common_types.go
```diff
@@ -67,3 +67,30 @@ type CommonSplunkSpec struct {
...
+	IndexList []IndexSpec `json:"indexes,omitempty"`
+}
+
+// VolumeSpec defines remote volume name and remote volume URI
+type VolumeSpec struct {
+	// Remote volume name
+	Name string `json:"name"`
+
+	// Remote volume URI
+	S3Endpoint string `json:"endpoint"`
```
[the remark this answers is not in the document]
sgontla: Currently, the config we are supporting is S3 specific, and when we start supporting multiple storages, we may need to name it based on that. One way may be introducing a type of field in addition. We can take a call based on the need at that time?

pkg/splunk/enterprise/configuration.go
```diff
@@ -639,3 +642,69 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
 		podTemplateSpec.Spec.Containers[idx].Env = env
 	}
 }
+
+// LogSmartStoreVolumes logs smartstore volumes
+func LogSmartStoreVolumes(volumeList []enterprisev1.VolumeSpec) {
+	scopedLog := logC.WithName("LogtSmartStoreVolumes")
+	//var temp string
+	for _, volume := range volumeList {
+		scopedLog.Info("Volume: ", "name: ", volume.Name, "endpoint: ", volume.S3Endpoint)
```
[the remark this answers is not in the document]
sgontla: This is correct according to the log format: "Info(msg string, keysAndValues ...interface{})"

pkg/splunk/enterprise/configuration.go
```diff
@@ -639,3 +642,69 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
+	}
+}
+
+// LogSmartStoreIndexes logs smartstore indexes
+func LogSmartStoreIndexes(indexList []enterprisev1.IndexSpec) {
+	scopedLog := logC.WithName("LogSmartStoreIndexes")
+	for _, index := range indexList {
+		scopedLog.Info("Index: ", "name: ", index.Name, "location: ", index.RemoteLocation)
```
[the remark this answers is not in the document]
sgontla: This is correct according to the log format: "Info(msg string, keysAndValues ...interface{})"

pkg/splunk/enterprise/configuration.go
sgontla: This is not an error, as not configuring the smart store is perfectly fine(at least for this release)

pkg/splunk/enterprise/indexercluster.go
sgontla: Smartstore is optional for this release.

pkg/splunk/enterprise/configuration_test.go
sgontla: All these changes are assuming the multi site approach that is what we have today.

pkg/splunk/enterprise/configuration_test.go
sgontla: same as above: All these changes are assuming the multi site approach that is what we have today.

pkg/splunk/enterprise/configuration_test.go
sgontla: same as above

pkg/splunk/enterprise/configuration.go
gaurav-splunk: Ok, makes sense

pkg/splunk/enterprise/configuration_test.go
gaurav-splunk: I think you meant multi-part?

pkg/splunk/enterprise/configuration_test.go
gaurav-splunk: I think you meant multi-part?

pkg/apis/enterprise/v1alpha3/common_types.go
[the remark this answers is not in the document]
sgontla: Actually, we can remove the S3 like how we removed from Spec, that way we don't need to worry later...

pkg/splunk/enterprise/configuration_test.go
sgontla: c&p error. Let me fix that.

pkg/splunk/enterprise/configuration_test.go
sgontla: Yes, in this case, its S3 end point. Point is to have some invaid smartstore config. Its tough to get in to all the combinations.

PR #109, opened 2020-08-25
Added versioned statefulSet based secret handling

akondur:
- Util APIs to create versioned secrets and test cases for the same.
- Creation and adoption of versioned statefulSetbased secret on PodTemplateSpec
- Fixed SPLUNK_DECLARATIVE_ADMIN_PASSWORD typo.
- Changed name of common object from "splunk-secrets" to "splunk-<namescope>-secret"
- Deleted old secret handling code - GetSplunkSecretsName, getSplunkSecrets APIs and secretTemplateStr 
- Updated older unit tests to accommodate changes and improve coverage.
- Set owner references for versioned secret objects created for easy cleanup.


pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
 // kubernetes logger used by splunk.enterprise package
 var log = logf.Log.WithName("splunk.enterprise")
 
-// ApplyCommonSecretObject creates/updates the namespace scoped "splunk-secrets" K8S secret object
-func ApplyCommonSecretObject(client splcommon.ControllerClient, namespace string) (*corev1.Secret, error) {
+// GetVersionedSecretName returns a versioned secret name
+func GetVersionedSecretName(versionedSecretIdentifier string, version string) string {
+	return fmt.Sprintf("%s-secret-v%s", versionedSecretIdentifier, version)
```
gaurav-splunk, line 42: How about assigning "%s-secret-v%s" also a name in names.go ?

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+	if err != nil || len(secretList.Items) == 0 {
+		scopedLog.Error(err, "Secrets not found in namespace")
+		return nil, -1, false
+	}
+
+	// secretFound is true when we find atleast one versionedSecretIdentifier based secret AND we can extract its version successfully
+	var secretFound bool = false
+
+	// existingLatestVersion holds the version number of the latest versionedSecretIdentifier based secret, if atleast one exists(defaults to -1)
+	var existingLatestVersion int = -1
```
gaurav-splunk, line 114: nit: should the default be 0(since you are expecting the versions to be >0 ) ?

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+
+	// existingLatestVersionedSecret holds the latest versionedSecretIdentifier based secret, if atleast one exists
+	var existingLatestVersionedSecret corev1.Secret
+
+	// Loop through all secrets in K8S cluster
+	for _, secret := range secretList.Items {
```
gaurav-splunk, line 120: If only we can sort the secret list based on the version number, then we won't need to loop through all the versioned secrets and get the latest versioned secret. OR if we can only keep the latest version secret only and delete all the previous versions, then we can just get the current latest version and update it.

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+
+// GetSplunkReadableNamespaceScopedSecretData retreives the namespace scoped secret's data and converts it into Splunk readable format if possible
+func GetSplunkReadableNamespaceScopedSecretData(c splcommon.ControllerClient, namespace string) (map[string][]byte, error) {
+	// Get namespace scoped secret
+	namespaceScopedSecret, err := GetNamespaceScopedSecret(c, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	secretTokenTypes := []string{"hec_token", "password", "pass4SymmKey", "idxc_secret", "shc_secret"}
```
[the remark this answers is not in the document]
akondur: Done

pkg/splunk/enterprise/names.go
```diff
@@ -51,7 +48,22 @@ const (
...
 
 	// namespace scoped secret name
-	commonSecretName = "splunk-secrets"
+	namespaceScopedSecretName = "splunk-%s-secret"
+
+	// firstVersion secret
+	firstVersion = "1"
+
+	// Less than zero version error
+	lessThanOrEqualToZeroVersionError = "Versions shouldn't be <= 0"
```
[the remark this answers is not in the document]
akondur: Done

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+	for _, tokenType := range secretTokenTypes {
+		if _, ok := namespaceScopedSecret.Data[tokenType]; !ok {
+			return nil, fmt.Errorf(missingTokenError, tokenType)
+		}
+	}
+
+	// Prepare versionedSecretIdentifier based secret data
+	return map[string][]byte{
+		"hec_token":    namespaceScopedSecret.Data["hec_token"],
```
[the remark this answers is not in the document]
akondur: Here I have used a common API for individual token types but for default.yml used the raw values. I could use a map of map i.e map[token_type] = token_type to ensure I can access all token_type strings at one go but it might be an overkill.

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+    hec_enableSSL: 0
+    hec_token: "%s"
+    password: "%s"
+    pass4SymmKey: "%s"
+    idxc:
+        secret: "%s"
+    shc:
+        secret: "%s"
+`,
+			namespaceScopedSecret.Data["hec_token"],
```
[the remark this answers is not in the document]
akondur: Same comment as above.

pkg/splunk/enterprise/util.go
```diff
@@ -34,15 +37,244 @@ import (
...
+			existingLatestVersionedSecret = secret
+		}
+		secretFound = true
+	}
+
+	return &existingLatestVersionedSecret, existingLatestVersion, secretFound
+}
+
+// GetLatestVersionedSecret is used to create/retreive latest versionedSecretIdentifier based secret, cr is optional for owner references(pass nil if not required)
+func GetLatestVersionedSecret(c splcommon.ControllerClient, cr splcommon.MetaObject, namespace string, versionedSecretIdentifier string) (*corev1.Secret, error) {
```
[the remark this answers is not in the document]
akondur: Versioned secret generation is a generic util which can be used by any application/entity and hence I kept it as versionedSecretIdentifier. StatefulSet based versioned secrets is one of the users of the util.

pkg/splunk/enterprise/util.go
akondur: I kept it as an API to avoid fmt.Sprintf's everywhere with the names.go approach.

pkg/splunk/enterprise/util.go
akondur: I used -1 vs 0 to explicitly show that it was an INVALID return type. Let me know if 0 is better I can change it back to that.

pkg/splunk/enterprise/util.go
akondur: **Deletion of secrets:**
We are handling deletion of secrets separately via a different mechanism. 


**Sorting:**
corev1.SecretList{} doesn't support indexing each entry and thus sorting via a particular field would not be possible without transferring it to a different slice of strings with just names. Since we will take care of deletion of older secrets the operation of running through should be relatively quick. Please let me know your thoughts/alternate approaches to sorting if any.

Note: After the sort, we would need to still search through the sorted list to find the first secret which includes the versionedSecretIdentifier.

pkg/splunk/enterprise/util.go
gaurav-splunk: Other names such as `defaultsTemplateStr`, `serviceTemplateStr` follow the pattern which I mentioned, so I think its better to follow the same format which has been implemented already, just for the sake of consistency.

pkg/splunk/enterprise/util.go
gaurav-splunk: Its ok, -1 is also fine :)

pkg/splunk/enterprise/util.go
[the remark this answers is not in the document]
gaurav-splunk: Oh is it? I thought you are creating statefulsets specific secrets and that's why caller of this function returns statefulsets based secrets, isn't it? Please correct me if I am wrong!

pkg/splunk/enterprise/messages.go
```diff
@@ -0,0 +1,15 @@
+package enterprise
```
gaurav-splunk, line 1: Thank you :) much better now! 👍

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -33,16 +34,24 @@ import (
 
 func TestApplyIndexerCluster(t *testing.T) {
 	funcCalls := []spltest.MockFuncCall{
-		{MetaName: "*v1.Secret-test-splunk-secrets"},
-		{MetaName: "*v1.Secret-test-splunk-stack1-indexer-secrets"},
+		{MetaName: fmt.Sprintf("*v1.Secret-test-%s", GetNamespaceScopedSecretName("test"))},
```
mikedickey, line 37: You may want to be more explicit with these names, to also ensure that the methods `GetNamespaceScopedSecretName` and `GetVersionedSecretName` are returning the expected values

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -353,27 +374,33 @@ func TestGetIndexerStatefulSet(t *testing.T) {
...
...ce","operator":"In","values":["splunk-stack1-indexer"]}]},"topologyKey":"kubernetes.io/hostname"}}]}},"schedulerName":"default-scheduler"}},"volumeClaimTemplates":[{"metadata":{"name":"pvc-etc","namespace":"test","creationTimestamp":null,"labels":{"app.kubernetes.io/component":"indexer","app.kubernetes.io/instance":"splunk-stack1-indexer","app.kubernetes.io/managed-by":"splunk-operator","app.kubernetes.io/name":"indexer","app.kubernetes.io/part-of":"splunk-stack1-indexer"}},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"10Gi"}}},"status":{}},{"metadata":{"name":"pvc-var","namespace":"test","creationTimestamp":null,"labels":{"app.kubernetes.io/component":"indexer","app.kubernetes.io/instance":"splunk-stack1-indexer","app.kubernetes.io/managed-by":"splunk-operator","app.kubernetes.io/name":"indexer","app.kubernetes.io/part-of":"splunk-stack1-indexer"}},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"100Gi"}}},"status":{}}],"serviceName":"splunk-stack1-indexer-headless","podManagementPolicy":"Parallel","updateStrategy":{"type":"OnDelete"}},"status":{"replicas":0}}`, GetVersionedSecretName("splunk-stack1-indexer", firstVersion)))
```
mikedickey, line 394: same consideration: comparison to explicit value expected, versus using method to calculate it

pkg/splunk/enterprise/util.go
```diff
@@ -34,20 +37,248 @@ import (
...
+	scopedLog := log.WithName("GetExistingLatestVersionedSecret").WithValues(
+		"versionedSecretIdentifier", versionedSecretIdentifier,
+		"namespace", namespace)
+
+	// Get list of secrets in K8S cluster
+	secretList := corev1.SecretList{}
+
+	// Retreive only secrets only from namespace
+	listOpts := []client.ListOption{
+		client.InNamespace(namespace),
```
mikedickey, line 101: Just a thought is I wonder if any more criteria could be safely added to this query filter, such as owner is splunk operator?

pkg/splunk/enterprise/util.go
```diff
@@ -34,20 +37,248 @@ import (
...
+	}
+
+	err := c.List(context.TODO(), &secretList, listOpts...)
+	if err != nil || len(secretList.Items) == 0 {
+		scopedLog.Error(err, "Secrets not found in namespace")
+		return nil, -1, false
+	}
+
+	// secretFound is true when we find atleast one versionedSecretIdentifier based secret AND we can extract its version successfully
+	var secretFound bool = false
```
mikedickey, line 111: is this necessary? could we just use `existingLatestVersion != -1` ?

mikedickey: and subsequently (for return values) `existingLatestVersionedSecret != null`

pkg/splunk/enterprise/util.go
```diff
@@ -34,20 +37,248 @@ import (
...
+	}
+
+	return &existingLatestVersionedSecret, existingLatestVersion, secretFound
+}
+
+// GetLatestVersionedSecret is used to create/retreive latest versionedSecretIdentifier based secret, cr is optional for owner references(pass nil if not required)
+func GetLatestVersionedSecret(c splcommon.ControllerClient, cr splcommon.MetaObject, namespace string, versionedSecretIdentifier string) (*corev1.Secret, error) {
+	var latestVersionedSecret *corev1.Secret
+	var err error
+	var secretFound bool
```
mikedickey, line 144: similar as above, i wonder if we need this extra bool?

pkg/splunk/enterprise/util.go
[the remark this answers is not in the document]
akondur: StatefulSet are using the versioned secret utility to created versioned secrets. Similarly any other component can do the same too based on a unique versionedSecretIdentifier

pkg/splunk/enterprise/indexercluster_test.go
akondur: Done. Changed it back to test GetNamespaceScopedSecretName and GetVersionedSecretName APIs

pkg/splunk/enterprise/indexercluster_test.go
akondur: Done

pkg/splunk/enterprise/util.go
akondur: As per our discussion, list options doesn't have such a utility [reflink](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#ListOption).

pkg/splunk/enterprise/util.go
akondur: Done. Removed the bool.

pkg/splunk/enterprise/util.go
akondur: Done. Removed the bool.

pkg/splunk/enterprise/util.go
akondur: Done

pkg/splunk/enterprise/secrets.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Retreive only secrets only from namespace
+	listOpts := []client.ListOption{
+		client.InNamespace(namespace),
+	}
+
+	err := c.List(context.TODO(), &secretList, listOpts...)
+	if err != nil || len(secretList.Items) == 0 {
+		scopedLog.Error(err, "Secrets not found in namespace")
+		return nil, -1
```
sgontla, line 78: Nit: can be existingLatestVersion

pkg/splunk/enterprise/secrets.go
```diff
@@ -0,0 +1,293 @@
...
+// ApplySplunkSecret creates/updates a secret using secretData(which HAS to be of ansible readable format) or namespace scoped secret data if not specified
+func ApplySplunkSecret(c splcommon.ControllerClient, cr splcommon.MetaObject, secretData map[string][]byte, secretName string, namespace string) (*corev1.Secret, error) {
+	var current corev1.Secret
+	var newSecretData map[string][]byte
+	var err error
+
+	// Prepare secret data
+	if secretData != nil {
+		newSecretData = secretData
+	} else {
```
sgontla, line 192: Curious, when does this happen?

pkg/splunk/enterprise/secrets.go
akondur: I thought of avoiding creation of a variable existingLatestVersion if there is no secret in the namespace to save memory?

pkg/splunk/enterprise/secrets.go
akondur: ApplySplunkSecret can be used by any application to create/update a secret(with name <secretName>) whose data will be derived from the namespace scoped secret if secretData is passed as nil. It makes is easy when someone wants to use the namespace scoped secret's data, transform it to ansible readable format and create/update secret based on the same. The intention is to make the caller agnostic of secret related operations.

PR #114, opened 2020-08-31
Fix CSPL_300 - update secret scenario and added unit testing

akondur:
Fixing https://jira.splunk.com/browse/CSPL-300. Fixed update scenario + Unit testing.

Added a copyright message also in a couple of newly created files.

### DIFF
--- a/pkg/splunk/controller/messages.go
+++ b/pkg/splunk/controller/messages.go
@@ -12,18 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package enterprise
+package controller
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/splunk/controller/messages.go
```diff
@@ -0,0 +1,5 @@
+package controller
```
sgontla, line 1: Is this needed?

akondur: Yes it is required. Every go file must be a part of a package.

sgontla: NVM, i just noticed its a new file.

PR #116, opened 2020-09-03
Adding monitoring console. On implementation of Standalone/LicenseMas…

kashok-splunk:
[CSPL-383] Adding a monitoring console. On implementation of Standalone/LicenseMaster/SearchHeadCluster custom resource, a Splunk monitoring console deployment will be deployed to monitor all Splunk instance in that namespace

### DIFF
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -121,6 +121,10 @@ func getSplunkService(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSp
 	service.ObjectMeta.Namespace = cr.GetNamespace()
 	instanceIdentifier := cr.GetName()
 	var partOfIdentifier string
+	if instanceType == SplunkMonitoringConsole {
+		service.ObjectMeta.Name = GetSplunkServiceName(instanceType, cr.GetNamespace(), isHeadless)
+		instanceIdentifier = cr.GetNamespace()
+	}
 	if instanceType == SplunkIndexer {
 		if len(spec.IndexerClusterRef.Name) == 0 {
 			// Do not specify the instance label in the selector of IndexerCluster services, so that the services of the main part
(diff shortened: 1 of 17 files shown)


pkg/splunk/enterprise/types.go
```diff
@@ -58,6 +61,8 @@ func (instanceType InstanceType) ToRole() string {
 		role = "splunk_deployer"
 	case SplunkLicenseMaster:
 		role = "splunk_license_master"
+	case SplunkMonitoringConsole:
+		role = "splunk_monitor"
```
sgontla, line 65: How about naming it as "splunk_monitoring_conosole" to be in line with others?

pkg/splunk/enterprise/configuration.go
```diff
@@ -247,6 +251,12 @@ func getSplunkPorts(instanceType InstanceType) map[string]int {
 	}
 
 	switch instanceType {
+	case SplunkMonitoringConsole:
+		result["dfccontrol"] = 17000
```
sgontla, line 255: Do we need DFS ports on MC?

pkg/splunk/enterprise/types.go
kashok-splunk: This is as per the naming convention in "ansible playbooks". Splunk monitoring console has "splunk_monitor" role

pkg/splunk/enterprise/configuration.go
kashok-splunk: we can actually avoid this here since, in theory, it is like a search head but we will not be running DFS searches/service on it

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,348 @@
...
+	var err error
+
+	secrets, err = GetLatestVersionedSecret(client, cr, cr.GetNamespace(), GetSplunkMonitoringConsoleDeploymentName(SplunkMonitoringConsole, cr.GetNamespace()))
+
+	secretName := ""
+	if secrets != nil {
+		secretName = secrets.GetName()
+	}
+
+	if err != nil {
```
sgontla, line 49: Nit: I think the error has to go above

PR #117, opened 2020-09-03
Fix CSPL_450 to address empty secret data map

akondur:
The data in the namespace scoped secret on deletion of all tokens reverts to a nil map. In such a case making sure that we are creating an empty map to avoid a nil pointer reference(hence crash on operator).

Fixing https://jira.splunk.com/browse/CSPL-450

### DIFF
--- a/pkg/splunk/enterprise/secrets.go
+++ b/pkg/splunk/enterprise/secrets.go
@@ -240,7 +240,12 @@ func ApplyNamespaceScopedSecretObject(client splcommon.ControllerClient, namespa
 	namespacedName := types.NamespacedName{Namespace: namespace, Name: GetNamespaceScopedSecretName(namespace)}
 	err := client.Get(context.TODO(), namespacedName, &current)
 	if err == nil {
-		// Found, generate values for only missing types of tokens them
+		// Found, if data is non-existent, create a map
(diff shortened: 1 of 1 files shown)


smohan-splunk (approved):

sgontla (approved):

pkg/splunk/enterprise/secrets.go
```diff
@@ -240,7 +240,12 @@ func ApplyNamespaceScopedSecretObject(client splcommon.ControllerClient, namespa
 	namespacedName := types.NamespacedName{Namespace: namespace, Name: GetNamespaceScopedSecretName(namespace)}
 	err := client.Get(context.TODO(), namespacedName, &current)
 	if err == nil {
-		// Found, generate values for only missing types of tokens them
+		// Found, if data is non-existent, create a map
+		if current.Data == nil {
+			current.Data = make(map[string][]byte)
```
gaurav-splunk, line 245: What if we move this line to line 238 and remove this `if` statement? It should work right or Get will still make it null if it doesn't find anything ?

akondur: Addressed

gaurav-splunk (approved):

PR #120, opened 2020-09-04
CSPL-383

kashok-splunk:
[CSPL-383] Adding a monitoring console. On implementation of Standalone/LicenseMaster/SearchHeadCluster custom resource, a Splunk monitoring console deployment will be deployed to monitor all Splunk instance in that namespace"

### DIFF
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -490,7 +490,7 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec, 
 	// Explicitly set the default value here so we can compare for changes correctly with current statefulset.
 	configMapVolDefaultMode := int32(corev1.ConfigMapVolumeSourceDefaultMode)
 
-	// add inline defaults to all splunk containers
+	// add inline defaults to all splunk containers other than MC(where CR spec defaults are not needed)
 	if spec.Defaults != "" && instanceType != SplunkMonitoringConsole {
 		addSplunkVolumeToTemplate(podTemplateSpec, "defaults", corev1.VolumeSource{
 			ConfigMap: &corev1.ConfigMapVolumeSource{
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -121,6 +121,10 @@ func getSplunkService(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSp
 	service.ObjectMeta.Namespace = cr.GetNamespace()
 	instanceIdentifier := cr.GetName()
 	var partOfIdentifier string
+	if instanceType == SplunkMonitoringConsole {
+		service.ObjectMeta.Name = GetSplunkServiceName(instanceType, cr.GetNamespace(), isHeadless)
+		instanceIdentifier = cr.GetNamespace()
+	}
 	if instanceType == SplunkIndexer {
 		if len(spec.IndexerClusterRef.Name) == 0 {
 			// Do not specify the instance label in the selector of IndexerCluster services, so that the services of the main part
@@ -223,6 +227,19 @@ func getSplunkDefaults(identifier, namespace string, instanceType InstanceType, 
 	}
 }
 
+// getSplunkSmartstoreConfigMap returns a Kubernetes ConfigMap containing Splunk smartstore cofig in INI format
+func getSplunkSmartstoreConfigMap(identifier, namespace string, crKind string, smartstoreConf string) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetSplunkSmartstoreConfigMapName(identifier, crKind),
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			"indexes.conf": smartstoreConf,
+		},
+	}
+}
+
 // generateSplunkSecret returns a randomly generated Splunk secret.
 func generateSplunkSecret() []byte {
 	return splcommon.GenerateSecret(secretBytes, 24)
@@ -247,6 +264,9 @@ func getSplunkPorts(instanceType InstanceType) map[string]int {
 	}
 
 	switch instanceType {
+	case SplunkMonitoringConsole:
+		result["hec"] = 8088
+		result["s2s"] = 9997
 	case SplunkStandalone:
 		result["dfccontrol"] = 17000
 		result["datareceive"] = 19000
@@ -471,7 +491,7 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec, 
 	configMapVolDefaultMode := int32(corev1.ConfigMapVolumeSourceDefaultMode)
 
 	// add inline defaults to all splunk containers
-	if spec.Defaults != "" {
+	if spec.Defaults != "" && instanceType != SplunkMonitoringConsole {
 		addSplunkVolumeToTemplate(podTemplateSpec, "defaults", corev1.VolumeSource{
 			ConfigMap: &corev1.ConfigMapVolumeSource{
 				LocalObjectReference: corev1.LocalObjectReference{
@@ -600,7 +620,7 @@ func LogSmartStoreVolumes(volumeList []enterprisev1.VolumeSpec) {
 	scopedLog := logC.WithName("LogSmartStoreVolumes")
 	//var temp string
 	for _, volume := range volumeList {
-		scopedLog.Info("Volume: ", "name: ", volume.Name, "endpoint: ", volume.Endpoint)
+		scopedLog.Info("Volume: ", "name: ", volume.Name, "endpoint: ", volume.Endpoint, "path: ", volume.Path)
 	}
 }
 
@@ -608,7 +628,7 @@ func LogSmartStoreVolumes(volumeList []enterprisev1.VolumeSpec) {
 func LogSmartStoreIndexes(indexList []enterprisev1.IndexSpec) {
 	scopedLog := logC.WithName("LogSmartStoreIndexes")
 	for _, index := range indexList {
-		scopedLog.Info("Index: ", "name: ", index.Name, "location: ", index.RemoteLocation)
+		scopedLog.Info("Index: ", "name: ", index.Name, "remotePath: ", index.RemotePath, "volumeName", index.VolName)
 	}
 }
 
@@ -639,19 +659,32 @@ func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error
 	volume := smartstore.VolList[0]
 	// Make sure that the smartstore volume info is correct
 	if volume.Name == "" {
+		// To Do: cspl-348 sgontla: Once multiple volumes are supported, we can relax this by introducing default volume concept
 		return fmt.Errorf("Volume name is missing")
-	} else if volume.Endpoint == "" {
+	}
+
+	if volume.Endpoint == "" {
 		return fmt.Errorf("Volume Endpoint URI is missing")
+	}
+
+	if volume.Path == "" {
+		return fmt.Errorf("Volume Path is missing")
 	}
 
 	indexList := smartstore.IndexList
 	// Make sure that all the indexes are provided with the mandatory config values.
-	// Placeholder to fill any default values. All
+	// Placeholder to support any default volume.
 	for i, index := range indexList {
 		if index.Name == "" {
 			return fmt.Errorf("Index name is missing for index at: %d", i)
-		} else if index.RemoteLocation == "" {
-			return fmt.Errorf("Index RemoteLocation is missing for index at: %d", i)
+		}
+
+		if index.VolName == "" {
+			return fmt.Errorf("volumeName is missing for index: %s", index.Name)
+		}
+
+		if index.VolName != volume.Name {
+			return fmt.Errorf("Unknown volume %s configured for index: %s", index.VolName, index.Name)
 		}
 	}
 
@@ -659,4 +692,57 @@ func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error
 	LogSmartStoreVolumes(smartstore.VolList)
 	LogSmartStoreIndexes(indexList)
 	return nil
+}
+
+// GetSmartstoreVolumesConfig returns the list of Volumes configuration in INI format
+func GetSmartstoreVolumesConfig(client splcommon.ControllerClient, cr splcommon.MetaObject, smartstore *enterprisev1.SmartStoreSpec) (string, error) {
+	var (
+		volumesConf                   = ""
+		s3AccessKey, s3SecretKey, err = GetSmartstoreSecrets(client, cr, smartstore)
+	)
+
+	if err != nil {
+		return "", err
+	}
+
+	volumes := smartstore.VolList
+	for i := len(volumes) - 1; i >= 0; i-- {
+		volumesConf = fmt.Sprintf(`
+[volume:%s]
+storageType = remote
+path = s3://%s
+remote.s3.access_key = %s
+remote.s3.secret_key = %s
+remote.s3.endpoint = %s
+%s`, volumes[i].Name, volumes[i].Path, s3AccessKey, s3SecretKey, volumes[i].Endpoint, volumesConf)
+	}
+
+	return volumesConf, nil
+}
+
+// GetSmartstoreIndexesConfig returns the list of indexes configuration in INI format
+func GetSmartstoreIndexesConfig(indexes []enterprisev1.IndexSpec) string {
+
+	var (
+		indexesConf = ""
+		remotePath  string
+
+		//To Do: sgontla: Do we need to enforce $_index_name?
+		defaultRemotePath = "$_index_name"
+	)
+
+	for i := len(indexes) - 1; i >= 0; i-- {
+		if indexes[i].RemotePath != "" {
+			remotePath = indexes[i].RemotePath
+		} else {
+			remotePath = defaultRemotePath
+		}
+
+		indexesConf = fmt.Sprintf(`
+[%s]
+remotePath = volume:%s
+%s`, indexes[i].Name, remotePath, indexesConf)
+	}
+
+	return indexesConf
 }
--- a/pkg/splunk/enterprise/finalizers_test.go
+++ b/pkg/splunk/enterprise/finalizers_test.go
@@ -37,6 +37,8 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 		component = "standalone"
 	case "LicenseMaster":
 		component = "license-master"
+	case "MonitoringConsole":
+		component = "monitoring-console"
 	case "SearchHeadCluster":
 		component = "search-head"
 	case "IndexerCluster":
@@ -46,7 +48,10 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 	labels := map[string]string{
 		"app.kubernetes.io/part-of": fmt.Sprintf("splunk-%s-%s", cr.GetName(), component),
 	}
-	listOpts := []client.ListOption{
+	listOptsA := []client.ListOption{
+		client.InNamespace("test"),
+	}
+	listOptsB := []client.ListOption{
 		client.InNamespace(cr.GetNamespace()),
 		client.MatchingLabels(labels),
 	}
@@ -69,12 +74,38 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 		mockCalls["Update"] = []spltest.MockFuncCall{
 			{MetaName: fmt.Sprintf("*%s.%s-%s-%s", apiVersion.Version, cr.GetObjectKind().GroupVersionKind().Kind, cr.GetNamespace(), cr.GetName())},
 		}
-		if component != "" {
+		if component != "indexer" {
 			mockCalls["Delete"] = []spltest.MockFuncCall{
 				{MetaName: "*v1.PersistentVolumeClaim-test-splunk-pvc-stack1-var"},
 			}
 			mockCalls["List"] = []spltest.MockFuncCall{
-				{ListOpts: listOpts},
+				{ListOpts: listOptsA},
+				{ListOpts: listOptsB},
+			}
+			mockCalls["Get"] = []spltest.MockFuncCall{
+				{MetaName: "*v1.Secret-test-splunk-test-secret"},
+				{MetaName: "*v1.Secret-test-splunk-test-secret"},
+				{MetaName: "*v1.Secret-test-splunk-test-monitoring-console-secret-v1"},
+				{MetaName: "*v1.Service-test-splunk-test-monitoring-console-service"},
+				{MetaName: "*v1.Service-test-splunk-test-monitoring-console-headless"},
+				{MetaName: "*v1.ConfigMap-test-splunk-test-monitoring-console"},
+				{MetaName: "*v1.ConfigMap-test-splunk-test-monitoring-console"},
+				{MetaName: "*v1.Deployment-test-splunk-test-monitoring-console"},
+			}
+			mockCalls["Create"] = []spltest.MockFuncCall{
+				{MetaName: "*v1.Secret-test-splunk-test-secret"},
+				{MetaName: "*v1.Secret-test-splunk-test-monitoring-console-secret-v1"},
+				{MetaName: "*v1.Service-test-splunk-test-monitoring-console-service"},
+				{MetaName: "*v1.Service-test-splunk-test-monitoring-console-headless"},
+				{MetaName: "*v1.ConfigMap-test-splunk-test-monitoring-console"},
+				{MetaName: "*v1.Deployment-test-splunk-test-monitoring-console"},
+			}
+		} else if component == "indexer" {
+			mockCalls["Delete"] = []spltest.MockFuncCall{
+				{MetaName: "*v1.PersistentVolumeClaim-test-splunk-pvc-stack1-var"},
+			}
+			mockCalls["List"] = []spltest.MockFuncCall{
+				{ListOpts: listOptsB},
 			}
 		}
 	}
--- a/pkg/splunk/enterprise/finalizers_test.go
+++ b/pkg/splunk/enterprise/finalizers_test.go
@@ -37,8 +37,6 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 		component = "standalone"
 	case "LicenseMaster":
 		component = "license-master"
-	case "MonitoringConsole":
-		component = "monitoring-console"
 	case "SearchHeadCluster":
 		component = "search-head"
 	case "IndexerCluster":
--- a/pkg/splunk/enterprise/licensemaster.go
+++ b/pkg/splunk/enterprise/licensemaster.go
@@ -48,6 +48,17 @@ func ApplyLicenseMaster(client splcommon.ControllerClient, cr *enterprisev1.Lice
 		client.Status().Update(context.TODO(), cr)
 	}()
 
+	// create or update general config resources
+	_, err = ApplySplunkConfig(client, cr, cr.Spec.CommonSplunkSpec, SplunkLicenseMaster)
+	if err != nil {
+		return result, err
+	}
+
+	err = ApplyMonitoringConsole(client, cr, cr.Spec.CommonSplunkSpec, getLicenseMasterURL(cr, &cr.Spec.CommonSplunkSpec))
+	if err != nil {
+		return result, err
+	}
+
 	// check if deletion has been requested
 	if cr.ObjectMeta.DeletionTimestamp != nil {
 		terminating, err := splctrl.CheckForDeletion(cr, client)
@@ -56,12 +67,6 @@ func ApplyLicenseMaster(client splcommon.ControllerClient, cr *enterprisev1.Lice
 		} else {
 			result.Requeue = false
 		}
-		return result, err
-	}
-
-	// create or update general config resources
-	_, err = ApplySplunkConfig(client, cr, cr.Spec.CommonSplunkSpec, SplunkLicenseMaster)
-	if err != nil {
 		return result, err
 	}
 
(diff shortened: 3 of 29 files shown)


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// GetMonitoringConsoleDeployment returns a Kubernetes Deployment object for Splunk Enterprise monitoring console instance.
+func getMonitoringConsoleDeployment(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, instanceType InstanceType, replicas int32, configMapHash string, secretName string) (*appsv1.Deployment, error) {
+	ports := splcommon.SortContainerPorts(getSplunkContainerPorts(SplunkMonitoringConsole))
+	annotations := splcommon.GetIstioAnnotations(ports)
+	var partOfIdentifier string
+	selectLabels := getSplunkLabels(cr.GetNamespace(), instanceType, partOfIdentifier) //using Namespace here so that with every CR the name should remain same Ex- splunk-<namespace>-monitoring-console
+	affinity := splcommon.AppendPodAntiAffinity(&spec.Affinity, cr.GetNamespace(), instanceType.ToString())
+	configMap := GetSplunkDefaultsName(cr.GetNamespace(), SplunkMonitoringConsole)
```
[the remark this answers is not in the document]
kashok-splunk: Sure, taking care of this in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if configMapHash == "" {
```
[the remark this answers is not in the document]
kashok-splunk: This is just an extra check. Already taking care of empty value in configMapHash, Since we will always create an empty configMap so this value will never be null

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+	_, err = splctrl.ApplyDeployment(client, deploymentMC)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
+// GetMonitoringConsoleDeployment returns a Kubernetes Deployment object for Splunk Enterprise monitoring console instance.
+func getMonitoringConsoleDeployment(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, instanceType InstanceType, replicas int32, configMapHash string, secretName string) (*appsv1.Deployment, error) {
```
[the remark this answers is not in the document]
kashok-splunk: yes, thoughts are in the future we might have "GlobalMonitoringConsole" which can act as a monitoring console for all the namespaces so kept this function as generic, instead of making static with "SplunkMonitoringConsole" instanceType name as we may have something like "SplunkGlobalMonitoringConsole"

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+//ApplyMonitoringConsoleEnvConfigMap creates or updates a Kubernetes ConfigMap for extra env for monitoring console pod
+func ApplyMonitoringConsoleEnvConfigMap(client splcommon.ControllerClient, namespace string, newURLs []corev1.EnvVar, addNewURLs bool) (*corev1.ConfigMap, error) {
+
+	//1. Retrieve the existing McExtraEnvMap contents
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: GetSplunkDefaultsName(namespace, SplunkMonitoringConsole)}
+
+	var current corev1.ConfigMap
+
+	err := client.Get(context.TODO(), namespacedName, &current) //save existing in current
+	revised := current.DeepCopy()
```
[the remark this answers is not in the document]
kashok-splunk: Taking care of it in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+
+	current = corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetSplunkDefaultsName(namespace, SplunkMonitoringConsole),
+			Namespace: namespace,
+		},
+		Data: McExtraEnv,
+	}
+
+	err = splctrl.CreateResource(client, &current)
```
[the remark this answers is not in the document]
kashok-splunk: Controller "pkg" doesn't have access to instanceType to retrieve the name of configMap as in the case of "SplunkMonitoringConsole" the name of the configMap is based on the namespace and not on name, so just created resource here

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,347 @@
...
+	//now iterate through the map in sorted order
+	for _, k := range keys {
+		fmt.Fprintf(b, "%s=\"%s\"\n", k, contents[k])
+	}
+
+	if _, err := io.Copy(hash, b); err != nil {
+		return "", err
+	}
+
+	hashInBytes := hash.Sum(nil)[:16]
```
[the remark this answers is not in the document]
kashok-splunk: Let me check, and get back on this

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+		return err
+	}
+
+	addNewURLs := true
+
+	if cr.GetObjectMeta().GetDeletionTimestamp() != nil {
+		addNewURLs = false
+	}
+
+	//_, err = ApplyMonitoringConsoleEnvConfigMap(client, cr.GetNamespace(), extraEnv, addNewURLs)
```
akondur, line 70: Remove the comment?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return err
+}
+
+// GetMonitoringConsoleDeployment returns a Kubernetes Deployment object for Splunk Enterprise monitoring console instance.
+func getMonitoringConsoleDeployment(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, instanceType InstanceType, replicas int32, configMapHash string, secretName string) (*appsv1.Deployment, error) {
+	ports := splcommon.SortContainerPorts(getSplunkContainerPorts(SplunkMonitoringConsole))
+	annotations := splcommon.GetIstioAnnotations(ports)
+	var partOfIdentifier string
```
akondur, line 99: Nit: Move this up to the start of the function? Format looks better.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	selectLabels := getSplunkLabels(cr.GetNamespace(), instanceType, partOfIdentifier) //using Namespace here so that with every CR the name should remain same Ex- splunk-<namespace>-monitoring-console
```
akondur, line 100: Nit: Move the comment to the line above the code? And also add comments for ports, annotations(if explanation needed)?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: GetSplunkMonitoringconsoleConfigMapName(namespace, SplunkMonitoringConsole)}
+	var current corev1.ConfigMap
+
+	err := client.Get(context.TODO(), namespacedName, &current)
+	// if configMap doens't exist and we try to get configMap hash just return with empty string
+	if err != nil {
+		return "", err
+	}
+
+	contents := current.Data //map[string]string
```
akondur, line 317: Is this comment needed?

akondur, line 317: Also, if data is nil or an empty map we could just return here.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	b := new(bytes.Buffer)
+
+	//check key sorting. Collect keys in array and sort. Then use that sorted keys to append
+	var keys []string
+
+	for k := range contents {
+		keys = append(keys, k)
+	}
+
+	sort.Strings(keys) //sort keys here
```
akondur, line 329: Nit: Move comment to line above

pkg/splunk/test/controller.go
```diff
@@ -170,7 +170,10 @@ func (c MockClient) List(ctx context.Context, obj runtime.Object, opts ...client
 	})
 	listObj := c.ListObj
 	if listObj != nil {
-		copyMockObject(obj, listObj.(runtime.Object))
+		//if obj.GetObjectKind() == listObj.(runtime.Object).GetObjectKind() {
```
akondur, line 173: Remove comment?

pkg/splunk/test/controller.go
```diff
@@ -170,7 +170,10 @@ func (c MockClient) List(ctx context.Context, obj runtime.Object, opts ...client
...
+		if reflect.TypeOf(obj).String() == reflect.TypeOf(listObj.(runtime.Object)).String() {
```
akondur, line 174: Can we move this check to copyMockObject? That way any function which calls it with mis match of src and dst objects will be flagged with a log?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+		return err
+	}
+
+	// create or update a headless monitoring console service
+	err = splctrl.ApplyService(client, getSplunkService(cr, &spec, SplunkMonitoringConsole, true))
+	if err != nil {
+		return err
+	}
+
+	addNewURLs := true
```
akondur, line 64: Nit: Add a comment here explaining the use of addNewURLs?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	return deploymentMC, nil
+}
+
+//ApplyMonitoringConsoleEnvConfigMap creates or updates a Kubernetes ConfigMap for extra env for monitoring console pod
+func ApplyMonitoringConsoleEnvConfigMap(client splcommon.ControllerClient, namespace string, crName string, newURLs []corev1.EnvVar, addNewURLs bool) (*corev1.ConfigMap, error) {
+
+	var current corev1.ConfigMap
+	current.Data = make(map[string]string)
+
+	//1. Retrieve the existing McExtraEnvMap contents
```
akondur, line 180: Nit. Do we need the "1." ?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: GetSplunkMonitoringconsoleConfigMapName(namespace, SplunkMonitoringConsole)}
+	err := client.Get(context.TODO(), namespacedName, &current) //save existing in current
```
akondur, line 182: Nit: Move comment to line above? Remove if not needed?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+		err = splctrl.CreateResource(client, &current)
+		if err != nil {
+			return nil, err
+		}
+		return &current, nil
+	}
+
+	if err == nil {
+		revised := current.DeepCopy()
+		for _, url := range newURLs {
```
akondur, line 204: Can this for loop and the string operations inside be divided into functions? To make it clearer?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+			_, ok := revised.Data[url.Name]
+			if addNewURLs { //if new instances are coming up
```
akondur, line 206: Nit: Move comments above

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+								revised.Data[url.Name] = strings.Join(str, ",")
+							}
+						}
+					}
+				}
+			} else { //removing custom resource. Update configMap
+				//check if only dummy entery exist then no deletion can happen
+				if strings.Contains(revised.Data[url.Name], url.Value) {
+					revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], url.Value, "")
+					if strings.HasPrefix(revised.Data[url.Name], ",") {
```
akondur, line 260: The string operations seem to be repeated code. Could be moved into a different function.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,343 @@
...
+		//create a configMap with no values
+		current = corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      GetSplunkMonitoringconsoleConfigMapName(namespace, SplunkMonitoringConsole),
+				Namespace: namespace,
+			},
+			Data: make(map[string]string),
+		}
+
+		err = splctrl.CreateResource(client, &current)
```
akondur, line 195: Logic for CreateResource here could be combined with the logic below for CreateResource?

akondur (changes requested):

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Moved comment and no explanation needed for ports, etc

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Taking care of it in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: That's TODO there, will have a short story to improvise this logic

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Same as above. TODO

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: This is just an extra check, even if data is null or empty we create a hash of empty configMap

pkg/splunk/enterprise/configuration.go
```diff
@@ -260,6 +264,9 @@ func getSplunkPorts(instanceType InstanceType) map[string]int {
 	}
 
 	switch instanceType {
+	case SplunkMonitoringConsole:
+		result["hec"] = 8088
```
smohan-splunk, line 268: Just a comment, we may need to move some of these hardcoded ports to config later

pkg/splunk/enterprise/configuration.go
```diff
@@ -484,7 +491,7 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
 	configMapVolDefaultMode := int32(corev1.ConfigMapVolumeSourceDefaultMode)
 
 	// add inline defaults to all splunk containers
-	if spec.Defaults != "" {
+	if spec.Defaults != "" && instanceType != SplunkMonitoringConsole {
```
smohan-splunk, line 494: Why aren't defaults being added to MC? Can the reason be provided as part of the comment above?

pkg/splunk/enterprise/licensemaster.go
```diff
@@ -48,6 +48,17 @@ func ApplyLicenseMaster(client splcommon.ControllerClient, cr *enterprisev1.Lice
 		client.Status().Update(context.TODO(), cr)
 	}()
 
+	// create or update general config resources
+	_, err = ApplySplunkConfig(client, cr, cr.Spec.CommonSplunkSpec, SplunkLicenseMaster)
```
smohan-splunk, line 52: Why do we now reconcile before the "CheckForDeletion" check? Wouldn't it mean we are reconciling a deleted CR?

pkg/splunk/enterprise/monitoringconsole.go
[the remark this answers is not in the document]
smohan-splunk: Can the extra check be at the beginning of the function, even before deployment config is created?

pkg/splunk/enterprise/finalizers_test.go
```diff
@@ -37,6 +37,8 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 		component = "standalone"
 	case "LicenseMaster":
 		component = "license-master"
+	case "MonitoringConsole":
```
gaurav-splunk, line 40: Does Splunk Monitoring Console deployment automatically gets deleted if we delete, lets say, a SHC which also deployed MC with it?

gaurav-splunk (changes requested):

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,340 @@
...
+			current.Data[url.Name] = url.Value
+		}
+	}
+
+	current.ObjectMeta = metav1.ObjectMeta{
+		Name:      GetSplunkMonitoringconsoleConfigMapName(namespace, SplunkMonitoringConsole),
+		Namespace: namespace,
+	}
+
+	err = splctrl.CreateResource(client, &current)
```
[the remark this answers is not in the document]
kashok-splunk: This is the first time we are creating the resource here. The update part is in the above if (err == nil) case

pkg/splunk/enterprise/finalizers_test.go
kashok-splunk: This is a mistake, we actually don't need this case here. And no, irrespective of which CR creates the deployment it will not get deleted on the deletion of CR as we are not setting it as the owner of the MC pod

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,340 @@
...
+			}
+		}
+		return &current, nil
+	}
+
+	//If no configMap and deletion of CR is requested then create a empty configMap
+	if err != nil && addNewURLs == false {
+		current = corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      GetSplunkMonitoringconsoleConfigMapName(namespace, SplunkMonitoringConsole),
```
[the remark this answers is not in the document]
kashok-splunk: Sure. Will take care of it in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,340 @@
...
+			_, ok := revised.Data[url.Name]
+			//if new instances are coming up
+			if addNewURLs {
+				//key doesn't exist in the configMap, then add it
+				if !ok {
+					//if new url doesn't exist add it in configmap
+					revised.Data[url.Name] = url.Value
+				} else {
+					//TODO: For "SEARCH_HEAD_URL" need to find a better approach to update configMap, any other data structure?
+					if url.Name == "SPLUNK_SEARCH_HEAD_URL" {
```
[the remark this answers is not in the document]
kashok-splunk: Sure, During the scaling down/up of shc the "cr.ObjectMeta.DeletionTimestamp" is nill, statefulset manager checks the difference between desiredReplicas and readyReplicas and accordingly updates the stateful set. For MC to update its configMap we check it through deletion timestamp if the deletion is requested or a new resource is getting added. So in case of scaling since the deletion timestamp is always nil, MC doesn't know if we are adding instances or deleting, so first, it deletes all the URLs of that statefulset present in the configMap and then add what are the current URLs present in that statefulset as new. That's why I have separately written the case for Search head here. But I have a TODO here to improvise this logic.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	// create or update a regular monitoring console service
+	err = splctrl.ApplyService(client, getSplunkService(cr, &spec, SplunkMonitoringConsole, false))
+	if err != nil {
+		return err
+	}
+
+	// create or update a headless monitoring console service
+	err = splctrl.ApplyService(client, getSplunkService(cr, &spec, SplunkMonitoringConsole, true))
```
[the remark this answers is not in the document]
kashok-splunk: We need headless services for MC so that any pod can directly talk to MC pod

pkg/splunk/enterprise/names.go
```diff
@@ -38,6 +38,9 @@ const (
 	// identifier
 	defaultsTemplateStr = "splunk-%s-%s-defaults"
 
+	//identifier
+	monitoringconsoleTemplateStr = "splunk-%s-%s"
```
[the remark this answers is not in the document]
kashok-splunk: Same as above

pkg/splunk/enterprise/names.go
```diff
@@ -99,6 +107,11 @@ func GetSplunkDefaultsName(identifier string, instanceType InstanceType) string
 	return fmt.Sprintf(defaultsTemplateStr, identifier, instanceType.ToKind())
 }
 
+// GetSplunkMonitoringconsoleConfigMapName uses a template to name a Kubernetes ConfigMap for a SplunkEnterprise resource.
+func GetSplunkMonitoringconsoleConfigMapName(identifier string, instanceType InstanceType) string {
+	return fmt.Sprintf(monitoringconsoleTemplateStr, identifier, instanceType.ToKind())
```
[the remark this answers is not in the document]
kashok-splunk: Yes, realized this. Taking care of this in the latest commit

pkg/splunk/enterprise/standalone.go
```diff
@@ -144,3 +149,13 @@ func validateStandaloneSpec(spec *enterprisev1.StandaloneSpec) error {
 	spec.SparkImage = spark.GetSparkImage(spec.SparkImage)
 	return validateCommonSplunkSpec(&spec.CommonSplunkSpec)
 }
+
+// getStandaloneExtraEnv returns extra environment variables used by monitoring console
+func getStandaloneExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar {
```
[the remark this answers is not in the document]
kashok-splunk: Maybe I can take this in my next story after your PR is merged?

pkg/splunk/enterprise/standalone.go
```diff
@@ -65,6 +65,17 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
 		client.Status().Update(context.TODO(), cr)
 	}()
 
+	// create or update general config resources
+	_, err = ApplySplunkConfig(client, cr, cr.Spec.CommonSplunkSpec, SplunkStandalone)
```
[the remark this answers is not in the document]
kashok-splunk: ApplySplunkConfig() creates secrets that need to be set before we can call MC deployment so moved this before deletion

pkg/splunk/enterprise/licensemaster.go
kashok-splunk: To update MC configMap in the deletion case we need to call ApplyMonitoringConsole() with "cr.ObjectMeta.DeletionTimestamp != nil" , after the deployment is updated will just delete the CR as usual and return

pkg/splunk/enterprise/configuration.go
kashok-splunk: For MC we don't need to supply it with the spec defaults of the custom resource

pkg/splunk/enterprise/monitoringconsole.go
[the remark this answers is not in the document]
gaurav-splunk: And when does a pod directly talk to MC pod? I think MC is the one who communicates to other pods, isn't it? Although there is no harm in creating this service, but I just wanted to know the use-case of this service.

pkg/splunk/enterprise/monitoringconsole.go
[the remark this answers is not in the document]
gaurav-splunk: Oh ok cool, my bad I didn't see it.

pkg/splunk/enterprise/finalizers_test.go
gaurav-splunk: Ok, I think we should set the CR as owner of MC pod, just like we do it in the case of SHC where we create a deployer too with SHC members. But the deployer is also deleted when we delete SHC. From user experience, it is not a good experience if while deploying a CR we create an MC pod too but while deleting we have to delete the MC pod explicitly.

akondur (approved):
Please take care of remaining nit's as per comments. Otherwise looks ok :)

pkg/splunk/enterprise/finalizers_test.go
kashok-splunk: The idea for MC is to be always present. Any CR that comes up first just deploys the MC pod, after that as you apply new CR it just adds itself to the existing MC pod. Like deployer service is only attached to that specific shc, MC is for the entire namespace. In that way, even if cr gets deleted MC still needs to be running for that namespace. If this might create confusion among customers we can mention this flow in the release notes

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,342 @@
...
+					Affinity:      affinity,
+					Tolerations:   spec.Tolerations,
+					SchedulerName: spec.SchedulerName,
+					Containers: []corev1.Container{
+						{
+							Image:           spec.Image,
+							ImagePullPolicy: corev1.PullPolicy(spec.ImagePullPolicy),
+							Name:            "splunk",
+							Ports:           ports,
+							EnvFrom: []corev1.EnvFromSource{
```
mikedickey, line 146: Would you please explain again why we are setting env variables via a configmap, versus just embedding the key/values here within the pod spec?

pkg/splunk/test/controller.go
```diff
@@ -43,23 +43,25 @@ type MockObjectCopier func(dst, src runtime.Object) bool
...
-		*dst.(*corev1.PersistentVolumeClaim) = *src.(*corev1.PersistentVolumeClaim)
-	case *corev1.PersistentVolumeClaimList:
-		*dst.(*corev1.PersistentVolumeClaimList) = *src.(*corev1.PersistentVolumeClaimList)
-	case *corev1.Service:
-		*dst.(*corev1.Service) = *src.(*corev1.Service)
-	case *corev1.Pod:
-		*dst.(*corev1.Pod) = *src.(*corev1.Pod)
-	default:
-		return false
+	if reflect.TypeOf(dst).String() == reflect.TypeOf(src.(runtime.Object)).String() {
```
mikedickey, line 46: Is it a normal / expected condition for these types to not match? If not, why mask the error?

pkg/splunk/test/controller.go
```diff
@@ -170,7 +172,9 @@ func (c MockClient) List(ctx context.Context, obj runtime.Object, opts ...client
 	})
 	listObj := c.ListObj
 	if listObj != nil {
-		copyMockObject(obj, listObj.(runtime.Object))
+		{
```
mikedickey, line 175: Why the extra {}?

mikedickey (changes requested):

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: For MC, it needs to have URLs of all the splunk instances running in that namespace. If we just embedded the URLs directly as env variables, one CR will not have any knowledge about the other URLs from other statefulset or custom resource. So we have separate configMap where we add/delete all the URLs for that namespace

pkg/splunk/test/controller.go
kashok-splunk: Yes, we saw this failure for one of the UTs, so added check here

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,342 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetSplunkMonitoringConsoleDeploymentName(instanceType, cr.GetNamespace()),
+			Namespace: cr.GetNamespace(),
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: selectLabels,
+			},
+			Replicas: &replicas,
```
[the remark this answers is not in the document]
kashok-splunk: Sure. We can stick to static "1"

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -0,0 +1,342 @@
...
+		return nil
+	}
+
+	deploymentMC, err := getMonitoringConsoleDeployment(cr, &spec, SplunkMonitoringConsole, 1, configMapHash, secretName)
+	if err != nil {
+		return err
+	}
+
+	_, err = splctrl.ApplyDeployment(client, deploymentMC)
+	if err != nil {
```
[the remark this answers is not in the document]
kashok-splunk: Sure. Taking care of it in the latest commit

gaurav-splunk (approved):

smohan-splunk (approved):

pkg/splunk/test/controller.go
mikedickey: But in what case would it be different, other than an error in our source code/logic? If that's the only case, then I think it's better to panic versus ignore it. Otherwise, it could lead to very confusing results and make it hard-to-impossible to detect and resolve these types of errors.

kashok-splunk: In the case of deletion, we use "c.ListObj = &pvclist" for deleting PVC volume attached with custom resource. Now when we delete any custom resource we also update the configMap of mc and MC uses list{} for secrets. So in the case of "splunkDeletionTester" when the deletion is requested it also goes and call ApplyMonitoringConsole() which in turn uses list{} for secrets, so we got an error where we were trying to copy pvcList to secret.list, that's why added that extra check. And yes we could alternatively detect the different types and log an error ensuring that we are aware of the situation but avoiding a panic

pkg/splunk/enterprise/monitoringconsole.go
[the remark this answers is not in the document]
sgontla: This code is not needed, and ResourceVersion is a better alternative IMO. Otherwise, the changes look good to me.

pkg/splunk/test/controller.go
mikedickey: sorry, i just realized this was test library code versus something built into the product. makes total sense now!

mikedickey (approved):

PR #122, opened 2020-09-09
Separate CRD for ClusterMaster & change indexerClusterRef to clusterMasterRef

gaurav-splunk:
This PR targets the following changes -

Created a new CRD of type ClusterMaster. So now, to deploy an indexer cluster you need to create 2 CRs, one with type as ClusterMaster for CM and one for indexer cluster part with type as IndexerCluster.
With the introduction of new CRD for ClusterMaster, we don't need the old way of using replicas:0 to identify that this is for CM.
We have deprecated indexerClusterRef in favor of ClusterMasterRef. This is done to explicitly mention that any other resource which wants to connect to an indexer cluster will have to connect to Cluster Master part of the indexer cluster.
Moved all the SmartStore related logic to ClusterMaster code since SmartStore can only be configured through CM.
Made changes to UT and System tests accordingly.


pkg/splunk/enterprise/finalizers.go
```diff
@@ -42,7 +42,7 @@ func DeleteSplunkPvc(cr splcommon.MetaObject, c splcommon.ControllerClient) erro
 		component = "license-master"
 	case "SearchHeadCluster":
 		component = "search-head"
-	case "IndexerCluster":
+	case "IndexerCluster", "ClusterMaster":
 		component = "indexer"
```
kashok-splunk, line 46: Should we make the component as "cluster-master" here?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// create or update general config resources
+	_, err = ApplySplunkConfig(client, cr, cr.Spec.CommonSplunkSpec, SplunkIndexer)
+	if err != nil {
+		return result, err
+	}
+
+	// create or update a headless service for indexer cluster
+	err = splctrl.ApplyService(client, getSplunkService(cr, &cr.Spec.CommonSplunkSpec, SplunkIndexer, true))
```
kashok-splunk, line 96: Do we need to create SplunkIndexer service here? This can be done while creating an indexer

mikedickey: +1 seems the two services should be created by IndexerCluster CRD versus ClusterMaster.

Kind of separate topic but related to this PR: Is "IndexerCluster" still the right name?

mikedickey (changes requested):

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: This was deliberately done as part of multi-part approach. More discussion can be found here - https://github.com/splunk/splunk-operator/pull/99#discussion_r466094117.
PS: I raised the same question in that PR too, but I know now why it makes sense. :)

akondur (approved):
Looks good.

pkg/splunk/enterprise/clustermaster.go
mikedickey: oic forgot all about that.. so the cluster master's indexer service load balances across all indexers that are using the same cluster master? makes sense for regular (headed?) service, but not sure what the headless service would be for?

gaurav-splunk: hmm, ok I think we don't need it. Headless service on indexers should be enough.

pkg/splunk/enterprise/finalizers.go
gaurav-splunk: I thought about that but then I thought not to play around with it since it is based on the labels and if I change it I need to change the labels everywhere and need to change a bit of logic on CM side too. Hence for simplicity kept it like this.

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -338,38 +287,21 @@ func getIndexerStatefulSet(client splcommon.ControllerClient, cr *enterprisev1.I
...
-// getIndexerExtraEnv returns extra environment variables used by search head clusters
-func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar {
-	return []corev1.EnvVar{
-		{
-			Name:  "SPLUNK_INDEXER_URL",
-			Value: GetSplunkStatefulsetUrls(cr.GetNamespace(), SplunkIndexer, cr.GetName(), replicas, false),
-		},
+	// Multisite / multipart clusters: can't reference a cluster master located in another namespace because of Service and Secret limitations
+	if len(cr.Spec.ClusterMasterRef.Namespace) > 0 && cr.Spec.ClusterMasterRef.Namespace != cr.GetNamespace() {
+		return fmt.Errorf("Multisite cluster does not support cluster master to be located in a different namespace")
```
sgontla, line 304: AFAIK, we are filing just the name, and other fields are empty? Are we filling namespace somewhere?

sgontla (approved):

pkg/splunk/enterprise/finalizers.go
kashok-splunk: After the monitoring console changes come in, the code flow for clusterMaster will be different than the indexer, I can see it can cause some confusion there. Maybe I'll have to change this in the future

docs/CustomResources.md
```diff
@@ -4,14 +4,15 @@ The Splunk Operator provides a collection of
...
-* [SearchHeadCluster Resource Spec Parameters](#searchheadcluster-resource-spec-parameters)
-* [IndexerCluster Resource Spec Parameters](#indexercluster-resource-spec-parameters)
+- [Custom Resource Guide](#custom-resource-guide)
+  - [Metadata Parameters](#metadata-parameters)
+  - [Common Spec Parameters for All Resources](#common-spec-parameters-for-all-resources)
+  - [Common Spec Parameters for Splunk Enterprise Resources](#common-spec-parameters-for-splunk-enterprise-resources)
+  - [Spark Resource Spec Parameters](#spark-resource-spec-parameters)
+  - [LicenseMaster Resource Spec Parameters](#licensemaster-resource-spec-parameters)
+  - [Standalone Resource Spec Parameters](#standalone-resource-spec-parameters)
+  - [SearchHeadCluster Resource Spec Parameters](#searchheadcluster-resource-spec-parameters)
```
smohan-splunk, line 14: Shouldn't we add a note for the CM CR too?

docs/Examples.md
```diff
@@ -37,16 +44,31 @@ metadata:
 
 When growing, customers will typically want to first expand by upgrading
 to an [indexer cluster](https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Aboutindexesandindexers).
-The Splunk Operator makes creation of an indexer cluster as easy as creating an `IndexerCluster` resource:
+The Splunk Operator makes creation of an indexer cluster as easy as creating a `ClusterMaster` resource for Cluster Master part and an `IndexerCluster` resource for indexers part respectively:
 
+#### Cluster Master part
```
smohan-splunk, line 49: I suggest remove the word 'part'

docs/Examples.md
```diff
@@ -65,21 +87,24 @@ If you want more indexers, just update it to include a `replicas` parameter:
 
 ```yaml
 cat <<EOF | kubectl apply -f -
-apiVersion: enterprise.splunk.com/v1alpha2
+apiVersion: enterprise.splunk.com/v1alpha3
```
smohan-splunk, line 90: Generic question: wouldn't apiVersions change to beta at some point?

smohan-splunk (approved):

mikedickey (approved):

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: You can if you want to, but we are not supporting that. That's why this check :)

docs/Examples.md
gaurav-splunk: At some later point yes. Maybe we need to keep updating docs in such cases.

pkg/splunk/enterprise/indexercluster.go
sgontla: I see.. i was thinking it as a string:-)
IndexerClusterRef corev1.ObjectReference `json:"indexerClusterRef"`

docs/CustomResources.md
gaurav-splunk: Hmm I think we can. ClusterMaster CR does not have any required spec as such other than CommonSplunkSpec, but we can give an example of S2 configuration settings in the CM CR.

gaurav-splunk: Done

romain-bellanger:
Hi @gaurav-splunk,
I've reviewed the PR which looks great to me. I wanted to test it, I fetched the PR, built the image and deployed the operator, but I forgot to request our cluster admin to create the new CRD... This should be done tomorrow.
Just one comment: the definitions of the CRDs in deploy/crds/ have been reformatted. This is not really an issue, but it was easier for me to read the description of the properties split over multiple lines, than in a single very long line. I just wanted to share this feedback but it should not block merge.

gaurav-splunk:
> Hi @gaurav-splunk,
> I've reviewed the PR which looks great to me. I wanted to test it, I fetched the PR, built the image and deployed the operator, but I forgot to request our cluster admin to create the new CRD... This should be done tomorrow.
> Just one comment: the definitions of the CRDs in deploy/crds/ have been reformatted. This is not really an issue, but it was easier for me to read the description of the properties split over multiple lines, than in a single very long line. I just wanted to share this feedback but it should not block merge.

Hi @romain-bellanger Thanks for the feedback. This is an issue with the yq version I am using. I will update the PR with correct yq version installed on my machine. That should solve the indentation problem.

PR #129, opened 2020-09-24
Outdated_DoNotUse_cspl-333, cspl-445, cspl-446: propogate config map to standalone/CM pods

sgontla:
Support for:
 - Mounting the config to ../etc/apps/splunk-operator in case of standalone
 - Mounting the config to ../etc/master-aps/splunk-operator in case of CM
 - Multiple Remote end points
 - Default option
 - Additional Smartstore related config changes


pkg/splunk/enterprise/standalone_test.go
```diff
@@ -161,16 +172,74 @@ func TestApplyStandaloneWithSmartstore(t *testing.T) {
...
+	secret.Data[s3AccessKey] = []byte("changed")
+
+	_, err = splctrl.ApplySecret(client, secret)
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+
+	changed := AreRemoteVolumeKeysChanged(client, &current, SplunkStandalone, &current.Spec.SmartStore, &err)
+
+	if changed {
```
sgontla, line 240: Just noticed, checking for the wrong status. will fix this.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -388,27 +402,121 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+eviction_padding = 2048
+eviction_policy = lru
+hotlist_bloom_filter_recency_hours = 24
+hotlist_recency_secs = 86400
+max_cache_size = 20480
+max_concurrent_downloads = 6
+max_concurrent_uploads = 6
+`)
+
+	serverConfFroCacheManager := GetServerConfigEntries(&SmartStoreCacheManager)
```
gaurav-splunk, line 487: nit: %s/serverConfFroCacheManager/serverConfForCacheManager

pkg/splunk/enterprise/standalone_test.go
```diff
@@ -161,16 +172,74 @@ func TestApplyStandaloneWithSmartstore(t *testing.T) {
 		return err
 	}
 	spltest.ReconcileTesterWithoutRedundantCheck(t, "TestApplyStandalone", &current, revised, createCalls, updateCalls, reconcile, true, secret)
+}
 
-	// test deletion
```
gaurav-splunk, line 165: ditto: why we removed this?

gaurav-splunk (changes requested):

pkg/splunk/enterprise/standalone_test.go
sgontla: The short answer is we are not testing anything additional to what is already covered as part of TestApplyStandalone, when it comes to deletions/finalizers.

The existing UT Infra is scalable to only for default flows, assume that there will be no changes across the ApplyXXX() call even if it is called multiple times. Last time, to overcome the issue, I had to introduce  ReconcileTesterWithoutRedundantCheck() which will not call the given function without a change in the spec.
Good thing with ReconcileTester OR ReconcileTesterWithoutRedundantCheck, we explicitly pass the CRUD call list and their sequence of expectations when executing the same. 

Now, coming back to splunkDeletionTester(), we don't have have a way to pass the expectations explicitly to this function. And today, it's kind of hardcoded, assuming the default flow like TestApplyStandalone().

There are 3 possibilities here, that needs to be driven based on the caller expectations:

1. Default flows - Must be covered, and always predictable, across multiple calls for the same spec. and what we already have covered that.

2. Default flow + delta configuration for a specific feature under test:-  Good to cover if there is additional work to be done as part of finalizers(This is where it needs some overhaul to support this combination), at the same time, no need to do this test, if we are not having any additional handling for the feature under test as part of deletions/ finalizers.

3. Combinations of multiple features: It is not practical to support all possible cases, as we keep adding custom config. Once we have support for #2, its easy to do calls for multiple combinations of the features, in case if twe really want to test, assuming some inter-dependancy.

Here, we are falling in case #2, at the same time, we don't have any specific deletion work as part of the smartstore work. So, the existing default flow tester TestApplyStandalone() is good enough. I will be logging an enhancement to cover case #2, where special handling is needed as part of deletion work.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -177,14 +191,4 @@ func TestApplyClusterMasterWithSmartstore(t *testing.T) {
 		return err
 	}
 	spltest.ReconcileTesterWithoutRedundantCheck(t, "TestApplyClusterMasterWithSmartstore", &current, revised, createCalls, updateCalls, reconcile, true, secret)
-
-	// test deletion
```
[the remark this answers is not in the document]
sgontla: Please see the reply here at: https://github.com/splunk/splunk-operator/pull/129#discussion_r495539287

PR #130, opened 2020-09-24
Update Monitoring console for single site indexer clustering

kashok-splunk:
Update Monitoring console for indexer clustering


kashok-splunk:
Will create a new PR for both indexer clustering

PR #131, opened 2020-09-25
CSPL-465: Fix the updation of ConfigMap by adding resource version of ConfigMap as annotation in the CR

gaurav-splunk:
This PR fixes the case where if we try to update the `defaults` section in a CR thereby updating the default ConfigMap, pod should detect that change and should recycle so that it can come up with new configs. We solve this by annotating the CR with the Resource Version of the default ConfigMap. Any change in ConfigMap would lead to a new Resouce Version, thereby changing the pod template spec and hence resulting in the recycling of pod to pick up the latest changes.


pkg/splunk/enterprise/util.go
```diff
@@ -47,11 +49,34 @@ func ApplySplunkConfig(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
+// for the resource version of the ConfigMap in the pod template spec for the CR.
+func updateDefaultConfigMapAnnotations(client splcommon.ControllerClient, cr splcommon.MetaObject, configMap *corev1.ConfigMap) error {
+	namespacedName := types.NamespacedName{Namespace: configMap.GetNamespace(), Name: configMap.GetName()}
+	var current corev1.ConfigMap
+
+	err := client.Get(context.TODO(), namespacedName, &current)
+	if err != nil {
+		return err
+	}
+	annotationsMap := cr.GetObjectMeta().GetAnnotations()
```
sgontla, line 73: I am not sure if we have a reason to apply at CR object level. 
Also, I think the better place to add this is as part of updateSplunkPodTemplateWithConfig(). That way, the mount, and the annotation go together if we want to selectively apply the defaults configMap now(i see the Monitoring console not using it, at least for now) OR in the future.

pkg/splunk/enterprise/util.go
```diff
@@ -47,11 +49,34 @@ func ApplySplunkConfig(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
+	if annotationsMap != nil {
```
sgontla, line 74: Also, if the map is not allocated, we need to allocate one, and add the entry

sgontla (changes requested):

pkg/splunk/enterprise/util.go
gaurav-splunk: hmm not sure if the place would have mattered much since anyways we are appending the labels and annotations from the parent CR to the statefulset we would create or update. See `splcommon.AppendParentMeta(statefulSet.Spec.Template.GetObjectMeta(), cr.GetObjectMeta())` in `getSplunkStatefulSet`

pkg/splunk/enterprise/util.go
akondur: +1

akondur: +1

sgontla (approved):

pkg/splunk/enterprise/configuration.go
```diff
@@ -425,16 +427,35 @@ func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaOb
...
-	updateSplunkPodTemplateWithConfig(&statefulSet.Spec.Template, cr, spec, instanceType, extraEnv, statefulSetSecret.GetName())
+	updateSplunkPodTemplateWithConfig(client, &statefulSet.Spec.Template, cr, spec, instanceType, extraEnv, statefulSetSecret.GetName())
 
 	// make Splunk Enterprise object the owner
 	statefulSet.SetOwnerReferences(append(statefulSet.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
 
 	return statefulSet, nil
 }
 
+// getConfigMap gets the ConfigMap resource in a given namespace
```
akondur, line 438: Can we make this a part of pkg/splunk/controller/configmap.go since it is a generic API for configmaps?

pkg/splunk/enterprise/configuration.go
```diff
@@ -425,16 +427,35 @@ func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaOb
...
+	var configMap corev1.ConfigMap
+	err := client.Get(context.TODO(), namespacedName, &configMap)
+	if err != nil {
+		return configMap, err
+	}
+	return configMap, nil
+}
+
+// getConfigMapResourceVersion gets the Resource version of a configMap
+func getConfigMapResourceVersion(client splcommon.ControllerClient, namespacedName types.NamespacedName) (string, error) {
```
akondur, line 449: Do we need this API since we are extracting just a single field from the configMap?

pkg/splunk/enterprise/configuration.go
```diff
@@ -477,14 +498,27 @@ func updateSplunkPodTemplateWithConfig(podTemplateSpec *corev1.PodTemplateSpec,
...
 		})
+
+		scopedLog := log.WithName("updateSplunkPodTemplateWithConfig").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+		namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: configMapName}
+
+		// We will update the annotation for resource version in the pod template spec
+		// so that any change in the ConfigMap will lead to recycle of the pod.
+		configMapResourceVersion, err := getConfigMapResourceVersion(client, namespacedName)
+		if err == nil {
+			podTemplateSpec.ObjectMeta.Annotations["defaultConfigRev"] = configMapResourceVersion
```
akondur, line 518: Just checking here: Is the annotations map always created such that we are assigning a value for a random key? If yes please ignore.

akondur (changes requested):

kashok-splunk (approved):

pkg/splunk/enterprise/configuration.go
sgontla: Being it is metatdata info, and using across places, my vote is to have this API

gaurav-splunk: I am keeping this API for now. Its always good to have some APIs to get important fields.

pkg/splunk/enterprise/configuration.go
gaurav-splunk: Yes we do create it.

smohan-splunk (approved):

pkg/splunk/controller/configmap.go
```diff
@@ -49,3 +49,22 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
...
+	var configMap corev1.ConfigMap
+	err := client.Get(context.TODO(), namespacedName, &configMap)
+	if err != nil {
+		return configMap, err
+	}
+	return configMap, nil
+}
+
+// GetConfigMapResourceVersion gets the Resource version of a configMap
+func GetConfigMapResourceVersion(client splcommon.ControllerClient, namespacedName types.NamespacedName) (string, error) {
```
akondur, line 64: Can we make this an API to retrieve a specific field such as ResourceVersion, UID etc... based on a stringed parameter? If possible, that api can be used for any runtime object as we are using the namespaced name as the parameter for client.Get() anyways. Seems expensive to have an API just for resource version. We can take this up in the future as well :)

akondur (approved):

pkg/splunk/controller/configmap.go
sgontla: Its always good to have APIs for metadata. But, as we found that K8 already has the APIs, i guess, we don't have any reason to implement and maintain.

sgontla: Arjun, in the case of run-time objects, we should be still able to use the K8 APIs. Do you see any issues? I think we don't need to implement redundantly anything that is already there in K8 APIs, as it becomes redundant and additional maintenance(?).

akondur: If we already have the runtime object with us retrieved then we should use the inbuilt apis for sure. My point was if we are retrieving any runtime object as well extracting a particular field from it we can have a generic API for that so as to avoid having a get() API call in the calling code which might look unpleasant.

PR #135, opened 2020-10-02
Automate splunk monitoring console app for indexer clustering

kashok-splunk:
Automate Splunk monitoring console app for indexer clustering
- Add cluster master URL to MC configMap
- Add MC as a searchhead on CM to get indexer info
- Automate the "Apply Changes" button on the monitoring console app when IndexerCluster custom resource is applied
- For multi-site get cluster-info and append "SPLUNK_SITE" and "SPLUNK_MULTISITE_MASTER" env variables to MC

### DIFF
--- a/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
@@ -1,8 +1,13 @@
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: clustermasters.enterprise.splunk.com
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.phase
+    description: Status of cluster master
+    name: Phase
+    type: string
   group: enterprise.splunk.com
   names:
     kind: ClusterMaster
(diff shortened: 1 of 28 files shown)


smohan-splunk (approved):

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+		if key == "" {
+			break
+		} else {
+			err = c.UpdateDMCClusteringLabelGroup(key, value)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	apiResponseMCAssetTableBuild, err := c.GetMonitoringconsoleAssetTable()
```
akondur, line 721: Comments here would be nice to understand the functionality.

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+		if e == "indexer" {
+			indexerMemberList = "&member=localhost:localhost" + indexerMemberList
+		}
+		if e == "license_master" {
+			licenseMasterMemberList = licenseMasterMemberList + "&member=localhost:localhost"
+		}
+	}
+	reqBodyIndexer := indexerMemberList + "&default=true"
+	reqBodyLicenseMaster := licenseMasterMemberList + "&default=false"
+	err = c.UpdateDMCGroups("dmc_group_indexer", reqBodyIndexer)
```
akondur, line 680: If possible now or later, it would be nice to mention the format of the full string which would be passed onto the end point. Not a necessity but would add to readability of code

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+		return err
+	}
+	err = c.UpdateDMCGroups("dmc_group_license_master", reqBodyLicenseMaster)
+	if err != nil {
+		return err
+	}
+
+	clusterRoleDict := make(map[string][]string)
+	//map of Name to Roles
+	for _, e := range apiResponseMCDistributedPeers.Entry {
```
akondur, line 691: The logic here with multiple dicts seems unclear to me, could you please elaborate a little bit?

pkg/splunk/enterprise/finalizers_test.go
```diff
@@ -130,6 +142,61 @@ func splunkDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(spl
 	c.CheckCalls(t, "Testsplctrl.CheckForDeletion", mockCalls)
 }
 
+func splunkPVCDeletionTester(t *testing.T, cr splcommon.MetaObject, delete func(splcommon.MetaObject, splcommon.ControllerClient) (bool, error)) {
```
akondur, line 145: Is there a requirement for new pvc tester? Can't we tweak splunkDeletionTester? Although I feel like there might be a strong reason behind the same!

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -67,8 +67,18 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		}
 	}()
 
+	// create or update general config resources
```
akondur, line 70: Any reason we moved this up? Same comment for ClusterMaster.go

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -123,6 +126,10 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 
 	// no need to requeue if everything is ready
 	if cr.Status.Phase == splcommon.PhaseReady {
```
akondur, line 128: Are we sure that we need the MC only when the reconcile is complete? While scaling up/down?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -66,23 +79,59 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
 		return err
 	}
 
-	statefulsetMC, err := getMonitoringConsoleStatefulSet(client, cr, &spec, SplunkMonitoringConsole, secretName)
+	//by default set multisite to false
+	multiSite := "false"
+
+	//get cluster info from cluster master
+	if cr.GetObjectKind().GroupVersionKind().Kind == "ClusterMaster" && !spec.Mock {
+		mgr := monitoringConsolePodManager{cr: &cr, spec: &spec, secrets: secrets, newSplunkClient: splclient.NewSplunkClient}
```
akondur, line 87: We don't need this as we could stub out getClusterMasterClient() as a function outside of the IndexerClusterPodManager interface. The function would look like:

func (splcommon.MetaObject) getClusterMasterClient() *splclient.SplunkClient 

We could place it in a common folder which can be accessible to IDXC, CM as well as MC.

pkg/splunk/enterprise/util.go
```diff
@@ -69,6 +69,86 @@ func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar
 	}
 }
 
+// getClusterMasterExtraEnv returns extra environment variables used by indexer clusters
+func getClusterMasterExtraEnv(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec) []corev1.EnvVar {
```
akondur, line 73: Seems like we should have a single API getExtraEnv(with possible wrappers in CR specific files such as getClusterMasterExtraEnv etc.. ) i.e

getExtraEnv(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, int32 replicas) []corev1.EnvVar)

Maybe this can be taken up in the future.

akondur (changes requested):

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Sure, can we take this in future commits?

pkg/splunk/enterprise/indexercluster.go
kashok-splunk: Yes, scaling up/down will be taken here

pkg/splunk/enterprise/indexercluster.go
kashok-splunk: Same, what we did with other CRs, creation of secrets before we call ApplyMonitoringConsole() during deletion, for UTs

pkg/splunk/enterprise/finalizers_test.go
kashok-splunk: Yes, the call flow is different in both cases and cannot be merged together

pkg/splunk/client/enterprise.go
kashok-splunk: same above.

pkg/splunk/client/enterprise.go
kashok-splunk: same above

pkg/splunk/client/enterprise.go
kashok-splunk: we can take this offline, no need for too many comments in production code, I'll add comment on what the curl command request body requirements are. Example "curl -X POST -k -u admin:33Z9j0LZW3a6PoVTaJrncQvV https://127.0.0.1:8089/services/search/distributed/groups/dmc_group_indexer/edit -d member=192.168.31.25:8089 -d member=192.168.37.20:8089 -d member=192.168.47.138:8089 -d default=true"

akondur (approved):

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+}
+
+//UpdateDMCClusteringLabelGroup update respective clustering group
+func (c *SplunkClient) UpdateDMCClusteringLabelGroup(groupName string, groupMembers string) error {
+	endpoint := fmt.Sprintf("%s/services/search/distributed/groups/dmc_indexerclustergroup_%s/edit", c.ManagementURI, groupName)
+	reqBodyClusterGroup := groupMembers + "&default=false"
+	request, err := http.NewRequest("POST", endpoint, strings.NewReader(reqBodyClusterGroup))
+	err = c.Do(request, "200, 201, 409", nil)
+	if err != nil {
+		return err
```
sgontla, line 780: Info request: Now that you are moving the MC logic to a later stage for the associated CR phase to be ready. Having that clarity, do you see any risk for the CR deployment if we keep falling for any of these REST call failures for MC?

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+			Content MCAssetBuildTable `json:"content"`
+		} `json:"entry"`
+	}{}
+	path := "/servicesNS/nobody/splunk_monitoring_console/saved/searches/DMC%20Asset%20-%20Build%20Full"
+	err := c.Get(path, &apiResponseMCAssetTableBuild)
+	if err != nil {
+		return nil, err
+	}
+	if len(apiResponseMCAssetTableBuild.Entry) < 1 {
+		return nil, err
```
sgontla, line 804: err is already nil here, so, do we need to fill any explicit error here?

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+	reqBodyAssetTable := "&trigger_actions=true&dispatch.auto_cancel=" + apiResponseMCAssetTableBuild.DispatchAutoCancel + "&dispatch.buckets=" + strconv.FormatInt(apiResponseMCAssetTableBuild.DispatchBuckets, 10) + "&dispatch.enablePreview=true"
+	endpoint := fmt.Sprintf("%s", c.ManagementURI) + "/servicesNS/nobody/splunk_monitoring_console/saved/searches/DMC%20Asset%20-%20Build%20Full/dispatch"
+	request, err := http.NewRequest("POST", endpoint, strings.NewReader(reqBodyAssetTable))
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	err = c.Do(request, "201", nil)
+	if err != nil {
+		return err
+	}
+	return err
+}
```
sgontla, line 820: Just ` return err`

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+			Content UISettings `json:"content"`
+		} `json:"entry"`
+	}{}
+	path := "/servicesNS/nobody/splunk_monitoring_console/data/ui/nav/default.distributed"
+	err := c.Get(path, &apiResponseUISettings)
+	if err != nil {
+		return nil, err
+	}
+	if len(apiResponseUISettings.Entry) < 1 {
+		return nil, err
```
sgontla, line 844: Same as before..

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+func (c *SplunkClient) UpdateLookupUISettings(configuredPeers string, apiResponseUISettings *UISettings) error {
+	reqBodyMCLookups := "configuredPeers=" + configuredPeers + "&eai:appName=" + apiResponseUISettings.EaiAppName + "&eai:acl=" + apiResponseUISettings.EaiACL + "&eai:userName=" + apiResponseUISettings.EaiUserName + "&disabled=" + strconv.FormatBool(apiResponseUISettings.Disabled)
+	endpoint := fmt.Sprintf("%s/servicesNS/nobody/splunk_monitoring_console/configs/conf-splunk_monitoring_console_assets/settings", c.ManagementURI)
+	request, err := http.NewRequest("POST", endpoint, strings.NewReader(reqBodyMCLookups))
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	err = c.Do(request, "200", nil)
+	if err != nil {
+		return err
+	}
+	return err
```
sgontla, line 859: `return err`

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+
+//UpdateMonitoringConsoleApp updates the monitoring console app
+func (c *SplunkClient) UpdateMonitoringConsoleApp() error {
+	endpoint := fmt.Sprintf("%s/servicesNS/nobody/system/apps/local/splunk_monitoring_console", c.ManagementURI)
+	request, err := http.NewRequest("POST", endpoint, nil)
+	if err != nil {
+		return err
+	}
+	err = c.Do(request, "200", nil)
+	if err != nil {
```
sgontla, line 870: same as above

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+		return err
+	}
+	return err
+}
+
+//ClusterInfo is the struct for checking ClusterInfo
+type ClusterInfo struct {
+	MultiSite string `json:"multisite"`
+}
```
sgontla, line 879: Nit: Probably, move all the structure declarations to the beginning of the file?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -76,8 +76,18 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
+	if err != nil {
+		return result, err
+	}
+
 	// check if deletion has been requested
 	if cr.ObjectMeta.DeletionTimestamp != nil {
+		err = ApplyMonitoringConsole(client, cr, cr.Spec.CommonSplunkSpec, getClusterMasterExtraEnv(cr, &cr.Spec.CommonSplunkSpec))
+		if err != nil {
+			return result, err
+		}
```
sgontla, line 90: Sorry, if i missed it during our review. Any reason why we are doing this, when the underlying CR is dying?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -119,6 +123,10 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 
 	// no need to requeue if everything is ready
 	if cr.Status.Phase == splcommon.PhaseReady {
+		err = ApplyMonitoringConsole(client, cr, cr.Spec.CommonSplunkSpec, getClusterMasterExtraEnv(cr, &cr.Spec.CommonSplunkSpec))
+		if err != nil {
+			return result, err
+		}
```
sgontla, line 129: Though it is not harmful, Isn't it expensive to call again and again? Is it a good idea to detect that there is a deployment already in place, and that way bailout here, instead of going deep inside? 

Also, can you please confirm, if all those REST calls are going to be executed by any chance in this path?

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -61,6 +61,11 @@ func TestApplyClusterMaster(t *testing.T) {
 			Name:      "stack1",
 			Namespace: "test",
 		},
+		Spec: enterprisev1.ClusterMasterSpec{
+			CommonSplunkSpec: enterprisev1.CommonSplunkSpec{
+				Mock: true,
+			},
+		},
```
sgontla, line 68: Nit: For now it's OK. But I think a better way to handle this is by indicating a particular flag, other than the Spec, say, for example, a config map entry

pkg/splunk/enterprise/finalizers_test.go
sgontla: Thansk for adding this, and for now it helps us. 
Future action to us:  This DeletionTester requires an overhaul to accommodate multiple features, mainly to accommodate the caller to provide the list of function calls. Otherwise, it is hard to scale any features other than the default flows.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -42,6 +44,17 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
...
 
+	//For IndexerCluster custom resource click "Apply changes" on MC and return
+	if cr.GetObjectKind().GroupVersionKind().Kind == "IndexerCluster" {
+		mgr := monitoringConsolePodManager{cr: &cr, spec: &spec, secrets: secrets, newSplunkClient: splclient.NewSplunkClient}
+		c := mgr.getMonitoringConsoleClient(cr)
+		err := c.AutomateMCApplyChanges(spec.Mock)
+		if err != nil {
+			return err
+		}
+		return err
```
sgontla, line 55: Nit: `return err`

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -66,23 +79,59 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
...
 	}
 
 	mgr := splctrl.DefaultStatefulSetPodManager{}
-	_, err = mgr.Update(client, statefulsetMC, 1)
+	_, err = mgr.Update(client, statefulset, 1)
 	if err != nil {
 		return err
 	}
 
 	return err
```
sgontla, line 107: Nit: `return err`

pkg/splunk/enterprise/util.go
```diff
@@ -69,6 +69,86 @@ func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar
...
+	if spec.ClusterMasterRef.Name != "" {
+		clusterMasterURL := GetSplunkServiceName(SplunkClusterMaster, spec.ClusterMasterRef.Name, false)
+		if spec.ClusterMasterRef.Namespace != "" {
+			clusterMasterURL = splcommon.GetServiceFQDN(spec.ClusterMasterRef.Namespace, clusterMasterURL)
+		}
```
sgontla, line 78: I think we need an additional conditional check here, just to make sure that the current CRs namespace same as the ClusterMasterRef. If there is a mismatch, we should not proceed here... Thoughts?

Also, GetServiceFQDN is not an option here, and it should be the Master URL always?

pkg/splunk/enterprise/util.go
```diff
@@ -69,6 +69,86 @@ func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar
...
+	}
+}
+
+// getLicenseMasterURL returns URL of license master
+func getLicenseMasterURL(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec) []corev1.EnvVar {
+	if spec.LicenseMasterRef.Name != "" {
+		licenseMasterURL := GetSplunkServiceName(SplunkLicenseMaster, spec.LicenseMasterRef.Name, false)
+		if spec.LicenseMasterRef.Namespace != "" {
+			licenseMasterURL = splcommon.GetServiceFQDN(spec.LicenseMasterRef.Namespace, licenseMasterURL)
+		}
```
sgontla, line 110: Same as the case of Licence Master. I may be missing something here, but please let me know if we don't need FQDN in a certain cases?
Wondering how this is different from a case when we are dealing with the REST calls..

pkg/splunk/enterprise/util.go
```diff
@@ -69,6 +69,86 @@ func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar
...
+		{
+			Name:  "SPLUNK_SEARCH_HEAD_URL",
+			Value: GetSplunkStatefulsetUrls(cr.GetNamespace(), SplunkSearchHead, cr.GetName(), replicas, false),
+		}, {
+			Name:  "SPLUNK_SEARCH_HEAD_CAPTAIN_URL",
+			Value: GetSplunkStatefulsetURL(cr.GetNamespace(), SplunkSearchHead, cr.GetName(), 0, false),
+		},
+	}
+}
+
```
sgontla, line 151: Can these 2 become a single function?

sgontla (changes requested):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -1,13 +1,8 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: clustermasters.enterprise.splunk.com
 spec:
-  additionalPrinterColumns:
```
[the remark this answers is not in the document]
kashok-splunk: I did check yq version, I just have only one version 3.2.1 installed, maybe one of you can help me with this?

pkg/splunk/enterprise/util.go
kashok-splunk: For rest calls we are using fqdn service, For license master, we are not using it anywhere. This is just for MC to add LM as its peer

pkg/splunk/enterprise/monitoringconsole.go
sgontla: line 52 to 55 can be merged to  one line statment, i.e "return err"

pkg/splunk/client/enterprise.go
sgontla: In either case, we are returning error, so, like 833 to 836 can be replaced with "return err"

pkg/splunk/client/enterprise.go
sgontla: I think most of our existing code is following to have the declarations at the beginning. And we can follow the same.

pkg/splunk/enterprise/clustermaster.go
sgontla: My questions is a bit different, let us discuss

pkg/splunk/enterprise/util.go
sgontla: Are we supporting the case of indexers in one namespace and CM in another namespace? If so, how are we dealing with the two MCs in different namespaces? Are we officially supporting the deployments across the namespaces?

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+	apiResponseMCDistributedPeers := struct {
+		Entry []struct {
+			Name    string             `json:"name"`
+			Content MCDistributedPeers `json:"content"`
+		} `json:"entry"`
+	}{}
+	path := "/services/search/distributed/peers"
+	err = c.Get(path, &apiResponseMCDistributedPeers)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
sgontla: My ask was different to have a meaningful error based on the situation than a generic error. Anyhow, after going through a couple of error messages, it seems that SDK is also printing the line numbers as part of the stack trace, which has the line numbers, so we are good here.

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +613,291 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+	}
+
+	for _, e := range apiResponseMCDistributedPeers.Entry {
+		if configuredPeers == "" {
+			configuredPeers = e.Name
+		} else {
+			str := []string{configuredPeers, e.Name}
+			configuredPeers = strings.Join(str, ",")
+		}
```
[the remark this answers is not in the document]
sgontla: I think what you have is fine if it makes more sense for readability :-)

pkg/splunk/client/enterprise.go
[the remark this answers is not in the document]
kashok-splunk: https://github.com/splunk/splunk-operator/pull/135/files/c04d58a94806ee435c5fda45ea95a264aeff76b0#diff-f37c2254cc2ca855cf4bea2777b7e7c0R79 This will give the request URL and response code received, which I think should be sufficient to debug

pkg/splunk/client/enterprise.go
kashok-splunk: Oh yes, makes sense. Fixing this in the latest commit

sgontla (approved):

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -66,23 +76,56 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
...
 }
 
+// getMonitoringConsoleClient for monitoringConsolePodManager returns a SplunkClient for monitoring console
+func (mgr *monitoringConsolePodManager) getMonitoringConsoleClient(cr splcommon.MetaObject) *splclient.SplunkClient {
+	fqdnName := splcommon.GetServiceFQDN(cr.GetNamespace(), GetSplunkServiceName(SplunkMonitoringConsole, cr.GetNamespace(), false))
+	return mgr.newSplunkClient(fmt.Sprintf("https://%s:8089", fqdnName), "admin", string(mgr.secrets.Data["password"]))
+}
+
+// getClusterMasterClient for monitoringConsolePodManager returns a SplunkClient for cluster master
+func (mgr *monitoringConsolePodManager) getClusterMasterClient(cr splcommon.MetaObject) *splclient.SplunkClient {
```
gaurav-splunk, line 111: I think both these APIs, `getClusterMasterClient` and `getMonitoringConsoleClient`, can be a wrapper around a common API( like for e.g.`getSplunkClient`) which takes instanceType as one of the arguments and returns the `splclient.SplunkClient`.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -151,18 +194,23 @@ func getMonitoringConsoleStatefulSet(client splcommon.ControllerClient, cr splco
...
 	//update podTemplate annotation with configMap resource version
-	var monitoringConsoleConfigMap *corev1.ConfigMap
+	//use emptyEnv here to avoid adding site amd multisite_master info in configMap
+	//TODO: Update this configMap retrieval by API from PR #134
 	if cr.GetObjectMeta().GetDeletionTimestamp() != nil {
-		monitoringConsoleConfigMap, _ = ApplyMonitoringConsoleEnvConfigMap(client, cr.GetNamespace(), cr.GetName(), env, false)
+		monitoringConsoleConfigMap, _ = ApplyMonitoringConsoleEnvConfigMap(client, cr.GetNamespace(), cr.GetName(), emptyEnv, false)
 	} else {
-		monitoringConsoleConfigMap, _ = ApplyMonitoringConsoleEnvConfigMap(client, cr.GetNamespace(), cr.GetName(), env, true)
+		monitoringConsoleConfigMap, _ = ApplyMonitoringConsoleEnvConfigMap(client, cr.GetNamespace(), cr.GetName(), emptyEnv, true)
```
gaurav-splunk, line 213: so if we are passing `emptyEnv` in this function, then how will we provide `newURLs` to the configMap?

kashok-splunk:
Integration tests passing locally 
image

smohan-splunk:
> Integration tests passing locally
> image

Manual int-test run due to ongoing circleci limitations

pkg/splunk/client/enterprise.go
```diff
@@ -611,7 +624,282 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
...
+	if mock {
+		return nil
+	}
+	var configuredPeers, indexerMemberList, licenseMasterMemberList string
+	apiResponseServerRoles, err := c.GetMonitoringconsoleServerRoles()
+	if err != nil {
+		return err
+	}
+
+	apiResponseMCDistributedPeers := struct {
```
gaurav-splunk, line 653: I guess these few lines should be a separate function similar to what you did for `GetMonitoringconsoleServerRoles`.

pkg/splunk/enterprise/util.go
gaurav-splunk: No,I dont think so we are supporting indexers in a different namespace than CM. Check `validateIndexerClusterSpec`.
```// validateIndexerClusterSpec checks validity and makes default updates to a IndexerClusterSpec, and returns error if something is wrong.
func validateIndexerClusterSpec(cr *enterprisev1.IndexerCluster) error {
	// We cannot have 0 replicas in IndexerCluster spec, since this refers to number of indexers in an indexer cluster
	if cr.Spec.Replicas == 0 {
		cr.Spec.Replicas = 1
	}

	// Cannot leave clusterMasterRef field empty or else we cannot connect to CM
	if len(cr.Spec.ClusterMasterRef.Name) == 0 {
		return fmt.Errorf("IndexerCluster spec should refer to ClusterMaster via clusterMasterRef")
	}

	// Multisite / multipart clusters: can't reference a cluster master located in another namespace because of Service and Secret limitations
	if len(cr.Spec.ClusterMasterRef.Namespace) > 0 && cr.Spec.ClusterMasterRef.Namespace != cr.GetNamespace() {
		return fmt.Errorf("Multisite cluster does not support cluster master to be located in a different namespace")
	}
	return validateCommonSplunkSpec(&cr.Spec.CommonSplunkSpec)
}

File: pkg/splunk/enterprise/util.go [hunk 30]
```diff
@@ -69,6 +69,86 @@ func getIndexerExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.EnvVar
 	}
 }
 
+// getClusterMasterExtraEnv returns extra environment variables used by indexer clusters
+func getClusterMasterExtraEnv(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec) []corev1.EnvVar {
+	if spec.ClusterMasterRef.Name != "" {
```
gaurav-splunk, line 74: Can there be a case where this is called and `spec.ClusterMasterRef.Name == ""` ?

gaurav-splunk (changes requested):

gaurav-splunk (approved):

kashok-splunk:
Integration tests after the lastest commit
image

PR #137, opened 2020-10-06
Fix CSPL 481

akondur:
On an admin password change on a search head cluster we recycle the search head pods and mount them with the newly created statefulset based secret. The declarative admin password [feature](https://github.com/splunk/splunk-ansible/pull/517/files#) makes sure that the newer admin password is configured on the Splunk instance running inside of the Pod. However there is a destructive sync search head run after the declarative admin password configuration which resync’s the search head’s admin password with that of the captain(which still holds the older admin password pending a recycle from the operator) following which a rest api call fails preventing container start. 

The PR proposes a change for the operator to modify the admin password via a cli command on all members of the search head cluster(SHC) before redeploying them such that a resync during container start will not effect the admin password(as captain already has newer password) and the container start for search heads is successful with the newly configured admin password.


pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -232,6 +240,63 @@ func ApplyShcSecret(mgr *searchHeadClusterPodManager, replicas int32, mock bool)
...
+				return err
+			}
+			scopedLog.Info("Restarted Splunk")
+
+			// Set the adminSecretChanged changed flag to true
+			if i < int32(len(mgr.cr.Status.AdminSecretChanged)) {
+				mgr.cr.Status.AdminSecretChanged[i] = true
+			} else {
+				scopedLog.Info("Appending to AdminSecretChanged")
+				mgr.cr.Status.AdminSecretChanged = append(mgr.cr.Status.AdminSecretChanged, true)
```
gaurav-splunk, line 277: do we need to explicitly call `append` ? I think instead of this if/else you can just simply do `mgr.cr.Status.AdminSecretChanged[i] = true`, isn't it?

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -232,6 +240,63 @@ func ApplyShcSecret(mgr *searchHeadClusterPodManager, replicas int32, mock bool)
...
+		// Change admin password on all secrets mounted on pods to make sure that we want to stay in sync
+		for podSecret := range adminChangedSecrets {
+			podSecret.Data["password"] = []byte(nsAdminSecret)
+			_, err = splctrl.ApplySecret(mgr.c, podSecret)
+			if err != nil {
+				return err
+			}
+			scopedLog.Info("admin password changed on the secret mounted on pod")
+		}
 	}
```
sgontla, line 300: I think the problem with this logic is, in case if there is an error in the above for loop block, there we lose this map for the next iteration, and by that time, that Pod is already marked as Secret changed. 
May be Apply the Pod secret immediately after setting the "mgr.cr.Status.AdminSecretChanged[i] = true". If so, we don't even need the extra map.

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -232,6 +240,63 @@ func ApplyShcSecret(mgr *searchHeadClusterPodManager, replicas int32, mock bool)
...
+
+			// Change admin password on splunk instance of pod
+			command := fmt.Sprintf("/opt/splunk/bin/splunk cmd splunkd rest --noauth POST /services/admin/users/admin 'password=%s'", nsAdminSecret)
+			_, _, err = splutil.PodExecCommand(mgr.c, shPodName, mgr.cr.GetNamespace(), []string{"/bin/sh"}, command, false, false)
+			if err != nil {
+				if !mock {
+					return err
+				}
+			}
+			scopedLog.Info("admin password changed on the splunk instance of pod")
```
sgontla, line 262: Nit: Is it useful to log the pod id/name?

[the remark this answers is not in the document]
akondur: Done

akondur: Done

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -59,6 +59,13 @@ func ApplySearchHeadCluster(client splcommon.ControllerClient, cr *enterprisev1.
...
 	if cr.Status.ShcSecretChanged == nil {
 		cr.Status.ShcSecretChanged = []bool{}
+
+	}
+	if cr.Status.AdminSecretChanged == nil {
+		cr.Status.AdminSecretChanged = []bool{}
+	}
+	if cr.Status.AdminPasswordChangedSecrets == nil {
+		cr.Status.AdminPasswordChangedSecrets = make(map[string]bool)
 	}
```
sgontla, line 69: Nit: Now, we have multiple inits, probably, we need to have a separate init function. But it's not a blocker at this time.

pkg/splunk/enterprise/searchheadcluster.go
sgontla: Nit: In your example, you are using the pointer directly. Actually, go lang append API takes care of 'Nil' case, so that is not an issue.

sgontla (approved):

pkg/splunk/enterprise/searchheadcluster.go
gaurav-splunk: Ok, in that case you can just use `mgr.cr.Status.AdminSecretChanged = append(mgr.cr.Status.AdminSecretChanged, true)` and can get rid of if/else part.

akondur: That will not work either because if we have populated at an index **i** already then we shouldn't append another entry for it. We need the index of the slice to match the pod number.

gaurav-splunk: Oh okay I get it now!! Sorry for the confusion 👍

gaurav-splunk (approved):

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -233,6 +247,67 @@ func ApplyShcSecret(mgr *searchHeadClusterPodManager, replicas int32, mock bool)
...
+			// Adding to map of secrets to be synced
+			podSecret, err := splutil.GetSecretFromPod(mgr.c, shPodName, mgr.cr.GetNamespace())
+			if err != nil {
+				return err
+			}
+			mgr.cr.Status.AdminPasswordChangedSecrets[podSecret.GetName()] = true
+			scopedLog.Info("Secret mounted on pod(to be changed) added to map")
+		}
+	}
+
```
smohan-splunk, line 296: A comment to explain the reason behind using AdminPasswordChangedSecrets would be helpful

smohan-splunk (approved):

pkg/splunk/enterprise/searchheadcluster.go
akondur: Done

PR #139, opened 2020-10-06
Update config.yml

smohan-splunk:
Test Large executor


smohan-splunk:
Closing the dummy PR. Conclusion - circleCI with Large executors doesn't work well and we need to continue working with  XLarge

PR #142, opened 2020-10-07
CSPL-444-Added util to get rf sf status from cm

pdhanoyasplunk:


test/testenv/cmutil.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// ClusterMasterHealthResponse is a representation of the health response by a Splunk cluster-master
+// Endpoint: /services/cluster/master/health
+type ClusterMasterHealthResponse struct {
```
gaurav-splunk, line 11: do we actually need this?I don't see multiple entries in the CM health response. Maybe just rename `ClusterMasterHealthEntry` to `ClusterMasterHealthResponse` and use it directly like `restResponse.Content.SearchFactorMet` ?

test/smoke/smoke_test.go
```diff
@@ -309,6 +323,15 @@ var _ = Describe("Smoke test", func() {
...
 			}, deployment.GetTimeout(), PollInterval).Should(Equal(siteIndexerMap))
 
+			// Verify MC Pod is Ready
+			testenv.MCPodReady(testenvInstance.GetName(), deployment)
+
+			// Verify RF SF is met
+			Eventually(func() bool {
+				rfSfStatus := testenv.CheckRFSF(testenvInstance.GetName(), deployment.GetName())
+				testenvInstance.Log.Info("Verifying RF SF is met", "Status", rfSfStatus)
+				return rfSfStatus
```
sgontla, line 333: Looks like this code is repetition in all the above 3 places. Wondering if this function can be defined explicitly and called from `Eventually()`?

test/testenv/cmutil.go
```diff
@@ -0,0 +1,60 @@
...
+type ClusterMasterHealthContent struct {
+	AllDataIsSearchable      string `json:"all_data_is_searchable"`
+	AllPeersAreUp            string `json:"all_peers_are_up"`
+	Multisite                string `json:"multisite"`
+	NoFixupTasksInProgress   string `json:"no_fixup_tasks_in_progress"`
+	ReplicationFactorMet     string `json:"replication_factor_met"`
+	SearchFactorMet          string `json:"search_factor_met"`
+	SiteReplicationFactorMet string `json:"site_replication_factor_met"`
+	SiteSearchFactorMet      string `json:"site_search_factor_met"`
+}
```
sgontla, line 31: Aren't these be best represented with `bool` than `string`

pdhanoyasplunk: @sgontla the return from Splunk API is string.

test/testenv/cmutil.go
pdhanoyasplunk: @gaurav-splunk I kept the hierarchical order of API return for simplicity. In future I am thinking of adding more verifications in future therefore filtering more fields.

test/testenv/cmutil.go
```diff
@@ -0,0 +1,60 @@
...
+	podName := fmt.Sprintf("splunk-%s-cluster-master-0", deployment.GetName())
+	stdin := "curl -ks -u admin:$(cat /mnt/splunk-secrets/password) https://localhost:8089/services/cluster/master/health?output_mode=json"
+	command := []string{"/bin/sh"}
+	stdout, stderr, err := deployment.PodExecCommand(podName, command, stdin, false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to execute command on pod", "pod", podName, "command", command)
+		return false
+	}
+	logf.Log.Info("Command executed on pod", "pod", podName, "command", command, "stdin", stdin, "stdout", stdout, "stderr", stderr)
+	restResponse := ClusterMasterHealthResponse{}
```
kashok-splunk, line 45: Should we invest time in creating an HTTP client for the test suite also? Till then we can have a common API that can exec on the Pod get the json response, unmarshall, and return an interface with all the desired entries from the json. I see the same being done at multiple places Example: https://github.com/splunk/splunk-operator/blob/develop/test/testenv/search_utils.go#L89 / https://github.com/splunk/splunk-operator/blob/develop/test/ingest_search/ingest_search_test.go#L61

pdhanoyasplunk: @kashok-splunk sure we can take that up as a new task in upcoming sprint.

sgontla (approved):

kashok-splunk (approved):
@pdhanoyasplunk Sure.

smohan-splunk (approved):

PR #152, opened 2020-10-14
CSPL-499: Attaching PVC volumes to splunk monitoring console pod

kashok-splunk:
-Attaching PVC volumes to Splunk monitoring console pod and updating distsearch.conf when any search peer gets deleted
-Updating distsearch.conf on CR deletion
-Adding owner ref to MC configMap

### DIFF
--- a/pkg/splunk/controller/configmap.go
+++ b/pkg/splunk/controller/configmap.go
@@ -75,13 +75,7 @@ func SetConfigMapOwnerRef(client splcommon.ControllerClient, cr splcommon.MetaOb
 	// Owner ref doesn't exist, update statefulset with owner references
 	configMap.SetOwnerReferences(append(configMap.GetOwnerReferences(), splcommon.AsOwner(cr, false)))
 
-	// Update owner reference if needed
-	err = splutil.UpdateResource(client, configMap)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return splutil.UpdateResource(client, configMap)
 }
 
 // GetConfigMap gets the ConfigMap resource in a given namespace
--- a/pkg/splunk/controller/configmap.go
+++ b/pkg/splunk/controller/configmap.go
@@ -57,6 +57,33 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
 	return dataUpdated, err
 }
 
+// SetConfigMapOwnerRef sets owner references for configMap
+func SetConfigMapOwnerRef(client splcommon.ControllerClient, cr splcommon.MetaObject, namespacedName types.NamespacedName) error {
+	configMap, err := GetConfigMap(client, namespacedName)
+	if err != nil {
+		return err
+	}
+
+	currentOwnerRef := configMap.GetOwnerReferences()
+	// Check if owner ref exists
+	for i := 0; i < len(currentOwnerRef); i++ {
+		if reflect.DeepEqual(currentOwnerRef[i], splcommon.AsOwner(cr, false)) {
+			return nil
+		}
+	}
+
+	// Owner ref doesn't exist, update statefulset with owner references
+	configMap.SetOwnerReferences(append(configMap.GetOwnerReferences(), splcommon.AsOwner(cr, false)))
+
+	// Update owner reference if needed
+	err = splutil.UpdateResource(client, configMap)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // GetConfigMap gets the ConfigMap resource in a given namespace
 func GetConfigMap(client splcommon.ControllerClient, namespacedName types.NamespacedName) (*corev1.ConfigMap, error) {
 	var configMap corev1.ConfigMap
--- a/pkg/splunk/controller/configmap_test.go
+++ b/pkg/splunk/controller/configmap_test.go
@@ -21,9 +21,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
-	enterprisev1 "github.com/splunk/splunk-operator/pkg/apis/enterprise/v1beta1"
 	spltest "github.com/splunk/splunk-operator/pkg/splunk/test"
-	splutil "github.com/splunk/splunk-operator/pkg/splunk/util"
 )
 
 func TestApplyConfigMap(t *testing.T) {
--- a/pkg/splunk/controller/configmap_test.go
+++ b/pkg/splunk/controller/configmap_test.go
@@ -19,8 +19,11 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
+	enterprisev1 "github.com/splunk/splunk-operator/pkg/apis/enterprise/v1beta1"
 	spltest "github.com/splunk/splunk-operator/pkg/splunk/test"
+	splutil "github.com/splunk/splunk-operator/pkg/splunk/util"
 )
 
 func TestApplyConfigMap(t *testing.T) {
@@ -40,4 +43,70 @@ func TestApplyConfigMap(t *testing.T) {
 		return err
 	}
 	spltest.ReconcileTester(t, "TestApplyConfigMap", &current, revised, createCalls, updateCalls, reconcile, false)
+}
+
+func TestSetConfigMapOwnerRef(t *testing.T) {
+	cr := enterprisev1.Standalone{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "stack1",
+			Namespace: "test",
+		},
+	}
+
+	c := spltest.NewMockClient()
+	current := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "splunk-test-monitoring-console",
+			Namespace: "test",
+		},
+		Data: map[string]string{"a": "b"},
+	}
+
+	namespacedName := types.NamespacedName{Namespace: "test", Name: "splunk-test-monitoring-console"}
+
+	err := SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err.Error() != "NotFound" {
+		t.Errorf("Couldn't detect resource %s", current.GetName())
+	}
+
+	// Create configMap
+	err = splutil.CreateResource(c, &current)
+	if err != nil {
+		t.Errorf("Failed to create owner reference  %s", current.GetName())
+	}
+
+	// Test existing owner reference
+	err = SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err != nil {
+		t.Errorf("Couldn't set owner ref for resource %s", current.GetName())
+	}
+
+	// Try adding same owner again
+	err = SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err != nil {
+		t.Errorf("Couldn't set owner ref for statefulset %s", current.GetName())
+	}
+}
+
+func TestGetConfigMap(t *testing.T) {
+	c := spltest.NewMockClient()
+
+	current := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "splunk-test-monitoring-console",
+			Namespace: "test",
+		},
+		Data: map[string]string{"a": "b"},
+	}
+
+	_, err := ApplyConfigMap(c, &current)
+	if err != nil {
+		return
+	}
+
+	namespacedName := types.NamespacedName{Namespace: "test", Name: "splunk-test-monitoring-console"}
+	_, err = GetConfigMap(c, namespacedName)
+	if err != nil {
+		t.Errorf(err.Error())
+	}
 }
--- a/pkg/splunk/controller/configmap_test.go
+++ b/pkg/splunk/controller/configmap_test.go
@@ -66,19 +66,19 @@ func TestSetConfigMapOwnerRef(t *testing.T) {
 
 	err := SetConfigMapOwnerRef(c, &cr, namespacedName)
 	if err.Error() != "NotFound" {
-		t.Errorf("Couldn't detect resource %s", current.GetName())
+		t.Errorf("Found resource before creating it %s", current.GetName())
 	}
 
 	// Create configMap
 	err = splutil.CreateResource(c, &current)
 	if err != nil {
-		t.Errorf("Failed to create owner reference  %s", current.GetName())
+		t.Errorf("Failed to create resource %s", current.GetName())
 	}
 
 	// Test existing owner reference
 	err = SetConfigMapOwnerRef(c, &cr, namespacedName)
 	if err != nil {
-		t.Errorf("Couldn't set owner ref for resource %s", current.GetName())
+		t.Errorf("Couldn't set owner ref for configMap %s", current.GetName())
 	}
 
 	// Try adding same owner again
@@ -101,7 +101,7 @@ func TestGetConfigMap(t *testing.T) {
 
 	_, err := ApplyConfigMap(c, &current)
 	if err != nil {
-		return
+		t.Errorf(err.Error())
 	}
 
 	namespacedName := types.NamespacedName{Namespace: "test", Name: "splunk-test-monitoring-console"}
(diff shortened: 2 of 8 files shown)


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -345,4 +404,26 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+	}
+
+	for i := 0; i < len(deletedPeers); i++ {
+		command := fmt.Sprintf("/opt/splunk/bin/splunk remove search-server %s:8089 -auth admin:%s", deletedPeers[i], adminPwd)
+		_, _, err := splutil.PodExecCommand(c, cmPodName, cr.GetNamespace(), []string{"/bin/sh"}, command, false, false)
+		if err != nil {
+			if !spec.Mock {
+				return err
+			}
+		}
```
sgontla, line 426: When we are in this loop, and I think failing for any of the peers in the list will make us to restart doing the podExec commands again from the beginning. So, I think it will be great if we can validate and make sure that, failing this command for a non-existing host should not put an error condition, pushing us to an infinite loop.

pkg/splunk/enterprise/monitoringconsole_test.go
```diff
@@ -205,3 +205,39 @@ func TestApplyMonitoringConsoleEnvConfigMap(t *testing.T) {
...
+				{Ready: true},
+			},
+		},
+	}
+
+	c := spltest.NewMockClient()
+	c.AddObject(pod)
+	err := RemoveSearchPeers(c, &cr, cr.Spec.CommonSplunkSpec, deletedPeers)
+	if err != nil {
+		return
```
sgontla, line 241: t.Errorf() in case of failure?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -198,6 +207,50 @@ func getMonitoringConsoleStatefulSet(client splcommon.ControllerClient, cr splco
 		},
 	}
 
+	// update template to include storage for etc and var volumes
+	if spec.EphemeralStorage {
```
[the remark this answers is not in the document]
gaurav-splunk: You mean separate CR for MC?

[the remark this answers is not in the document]
sgontla: My bad, yes, MC.

pkg/splunk/enterprise/monitoringconsole_test.go
akondur: +1

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -345,4 +404,26 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
 			}
 		}
 	}
+
+	return deletedPeers
+}
+
+//RemoveSearchPeers removes deleted search peers and updates etc/system/local/distsearch.conf
+func RemoveSearchPeers(c splcommon.ControllerClient, cr splcommon.MetaObject, spec enterprisev1.CommonSplunkSpec, deletedPeers []string) error {
+	cmPodName := fmt.Sprintf("splunk-%s-monitoring-console-0", cr.GetNamespace())
```
akondur, line 413: nit: mcPodName vs cmPodName?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -313,16 +367,21 @@ func AddURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1
...
 //DeleteURLsConfigMap for deleting server peers to the monitoring console or scaling down
-func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1.EnvVar, deleteCR bool) {
+func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1.EnvVar, deleteCR bool) []string {
+	var deletedPeers []string
 	for _, url := range newURLs {
 		currentURLs := strings.Split(revised.Data[url.Name], ",")
 		sort.Strings(currentURLs)
 		for _, curr := range currentURLs {
 			//scale DOWN
 			if strings.Contains(curr, crName) && !strings.Contains(url.Value, curr) && !deleteCR {
```
gaurav-splunk, line 377: Maybe a better way to re-arrange this is - 
```
if strings.Contains(curr, crName) {
    if deleteCR {
        revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], url.Value, "")
        if url.Name != "SPLUNK_MULTISITE_MASTER" {
	    deletedPeers = strings.Split(url.Value, ",")
        }
    } else if !strings.Contains(url.Value, curr) {
             revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], curr, "")
	     deletedPeers = append(deletedPeers, curr)
    }
}

File: pkg/splunk/enterprise/monitoringconsole.go [hunk 6]
```diff
@@ -313,16 +367,21 @@ func AddURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1
...
 				revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], curr, "")
+				deletedPeers = append(deletedPeers, curr)
 			} else if strings.Contains(curr, crName) && deleteCR {
 				revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], url.Value, "")
+				if url.Name != "SPLUNK_MULTISITE_MASTER" {
+					deletedPeers = strings.Split(url.Value, ",")
```
akondur, line 383: Should we append here? Or directly assign?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -313,16 +367,21 @@ func AddURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1
 }
 
 //DeleteURLsConfigMap for deleting server peers to the monitoring console or scaling down
-func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1.EnvVar, deleteCR bool) {
+func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1.EnvVar, deleteCR bool) []string {
+	var deletedPeers []string
 	for _, url := range newURLs {
 		currentURLs := strings.Split(revised.Data[url.Name], ",")
```
akondur, line 373: Nit: Just a thought, are we making sure the **configMap revised's data section** has an entry for **key url.Name**? If yes, please ignore.

akondur (approved):

pkg/splunk/enterprise/monitoringconsole.go
gaurav-splunk: Wouldn't the `deletedPeers` be updated in every reconcile loop? If yes, then we are fine but if not, then we definitely need to keep a track of already removed peers.

gaurav-splunk (changes requested):

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Yes, deleted peers get updated in every reconcile loop.

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: url.Value is the entire string here with multiple URLs example "a,b,c,c" and we need []string = {"a", "b", "c"}

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: yes, revised configMap is always updated

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Directly assigning since either we will scale down or delete cr. So in case of scaling down, we are appending one by one and when deleting entire cr we add all the instances names that need to be deleted

pkg/splunk/enterprise/monitoringconsole_test.go
kashok-splunk: Sure, will be more meaningful. For consistency, I did something similar to https://github.com/splunk/splunk-operator/blob/develop/pkg/splunk/enterprise/searchheadcluster.go#L229

kashok-splunk: My bad. Taking care of this in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: I think it's a good idea to keep a track because in case we try to delete a peer that is deleted we may stuck in error loop forever. In the case of deleting multiple search servers together, we might face this. I should take care of this

sgontla: I guess, we can do a quick test to see the behavior, and if required, then only we can take care.

pkg/splunk/enterprise/monitoringconsole.go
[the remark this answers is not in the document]
sgontla: Since we have a separate STS for MC, I think we can just go with PVCs, so that, we don't need to worry about the combinations of storages across the CRs here.

sgontla (changes requested):
In the case of MC, I think we just need to go with the PVCs instead of making storage decision based on CRS

gaurav-splunk (approved):

kashok-splunk:
Int-tests are failing because they need ansible changes. Passing locally with new ansible changes  
image

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -198,6 +218,25 @@ func getMonitoringConsoleStatefulSet(client splcommon.ControllerClient, cr splco
 		},
 	}
 
+	// prepare and append persistent volume claims if storage is not ephemeral
+	var err error
+	statefulSet.Spec.VolumeClaimTemplates, err = getSplunkVolumeClaims(cr, spec, labels)
```
sgontla, line 223: I think we better avoid any spec-specific dependencies completely. Idea is, to generate the MC STS, and Pod templates that should be unique irrespective of the CR type, and their configurations.
For the Image parameter, I put a story CSPL-555. But, probably, we can use that to remove the dependency overall(?).

kashok-splunk: Sure, We can make CSPL-555 a generic story to remove any spec dependencies

kashok-splunk: Anything we need from this PR? @sgontla

akondur (approved):

pkg/splunk/controller/configmap.go
```diff
@@ -57,6 +57,33 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
...
+	// Owner ref doesn't exist, update statefulset with owner references
+	configMap.SetOwnerReferences(append(configMap.GetOwnerReferences(), splcommon.AsOwner(cr, false)))
+
+	// Update owner reference if needed
+	err = splutil.UpdateResource(client, configMap)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
sgontla, line 84: Nit: can be replaced with `return splutil.UpdateResource(client, configMap)`

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +44,69 @@ func TestApplyConfigMap(t *testing.T) {
...
+			Namespace: "test",
+		},
+		Data: map[string]string{"a": "b"},
+	}
+
+	namespacedName := types.NamespacedName{Namespace: "test", Name: "splunk-test-monitoring-console"}
+
+	err := SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err.Error() != "NotFound" {
+		t.Errorf("Couldn't detect resource %s", current.GetName())
```
sgontla, line 69: Nit: "Could detect resource, when it is not supposed to", or something like that..

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +44,69 @@ func TestApplyConfigMap(t *testing.T) {
...
+	}
+
+	// Create configMap
+	err = splutil.CreateResource(c, &current)
+	if err != nil {
+		t.Errorf("Failed to create owner reference  %s", current.GetName())
```
sgontla, line 75: Nit: "Failed to create resource"

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +44,69 @@ func TestApplyConfigMap(t *testing.T) {
...
+	}
+
+	// Test existing owner reference
+	err = SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err != nil {
+		t.Errorf("Couldn't set owner ref for resource %s", current.GetName())
```
sgontla, line 81: "Couldn't set owner ref for resource " ==> "Couldn't set owner ref for configMap "?

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +44,69 @@ func TestApplyConfigMap(t *testing.T) {
...
+	}
+
+	// Try adding same owner again
+	err = SetConfigMapOwnerRef(c, &cr, namespacedName)
+	if err != nil {
+		t.Errorf("Couldn't set owner ref for statefulset %s", current.GetName())
```
sgontla, line 87: Same as above

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +44,69 @@ func TestApplyConfigMap(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "splunk-test-monitoring-console",
+			Namespace: "test",
+		},
+		Data: map[string]string{"a": "b"},
+	}
+
+	_, err := ApplyConfigMap(c, &current)
+	if err != nil {
+		return
```
sgontla, line 104: t.Errorf()?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -313,16 +353,25 @@ func AddURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []corev1
...
 		for _, curr := range currentURLs {
-			//scale DOWN
-			if strings.Contains(curr, crName) && !strings.Contains(url.Value, curr) && !deleteCR {
-				revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], curr, "")
-			} else if strings.Contains(curr, crName) && deleteCR {
-				revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], url.Value, "")
+			if strings.Contains(curr, crName) {
+				if deleteCR {
+					revised.Data[url.Name] = strings.ReplaceAll(revised.Data[url.Name], url.Value, "")
+					if url.Name != "SPLUNK_MULTISITE_MASTER" && url.Name != "SPLUNK_DEPLOYER_URL" && url.Name != "SPLUNK_SEARCH_HEAD_CAPTAIN_URL" && url.Name != "SPLUNK_SITE" {
```
sgontla, line 365: I think the inclusion list check may give better clarity here, OR else, can we have a comment here?

pkg/splunk/controller/configmap.go
kashok-splunk: Fixing in the latest commit

pkg/splunk/controller/configmap_test.go
kashok-splunk: Fixing in the latest commit

pkg/splunk/controller/configmap_test.go
kashok-splunk: Fixing in the latest commit

pkg/splunk/controller/configmap_test.go
kashok-splunk: Fixing in the latest commit

pkg/splunk/controller/configmap_test.go
kashok-splunk: Fixing in the latest commit

pkg/splunk/enterprise/monitoringconsole.go
kashok-splunk: Done

pkg/splunk/controller/configmap_test.go
kashok-splunk: Fixing in the latest commit

sgontla (approved):

smohan-splunk (approved):

PR #153, opened 2020-10-14
CSPL-510: Updating Monitoring console on Indexer deletion

kashok-splunk:
Updating Monitoring console on Indexer deletion


sgontla (approved):

akondur (approved):
Looks good! :)

smohan-splunk:
Quick question: wouldn't Standalone SHs or SHC members connected to Indexer cluster not have the same problem (of unable to delete the respective CRs) if CM was to be deleted first?

kashok-splunk:
This issue is only when deleting indexer cr, as when deleting indexer cr we were doing some rest calls on mc, so if mc is not present those rest calls will fail and hence this issue. For another other cr deletion, we just update mc configMap, no rest calls involved

smohan-splunk (approved):

gaurav-splunk (approved):

PR #155, opened 2020-10-16
CSPL-511 - Avoid adding volume mounts for MC

kashok-splunk:
- when creating MC with license master it was adding license volume mount on MC

### DIFF
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -473,8 +473,8 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		}
 	}
 
-	// Add custom volumes to splunk containers
-	if spec.Volumes != nil {
+	// Add custom volumes to splunk containers other than MC(where CR spec volumes are not needed)
+	if spec.Volumes != nil && instanceType != SplunkMonitoringConsole {
 		podTemplateSpec.Spec.Volumes = append(podTemplateSpec.Spec.Volumes, spec.Volumes...)
 		for idx := range podTemplateSpec.Spec.Containers {
 			for v := range spec.Volumes {


pkg/splunk/enterprise/configuration.go
```diff
@@ -473,8 +473,8 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		}
 	}
 
-	// Add custom volumes to splunk containers
-	if spec.Volumes != nil {
+	// Add custom volumes to splunk containers other than MC(where CR spec volumes are not needed)
+	if spec.Volumes != nil && instanceType != SplunkMonitoringConsole {
 		podTemplateSpec.Spec.Volumes = append(podTemplateSpec.Spec.Volumes, spec.Volumes...)
```
sgontla, line 478: If not now, later, can we have a test case for the MC, that validates the  MC Pod spec, to cover any accidental additions going forward? such that, we can take care of MC Pod for its uniqueness, as we keep adding more and more resources to the Pod spec.

sgontla (approved):

pkg/splunk/enterprise/configuration.go
kashok-splunk: Yes, I agree there should be a test to verify the uniqueness of the MC pod. Will add that in going in future, need to verify all components we can add in the pod spec that can create an issue here

smohan-splunk (approved):

pkg/splunk/enterprise/configuration.go
akondur: As discussed, as a backlog item we can rephrase this API to **not** depend on **cr.spec** to make decisions on **PodTemplateSpec for MC**.

akondur (approved):

gaurav-splunk (approved):

PR #159, opened 2020-10-19
CSPL-521: Added more delay and checks for MC pod to be ready

pdhanoyasplunk:
Made following changes to address flakiness related to MC test. 

* Changed wait time to 1500 seconds. 
* Added check to verify MC pod stays in ready state.

### DIFF
--- a/test/testenv/testenv.go
+++ b/test/testenv/testenv.go
@@ -31,7 +31,7 @@ const (
 	defaultSparkImage    = "splunk/spark"
 
 	// defaultTestTimeout is the max timeout in seconds before async test failed.
-	defaultTestTimeout = 900
+	defaultTestTimeout = 1500
 
 	// PollInterval specifies the polling interval
 	PollInterval = 5 * time.Second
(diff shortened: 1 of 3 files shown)


test/testenv/testenv.go
```diff
@@ -31,7 +31,7 @@ const (
 	defaultSparkImage    = "splunk/spark"
 
 	// defaultTestTimeout is the max timeout in seconds before async test failed.
-	defaultTestTimeout = 900
+	defaultTestTimeout = 1500
```
gaurav-splunk, line 34: What is the reason for increasing this timeout?

pdhanoyasplunk: Because of MC deployment is taking longer hence 1500 seconds to make tests more stable. Also this worst case scenario ideally we should be waiting much less time.

akondur (approved):

kashok-splunk (approved):

gaurav-splunk (approved):

PR #164, opened 2020-10-22
CSPL-529: Fixing a bug where if we delete CM, it also deletes PVCs for IndexerCluster too

gaurav-splunk:
**Problem**
When we were trying to delete ClusterMaster first, then it was also deleting PVCs for IndexerCluster too( in addition to its own PVC).
This was happening because we were looking for PVCs which match the label **app.kubernetes.io/part-of** and all the PVCs of IndexerCluster and PVC of ClusterMaster had this same label. Hence it was deleting all the PVCs.

**Solution**
Now, instead of looking for PVCs with label **app.kubernetes.io/part-of**, we are looking for PVCs with label **app.kubernetes.io/instance**, which is unique per CR and hence only those PVCs which are associated with that PR will be deleted.

### DIFF
--- a/pkg/splunk/enterprise/finalizers.go
+++ b/pkg/splunk/enterprise/finalizers.go
@@ -54,7 +54,7 @@ func DeleteSplunkPvc(cr splcommon.MetaObject, c splcommon.ControllerClient) erro
 		return nil
 	}
 
-	// get list of PVCs for this cluster
+	// get list of PVCs associated with this CR
 	labels := map[string]string{
 		"app.kubernetes.io/instance": fmt.Sprintf("splunk-%s-%s", cr.GetName(), component),
 	}
--- a/pkg/splunk/enterprise/finalizers.go
+++ b/pkg/splunk/enterprise/finalizers.go
@@ -31,19 +31,24 @@ func init() {
 
 // DeleteSplunkPvc removes all corresponding PersistentVolumeClaims that are associated with a custom resource.
 func DeleteSplunkPvc(cr splcommon.MetaObject, c splcommon.ControllerClient) error {
-	scopedLog := log.WithName("DeleteSplunkPvc").WithValues("kind", cr.GetObjectKind().GroupVersionKind().Kind,
+	var objectKind string
+	objectKind = cr.GetObjectKind().GroupVersionKind().Kind
+
+	scopedLog := log.WithName("DeleteSplunkPvc").WithValues("kind", objectKind,
 		"name", cr.GetName(), "namespace", cr.GetNamespace())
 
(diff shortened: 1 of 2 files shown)


smohan-splunk (approved):
We should also test this extended scenario:
 - Deploy CM & Indexer cluster
 - Delete CM, ensure PVCs of indexers remain intact
 - Re-deploy CM, ensure Indexers continue being "part-of" the revived CM & healthy

gaurav-splunk:
> We should also test this extended scenario:
> 
> * Deploy CM & Indexer cluster
> * Delete CM, ensure PVCs of indexers remain intact
> * Re-deploy CM, ensure Indexers continue being "part-of" the revived CM & healthy

Yes, I have tried this and this works now :)

pkg/splunk/enterprise/finalizers.go
```diff
@@ -31,27 +31,32 @@ func init() {
...
+		component = "cluster-master"
 	default:
 		scopedLog.Info("Skipping PVC removal")
 		return nil
 	}
 
 	// get list of PVCs for this cluster
 	labels := map[string]string{
-		"app.kubernetes.io/part-of": fmt.Sprintf("splunk-%s-%s", cr.GetName(), component),
+		"app.kubernetes.io/instance": fmt.Sprintf("splunk-%s-%s", cr.GetName(), component),
```
akondur, line 59: Just a suggestion for now or in the future: 

To maintain consistency, it might be good to use the same API to set/retrieve labels(GetSplunkLabels). We could modify GetSplunkLabels's prototype to **getSplunkLabels(instanceIdentifier string, instanceType InstanceType, partOfIdentifier string, selectFew []string)** and call it using **var sfew []string = "instance"** as the selectFew parameter.

akondur (approved):

sgontla (approved):

kashok-splunk (approved):

pkg/splunk/enterprise/finalizers.go
gaurav-splunk: Probably will take it for future improvement.

pkg/splunk/enterprise/finalizers.go
```diff
@@ -31,27 +31,32 @@ func init() {
...
+	case "IndexerCluster":
 		component = "indexer"
+	case "ClusterMaster":
+		component = "cluster-master"
 	default:
 		scopedLog.Info("Skipping PVC removal")
 		return nil
 	}
 
 	// get list of PVCs for this cluster
```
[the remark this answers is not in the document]
gaurav-splunk: Done

PR #166, opened 2020-10-26
bugfix/CSPL-466: If in IndexerCluster spec, replicas < RF(which is configured on ClusterMaster), then set replicas=RF.

gaurav-splunk:
**Problem**
In case if IndexerCluster spec has replicas < replication_factor or site_replication_factor as set on ClusterMaster, then IndexerCluster goes in the reconcilation loop since CM is never in Ready state. And the only way to scale indexer cluster is using kubectl scale statefulset .....

**Solution**
Set replicas = RF if it is less than RF.

### DIFF
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -288,12 +288,20 @@ func ApplyIdxcSecret(mgr *indexerClusterPodManager, replicas int32, mock bool) e
 
 // Update for indexerClusterPodManager handles all updates for a statefulset of indexers
 func (mgr *indexerClusterPodManager) Update(c splcommon.ControllerClient, statefulSet *appsv1.StatefulSet, desiredReplicas int32) (splcommon.Phase, error) {
+
+	var err error
+	//Don't even try to create a statefulset and secret if CM is not ready yet.
+	if mgr.cr.Status.ClusterMasterPhase != splcommon.PhaseReady {
+		mgr.log.Error(err, "Cluster Master is not ready yet")
+		return splcommon.PhaseError, err
+	}
+
 	// Assign client
 	if mgr.c == nil {
 		mgr.c = c
 	}
 	// update statefulset, if necessary
-	_, err := splctrl.ApplyStatefulSet(mgr.c, statefulSet)
+	_, err = splctrl.ApplyStatefulSet(mgr.c, statefulSet)
 	if err != nil {
 		return splcommon.PhaseError, err
 	}
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -433,7 +433,7 @@ func getSiteRepFactorOriginCount(siteRepFactor string) int32 {
 }
 
 // verifyRFPeers verifies the number of peers specified in the replicas section
-// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
+// of IndexerClsuster CR. If it is less than RF, than we set it to RF.
 func (mgr *indexerClusterPodManager) verifyRFPeers(c splcommon.ControllerClient) error {
 	if mgr.c == nil {
 		mgr.c = c
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
+	"strconv"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -73,6 +75,15 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		return result, err
 	}
 
+	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient}
+	// Check if we have configured enough number(<= RF) of replicas
+	if !mgr.cr.Spec.Mock {
+		err = mgr.verifyRFPeers(client)
+		if err != nil {
+			return result, err
+		}
+	}
+
 	// check if deletion has been requested
 	if cr.ObjectMeta.DeletionTimestamp != nil {
 		DeleteOwnerReferencesForResources(client, cr, nil)
@@ -114,7 +125,7 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 	if err != nil {
 		return result, err
 	}
-	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient}
+
 	phase, err := mgr.Update(client, statefulSet, cr.Spec.Replicas)
 	if err != nil {
 		return result, err
@@ -412,6 +423,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
 	}
 
 	return mgr.newSplunkClient(fmt.Sprintf("https://%s:8089", fqdnName), "admin", adminPwd)
+}
+
+func getSiteRepFactor(siteRepFactor string) int32 {
+	//re := regexp.MustCompile(".*total:(?P<rf>.*) }")
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
+	match := re.FindStringSubmatch(siteRepFactor)
+	siteRF, _ := strconv.Atoi(match[1])
+	return int32(siteRF)
+}
+
+// verifyRFPeers verifies the number of peers specified in the replicas section
+// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
+func (mgr *indexerClusterPodManager) verifyRFPeers(c splcommon.ControllerClient) error {
+	if mgr.c == nil {
+		mgr.c = c
+	}
+	cm := mgr.getClusterMasterClient()
+	clusterInfo, err := cm.GetClusterInfo(false)
+	if err != nil {
+		return fmt.Errorf("Could not get cluster info from cluster master")
+	}
+	var replicationFactor int32
+	// if it is a multisite indexer cluster, check site_replication_factor
+	if clusterInfo.MultiSite == "true" {
+		replicationFactor = getSiteRepFactor(clusterInfo.SiteReplicationFactor)
+	} else { // for single site, check replication factor
+		replicationFactor = clusterInfo.ReplicationFactor
+	}
+
+	if mgr.cr.Spec.Replicas < replicationFactor {
+		mgr.log.Info("Changing number of replicas as it is less than RF number of peers", "replicas", mgr.cr.Spec.Replicas)
+		mgr.cr.Spec.Replicas = replicationFactor
+	}
+	return nil
 }
 
 // updateStatus for indexerClusterPodManager uses the REST API to update the status for an IndexerCluster custom resource
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -426,7 +426,6 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
 }
 
 func getSiteRepFactor(siteRepFactor string) int32 {
-	//re := regexp.MustCompile(".*total:(?P<rf>.*) }")
 	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
 	match := re.FindStringSubmatch(siteRepFactor)
 	siteRF, _ := strconv.Atoi(match[1])
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -425,7 +425,8 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
 	return mgr.newSplunkClient(fmt.Sprintf("https://%s:8089", fqdnName), "admin", adminPwd)
 }
 
-func getSiteRepFactor(siteRepFactor string) int32 {
+// getSiteRepFactorOriginCount gets the origin count of the site_replication_factor
+func getSiteRepFactorOriginCount(siteRepFactor string) int32 {
 	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
 	match := re.FindStringSubmatch(siteRepFactor)
 	siteRF, _ := strconv.Atoi(match[1])
@@ -446,7 +447,7 @@ func (mgr *indexerClusterPodManager) verifyRFPeers(c splcommon.ControllerClient)
 	var replicationFactor int32
 	// if it is a multisite indexer cluster, check site_replication_factor
 	if clusterInfo.MultiSite == "true" {
-		replicationFactor = getSiteRepFactor(clusterInfo.SiteReplicationFactor)
+		replicationFactor = getSiteRepFactorOriginCount(clusterInfo.SiteReplicationFactor)
 	} else { // for single site, check replication factor
 		replicationFactor = clusterInfo.ReplicationFactor
 	}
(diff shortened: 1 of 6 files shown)


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +425,39 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
 	return mgr.newSplunkClient(fmt.Sprintf("https://%s:8089", fqdnName), "admin", adminPwd)
 }
 
+func getSiteRepFactor(siteRepFactor string) int32 {
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
```
gaurav-splunk, line 429: I am checking for `origin` instead of `total` here because each CR represents number of peers in each site and we should have minimum of `site_replication_factor= origin:` number of peers in each site.

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +425,39 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
 	return mgr.newSplunkClient(fmt.Sprintf("https://%s:8089", fqdnName), "admin", adminPwd)
 }
```
akondur, line 427: Nit: Add a comment here for the function?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +425,39 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
 
+func getSiteRepFactor(siteRepFactor string) int32 {
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
+	match := re.FindStringSubmatch(siteRepFactor)
+	siteRF, _ := strconv.Atoi(match[1])
+	return int32(siteRF)
+}
+
+// verifyRFPeers verifies the number of peers specified in the replicas section
+// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
```
akondur, line 436: Under what scenarios will the idc CR not have .spec.replicas updated? According to [this](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource) shouldn't it be automatically updated?

akondur (changes requested):

smohan-splunk:
Can we update the documentation as part of this PR to indicate the Operator's corrective measures when replicas<RF?

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Not sure if I understood your question, but in our case when we tried to call `kubectl scale idxc --replicas=<new_replica_count>`, it did not update the spec since we were bailing out before we even called `updateStatefulSetPods`.

akondur: The `kubectl scale idxc --replicas=<new_replica_count>` command for idxc CR is supposed to update its `spec.replicas` field for the IDXC CR. `updateStatefulSetPods` is responsible for scaling up/down pods based on this field.

akondur: As per discussion, we are self-correcting in the case of Replicas < SF, RF and we take care of deploying the right number of indexers. If we already met RF/SF, kubectl scale should work.

akondur (approved):

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -73,6 +75,15 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		return result, err
 	}
 
+	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient}
```
smohan-splunk, line 78: Why did we have to add/move the code block to before the `DeletionTimestamp` check?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +425,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+		return fmt.Errorf("Could not get cluster info from cluster master")
+	}
+	var replicationFactor int32
+	// if it is a multisite indexer cluster, check site_replication_factor
+	if clusterInfo.MultiSite == "true" {
+		replicationFactor = getSiteRepFactorOriginCount(clusterInfo.SiteReplicationFactor)
+	} else { // for single site, check replication factor
+		replicationFactor = clusterInfo.ReplicationFactor
+	}
+
```
smohan-splunk, line 454: General question: would be able to recover from a scenario where both the RF & Replicas provided by admin is 0, 1 or a negative number?

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Because we want to check the validity of the spec in the beginning and you need `mgr` for that.

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Yup there is a check in `validateIndexerClusterSpec` for that.

smohan-splunk: I see a check for Replicas being 0 in `validateIndexerClusterSpec`. Would be interesting to test other negative scenarios cc: @pdhanoyasplunk

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +424,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+// getSiteRepFactorOriginCount gets the origin count of the site_replication_factor
+func getSiteRepFactorOriginCount(siteRepFactor string) int32 {
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
+	match := re.FindStringSubmatch(siteRepFactor)
+	siteRF, _ := strconv.Atoi(match[1])
+	return int32(siteRF)
+}
+
+// verifyRFPeers verifies the number of peers specified in the replicas section
+// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
```
smohan-splunk, line 436: Nitpick: typos in the comment

smohan-splunk (approved):

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -73,6 +75,15 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		return result, err
 	}
 
+	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient}
+	// Check if we have configured enough number(<= RF) of replicas
+	if !mgr.cr.Spec.Mock {
```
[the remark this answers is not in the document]
smohan-splunk: Latest commit handles this

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +424,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
+	match := re.FindStringSubmatch(siteRepFactor)
+	siteRF, _ := strconv.Atoi(match[1])
+	return int32(siteRF)
+}
+
+// verifyRFPeers verifies the number of peers specified in the replicas section
+// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
+func (mgr *indexerClusterPodManager) verifyRFPeers(c splcommon.ControllerClient) error {
+	if mgr.c == nil {
```
akondur, line 438: The assignment of client is done in **IndexerClusterPodManager::Update** as well. Can we move the assignment to this line as: 
`mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient, c:client}`

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +424,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+		mgr.c = c
+	}
+	cm := mgr.getClusterMasterClient()
```
akondur, line 441: In the future it is better to make the **getClusterMasterClient()** and **GetClusterInfo()** APIs independent of the Pod Manager as they are used in MC implementation as well. Mentioned this to @kashok-splunk as a part of her review as well.

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +424,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+func getSiteRepFactorOriginCount(siteRepFactor string) int32 {
+	re := regexp.MustCompile(".*origin:(?P<rf>.*),.*")
+	match := re.FindStringSubmatch(siteRepFactor)
+	siteRF, _ := strconv.Atoi(match[1])
+	return int32(siteRF)
+}
+
+// verifyRFPeers verifies the number of peers specified in the replicas section
+// of IndexerClsuster CR. If it is less than RF, than we se it to RF.
+func (mgr *indexerClusterPodManager) verifyRFPeers(c splcommon.ControllerClient) error {
```
akondur, line 437: Did we make this as a part of the PodManager because it has access to the **getClusterMasterClient()** and **GetClusterInfo()** APIs?

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -145,6 +141,81 @@ func indexerClusterPodManagerTester(t *testing.T, method string, mockHandlers []
...
 	}
+	return mgr, mockSplunkClient
+}
+
+// indexerClusterpodManagerVerifyRFPeersTester is used to verify replicas against RF using a indexerClusterPodManager
+func indexerClusterPodManagerVerifyRFPeersTester(t *testing.T, method string, mgr *indexerClusterPodManager,
+	desiredReplicas int32, wantPhase splcommon.Phase, wantCalls map[string][]spltest.MockFuncCall, wantError error) {
+
+	// initialize client
+	c := spltest.NewMockClient()
```
akondur, line 152: We could assign the mock client in the **getIndexerClusterPodManagerAndClient** API? So that we don't have to assign again here.

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -145,6 +141,81 @@ func indexerClusterPodManagerTester(t *testing.T, method string, mockHandlers []
...
+	}
+
+	if mgr.cr.Spec.Replicas != desiredReplicas {
+		t.Errorf("spec has replicas as %d ; want %d", mgr.cr.Spec.Replicas, desiredReplicas)
+	}
+	// check calls
+	c.CheckCalls(t, method, wantCalls)
+}
+
+func indexerClusterPodManagerReplicasTester(t *testing.T, method string, mockHandlers []spltest.MockHTTPHandler,
```
akondur, line 169: Is this API for multiple types of replica testing in the future? (In this case we are testing only RF peer replicas tester)

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Yes

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Yup, probably we can take it as part of another story for future as part of some code clean-up.

sgontla (approved):

kashok-splunk (approved):

pkg/splunk/enterprise/indexercluster_test.go
gaurav-splunk: I could have done that, but if you assign it in **getIndexerClusterPodManagerAndClient** and then initialize `mgr` like this - 
```// initialize client
	client := spltest.NewMockClient()

	mgr := &indexerClusterPodManager{
		log:     scopedLog,
		cr:      &cr,
		secrets: secrets,
		newSplunkClient: func(managementURI, username, password string) *splclient.SplunkClient {
			c := splclient.NewSplunkClient(managementURI, username, password)
			c.Client = mockSplunkClient
			return c
		},
		c: client,
	}

then, you wont be able to call `c.CheckCalls(t, method, wantCalls)` at line 166, since `CheckCalls` is part of `spltest.MockClient` but `mgr.c` is of type `splcommon.ControllerClient`. Hence I think I am keeping it this way for now.

File: pkg/splunk/enterprise/indexercluster.go (hunk 8 above)
Comment (gaurav-splunk) [reply pkg/splunk/enterprise/indexercluster.go:438]: I could do that but since then the UT will fail because of the reason I gave in the below comment. So leaving it like this for now. But good suggestion though. Might have to think of some other way for this to work with UT as well.

Comment (akondur) [reply pkg/splunk/enterprise/indexercluster.go:438]: Agreed

**Review by akondur — APPROVED** [review]

File: pkg/splunk/util/secrets_test.go [hunk 13]
```diff
@@ -428,7 +428,7 @@ func TestGetVersionedSecretVersion(t *testing.T) {
 
 	// Negative testing with non-integer version
 	for testVersion := 0; testVersion < 10; testVersion++ {
-		testSecretName = splcommon.GetVersionedSecretName(versionedSecretIdentifier, string('A'-1+testVersion))
+		testSecretName = splcommon.GetVersionedSecretName(versionedSecretIdentifier, string(rune('A'-1+testVersion)))
```
gaurav-splunk, line 431: Had to use `rune` to fix a compilation error if using go version >=1.15. This was the error when we ran `make test` -
```pkg/splunk/util/secrets_test.go:431:80: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)```

File: pkg/splunk/enterprise/indexercluster.go [hunk 14]
```diff
@@ -97,24 +119,12 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
...
-		cr.Status.ClusterMasterPhase = splcommon.PhaseError
-	}
-
 	// create or update statefulset for the indexers
 	statefulSet, err := getIndexerStatefulSet(client, cr)
 	if err != nil {
 		return result, err
 	}
-	mgr := indexerClusterPodManager{log: scopedLog, cr: cr, secrets: namespaceScopedSecret, newSplunkClient: splclient.NewSplunkClient}
+
```
akondur, line 127: Nit: Does it make sense to move below piece of code to here so we avoid a call to **mgr.Update**?

```
	if mgr.cr.Status.ClusterMasterPhase != splcommon.PhaseReady {
		mgr.log.Error(err, "Cluster Master is not ready yet")
		return splcommon.PhaseError, err
	}
```

akondur (approved):

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: Can do that but doesn't make much of a difference, anyways you are returning in the beginning.

smohan-splunk (approved):

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -414,6 +432,40 @@ func (mgr *indexerClusterPodManager) getClusterMasterClient() *splclient.SplunkC
...
+	var replicationFactor int32
+	// if it is a multisite indexer cluster, check site_replication_factor
+	if clusterInfo.MultiSite == "true" {
+		replicationFactor = getSiteRepFactorOriginCount(clusterInfo.SiteReplicationFactor)
+	} else { // for single site, check replication factor
+		replicationFactor = clusterInfo.ReplicationFactor
+	}
+
+	if mgr.cr.Spec.Replicas < replicationFactor {
+		mgr.log.Info("Changing number of replicas as it is less than RF number of peers", "replicas", mgr.cr.Spec.Replicas)
```
kashok-splunk, line 463: super nit: Maybe can change the logging to something like "Increasing number of indexer replicas to match RF number of peers" or something similar

sgontla (approved):

kashok-splunk (approved):

PR #173, opened 2020-11-06
CSPL-532: Set idxc.secret on the secrets mounted on the indexer pods when idxc.…

akondur:
During the recycle of indexer pods due to an idxc secret change, if there is a container restart(for example if the splunkd process dies) before the operator deletes the pod, the container restart fails due to mismatch of idxc password between Cluster master and that particular indexer.

Changing the idxc passwords on the secrets mounted on the indexer pods to avoid the above.

### DIFF
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -60,6 +60,9 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 	if cr.Status.IndexerSecretChanged == nil {
 		cr.Status.IndexerSecretChanged = []bool{}
 	}
+	if cr.Status.IdxcPasswordChangedSecrets == nil {
+		cr.Status.IdxcPasswordChangedSecrets = make(map[string]bool)
+	}
 	defer func() {
 		err = client.Status().Update(context.TODO(), cr)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -273,6 +288,41 @@ func ApplyIdxcSecret(mgr *indexerClusterPodManager, replicas int32, mock bool) e
...
+			if err != nil {
+				return err
+			}
+			scopedLog.Info("idxc password changed on the secret mounted on pod", "Secret on Pod:", podSecretName)
+
+			// Delete entry in map once value changed
+			delete(mgr.cr.Status.IdxcPasswordChangedSecrets, podSecretName)
+		}
+	}
+
```
sgontla, line 325: I may be wrong in understanding, but wanted to check, If we take this approach, then we are going to overwrite the existing versioned secret than creating a new version, if so, the real purpose of the versioning is compromised?

akondur: The operator will overwrite the existing versioned secret mounted on the Pod(which will stay mounted on the Pod until the operator deletes the pod). Upon deletion, the newer versioned secret will be mounted validating one of the benefits of using the versioning of secrets.

smohan-splunk (approved):

pkg/splunk/enterprise/indexercluster.go
sgontla: As discussed, since this approach is deviating from the versioning logic, I think, we need to update the corresponding .md file for now or somewhere as a Cavea(?),  before we take a final formal approach to address this issue.

kashok-splunk (approved):
Looks good. Just to keep track with this PR CSPL-580 will cover scenarios for any other token change

sgontla (approved):

gaurav-splunk (approved):

PR #174, opened 2020-11-06
CSPL-498 Explicit removal of owner ref from MC statefulset when deleting CR

kashok-splunk:
- Delete CR as the owner of MC statefulset when deleting CR

### DIFF
--- a/pkg/splunk/enterprise/monitoringconsole.go
+++ b/pkg/splunk/enterprise/monitoringconsole.go
@@ -66,9 +66,14 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
 
 	//by default assume we are adding new instances in the monitoring console configMap
 	addNewURLs := true
-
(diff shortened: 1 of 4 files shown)


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -66,9 +66,14 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr splcommon.Meta
...
 	addNewURLs := true
-
+	namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: GetSplunkStatefulsetName(SplunkMonitoringConsole, cr.GetNamespace())}
 	if cr.GetObjectMeta().GetDeletionTimestamp() != nil {
 		addNewURLs = false
+		//explicit removal of owners for MC
+		_, err = splctrl.RemoveStatefulSetOwnerRef(client, cr, namespacedName)
+		if err != nil {
+			return err
+		}
```
sgontla, line 76: Just thinking this, Assuming this change is to have the MC running even if the last CR is deleted(?), then we can avoid setting the owner reference in the first place unless the CRs are really interested in the MC STS state changes(?). If so, I guess, we don't need any of these changes.

PR #175, opened 2020-11-12
cspl-558: Improve code coverage for Index management

sgontla:
UT Code coverage for Smartstore/Indexmanagement area of code

### DIFF
--- a/pkg/splunk/client/enterprise.go
+++ b/pkg/splunk/client/enterprise.go
@@ -629,10 +629,7 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error 
 }
 
 // BundlePush pushes the CM master apps bundle to all the indexer peers
-func (c *SplunkClient) BundlePush(ignoreIdenticalBundle bool, mock bool) error {
-	if mock {
-		return nil
-	}
+func (c *SplunkClient) BundlePush(ignoreIdenticalBundle bool) error {
 	endpoint := fmt.Sprintf("%s/services/cluster/master/control/default/apply", c.ManagementURI)
 	reqBody := fmt.Sprintf("&ignore_identical_bundle=%t", ignoreIdenticalBundle)
 
--- a/pkg/splunk/controller/configmap_test.go
+++ b/pkg/splunk/controller/configmap_test.go
@@ -19,6 +19,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	spltest "github.com/splunk/splunk-operator/pkg/splunk/test"
 )
@@ -40,4 +41,63 @@ func TestApplyConfigMap(t *testing.T) {
 		return err
 	}
 	spltest.ReconcileTester(t, "TestApplyConfigMap", &current, revised, createCalls, updateCalls, reconcile, false)
+}
+
+func TestGetConfigMap(t *testing.T) {
+	current := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "defaults",
+			Namespace: "test",
+		},
+	}
+
+	client := spltest.NewMockClient()
+	namespacedName := types.NamespacedName{Namespace: current.GetNamespace(), Name: current.GetName()}
+
+	_, err := GetConfigMap(client, namespacedName)
+	if err == nil {
+		t.Errorf("Should return an error, when the configMap doesn't exist")
+	}
+
+	_, err = ApplyConfigMap(client, &current)
+	if err != nil {
+		t.Errorf("Failed to create the configMap. Error: %s", err.Error())
+	}
+
+	_, err = GetConfigMap(client, namespacedName)
+	if err != nil {
+		t.Errorf("Should not return an error, when the configMap exists")
+	}
+}
+
+func TestGetConfigMapResourceVersion(t *testing.T) {
+	current := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "defaults",
+			Namespace: "test",
+		},
+	}
+
+	client := spltest.NewMockClient()
+	namespacedName := types.NamespacedName{Namespace: current.GetNamespace(), Name: current.GetName()}
+
+	_, err := GetConfigMap(client, namespacedName)
+	if err == nil {
+		t.Errorf("Should return an error, when the configMap doesn't exist")
+	}
+
+	_, err = GetConfigMapResourceVersion(client, namespacedName)
+	if err == nil {
+		t.Errorf("Should return an error, when the configMap doesn't exist")
+	}
+
+	_, err = ApplyConfigMap(client, &current)
+	if err != nil {
+		t.Errorf("Failed to create the configMap. Error: %s", err.Error())
+	}
+
+	_, err = GetConfigMapResourceVersion(client, namespacedName)
+	if err != nil {
+		t.Errorf("Should not return an error, when the configMap exists")
+	}
 }
--- a/pkg/splunk/enterprise/clustermaster_test.go
+++ b/pkg/splunk/enterprise/clustermaster_test.go
@@ -16,6 +16,7 @@ package enterprise
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -243,6 +244,13 @@ func TestPerformCmBundlePush(t *testing.T) {
 
 	client := spltest.NewMockClient()
 
+	// When the secret object is not present, should return an error
+	current.Status.BundlePushTracker.NeedToPushMasterApps = true
+	err := PerformCmBundlePush(client, &current)
+	if err == nil {
+		t.Errorf("Should return error, when the secret object is not present")
+	}
+
 	secret, err := splutil.ApplyNamespaceScopedSecretObject(client, "test")
 	if err != nil {
 		t.Errorf(err.Error())
@@ -261,23 +269,36 @@ func TestPerformCmBundlePush(t *testing.T) {
 		Data: map[string]string{configToken: ""},
 	}
 
-	current.Status.BundlePushTracker.NeedToPushMasterApps = true
-	current.Status.BundlePushTracker.LastCheckInterval = time.Now().Unix() - 100
-
 	_, err = splctrl.ApplyConfigMap(client, &smartstoreConfigMap)
-
-	err = PerformCmBundlePush(client, &current)
 	if err != nil {
-		t.Errorf("Bundle Push failed")
+		t.Errorf(err.Error())
 	}
 
+	current.Status.BundlePushTracker.NeedToPushMasterApps = true
+
+	//Re-attempting to push the CM bundle in less than 5 seconds should return an error
+	current.Status.BundlePushTracker.LastCheckInterval = time.Now().Unix() - 1
+	err = PerformCmBundlePush(client, &current)
+	if err == nil {
+		t.Errorf("Bundle Push Should fail, if attempted to push within 5 seconds interval")
+	}
+
+	//Re-attempting to push the CM bundle after 5 seconds passed, should not return an error
+	current.Status.BundlePushTracker.LastCheckInterval = time.Now().Unix() - 10
+	err = PerformCmBundlePush(client, &current)
+	if err != nil && strings.HasPrefix(err.Error(), "Will re-attempt to push the bundle after the 5 seconds") {
+		t.Errorf("Bundle Push Should not fail if reattempted after 5 seconds interval passed. Error: %s", err.Error())
+	}
+
+	// When the CM Bundle push is not pending, should not return an error
+	current.Status.BundlePushTracker.NeedToPushMasterApps = false
 	err = PerformCmBundlePush(client, &current)
 	if err != nil {
-		t.Errorf("Bundle Push failed")
+		t.Errorf("Should not return an error when the Bundle push is not required. Error: %s", err.Error())
 	}
 }
 
-func TestPushMasterAppsBundlePush(t *testing.T) {
+func TestPushMasterAppsBundle(t *testing.T) {
 
 	current := enterprisev1.ClusterMaster{
 		TypeMeta: metav1.TypeMeta{
@@ -296,6 +317,12 @@ func TestPushMasterAppsBundlePush(t *testing.T) {
 
 	client := spltest.NewMockClient()
 
+	//Without global secret object, should return an error
+	err := PushMasterAppsBundle(client, &current)
+	if err == nil {
+		t.Errorf("Bundle push should fail, when the secret object is not found")
+	}
+
 	secret, err := splutil.ApplyNamespaceScopedSecretObject(client, "test")
 	if err != nil {
 		t.Errorf(err.Error())
@@ -306,8 +333,10 @@ func TestPushMasterAppsBundlePush(t *testing.T) {
 		t.Errorf(err.Error())
 	}
 
+	//Without password, should return an error
+	delete(secret.Data, "password")
 	err = PushMasterAppsBundle(client, &current)
-	if err != nil {
-		t.Errorf("Bundle Push failed")
+	if err == nil {
+		t.Errorf("Bundle push should fail, when the password is not found")
 	}
 }
--- a/pkg/splunk/enterprise/configuration_test.go
+++ b/pkg/splunk/enterprise/configuration_test.go
@@ -21,6 +21,9 @@ import (
 
 	enterprisev1 "github.com/splunk/splunk-operator/pkg/apis/enterprise/v1beta1"
 	splcommon "github.com/splunk/splunk-operator/pkg/splunk/common"
+	splctrl "github.com/splunk/splunk-operator/pkg/splunk/controller"
+	spltest "github.com/splunk/splunk-operator/pkg/splunk/test"
+	splutil "github.com/splunk/splunk-operator/pkg/splunk/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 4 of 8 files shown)


pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +42,62 @@ func TestApplyConfigMap(t *testing.T) {
...
+			Name:      "defaults",
+			Namespace: "test",
+		},
+	}
+
+	client := spltest.NewMockClient()
+	namespacedName := types.NamespacedName{Namespace: current.GetNamespace(), Name: current.GetName()}
+
+	_, err := GetConfigMap(client, namespacedName)
+	if err == nil {
```
akondur, line 58: Nit: We could expect an error **"NotFound"** as follows:

```
if err.Error() != "NotFound" {
    t.Errorf("Failed to recognize non-existence of configMap") 
}
```
Also we could make **NotFound** as an entry in **pkg/splunk/common/messages.go**

pkg/splunk/client/enterprise.go
```diff
@@ -629,10 +629,7 @@ func (c *SplunkClient) DecommissionIndexerClusterPeer(enforceCounts bool) error
 }
 
 // BundlePush pushes the CM master apps bundle to all the indexer peers
-func (c *SplunkClient) BundlePush(ignoreIdenticalBundle bool, mock bool) error {
-	if mock {
-		return nil
-	}
+func (c *SplunkClient) BundlePush(ignoreIdenticalBundle bool) error {
```
akondur, line 632: QQ: Why did we have the mock bool in the first place if it weren't used?

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +42,62 @@ func TestApplyConfigMap(t *testing.T) {
...
```
akondur, line 85: Same as above.

pkg/splunk/controller/configmap_test.go
```diff
@@ -41,3 +42,62 @@ func TestApplyConfigMap(t *testing.T) {
...
+		t.Errorf("Should return an error, when the configMap doesn't exist")
+	}
+
+	_, err = GetConfigMapResourceVersion(client, namespacedName)
+	if err == nil {
```
akondur, line 90: Same as above.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -243,6 +244,13 @@ func TestPerformCmBundlePush(t *testing.T) {
 
 	client := spltest.NewMockClient()
 
+	// When the secret object is not present, should return an error
+	current.Status.BundlePushTracker.NeedToPushMasterApps = true
+	err := PerformCmBundlePush(client, &current)
+	if err == nil {
```
akondur, line 250: Same as above. Checking for a deterministic error might be better in all cases.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -261,23 +269,33 @@ func TestPerformCmBundlePush(t *testing.T) {
...
+	_, err = splctrl.ApplyConfigMap(client, &smartstoreConfigMap)
+
 	current.Status.BundlePushTracker.NeedToPushMasterApps = true
-	current.Status.BundlePushTracker.LastCheckInterval = time.Now().Unix() - 100
 
-	_, err = splctrl.ApplyConfigMap(client, &smartstoreConfigMap)
+	//Re-attempting to push the CM bundle in less than 5 seconds should return an error
+	current.Status.BundlePushTracker.LastCheckInterval = time.Now().Unix() - 1
+	err = PerformCmBundlePush(client, &current)
+	if err == nil {
```
akondur, line 279: Same as above.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -306,8 +330,10 @@ func TestPushMasterAppsBundlePush(t *testing.T) {
 		t.Errorf(err.Error())
 	}
 
+	//Without password, should return an error
+	delete(secret.Data, "password")
 	err = PushMasterAppsBundle(client, &current)
-	if err != nil {
-		t.Errorf("Bundle Push failed")
+	if err == nil {
```
akondur, line 336: Same as above.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+			},
+			{Name: "salesdata3", RemotePath: "remotepath3",
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol"},
+			},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreWithoutVolumes)
+	if err == nil {
```
akondur, line 483: Same as above.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+			},
+			{Name: "salesdata3", RemotePath: "remotepath3",
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol"},
+			},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreWithDuplicateVolumes)
+	if err == nil {
```
akondur, line 511: Same as above.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+			IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+				VolName: "msos_s2s3_vol-2"},
+		},
+		VolList: []enterprisev1.VolumeSpec{
+			{Name: "msos_s2s3_vol-1", Endpoint: "https://s3-eu-west-2.amazonaws.com", Path: "testbucket-rs-london", SecretRef: "s3-secret"},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreDefaultsWithNonExistingVolume)
+	if err == nil {
```
akondur, line 527: Same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+			},
+			{Name: "salesdata1", RemotePath: "remotepath2",
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol"},
+			},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreWithDuplicateIndexes)
+	if err == nil {
```
akondur, line 549: Same as above.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+			{Name: "salesdata1", RemotePath: "remotepath1"},
+			{Name: "salesdata2", RemotePath: "remotepath2",
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol"},
+			},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreVolumeMissingBothFromDefaultsAndIndex)
+	if err == nil {
```
akondur, line 569: Same as above.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -457,9 +460,199 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+		IndexList: []enterprisev1.IndexSpec{
+			{Name: "salesdata1", RemotePath: "remotepath1",
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol-2"},
+			},
+		},
+	}
+
+	err = ValidateSplunkSmartstoreSpec(&SmartStoreIndexesWithInvalidVolumeName)
+	if err == nil {
```
akondur, line 587: Same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -520,3 +720,128 @@ maxGlobalRawDataSizeMB = 61440
...
+				IndexAndGlobalCommonSpec: enterprisev1.IndexAndGlobalCommonSpec{
+					VolName: "msos_s2s3_vol"},
+			},
+		},
+	}
+
+	// Volume that doesn't should error out
+	_, err := checkIfVolumeExists(SmartStoreConfig.VolList, "random_volume_name")
+
+	if err == nil {
```
akondur, line 748: Same as above

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -520,3 +720,128 @@ maxGlobalRawDataSizeMB = 61440
...
+	client := spltest.NewMockClient()
+	var err error
+	ResourceRev := map[string]string{
+		"secret1": "12345",
+		"secret2": "67890",
+	}
+
+	// Missing secret object should return an error
+	keysChanged := AreRemoteVolumeKeysChanged(client, &cr, SplunkClusterMaster, &cr.Spec.SmartStore, ResourceRev, &err)
+	if err == nil {
```
akondur, line 799: Same as above. (You could avoid it here since we have an else condition)

pkg/splunk/enterprise/util_test.go
```diff
@@ -202,6 +202,13 @@ func TestApplySmartstoreConfigMap(t *testing.T) {
 	}
 
 	test(client, &cr, &cr.Spec.SmartStore, `{"metadata":{"name":"splunk-idxCluster--smartstore","namespace":"test","creationTimestamp":null},"data":{"conftoken":"1601945361","indexes.conf":"[default]\nrepFactor = auto\nmaxDataSize = auto\nhomePath = $SPLUNK_DB/$_index_name/db\ncoldPath = $SPLUNK_DB/$_index_name/colddb\nthawedPath = $SPLUNK_DB/$_index_name/thaweddb\n \n[volume:msos_s2s3_vol]\nstorageType = remote\npath = s3://testbucket-rs-london\nremote.s3.access_key = abcdJDckRkxhMEdmSk5FekFRRzBFOXV6bGNldzJSWE9IenhVUy80aa\nremote.s3.secret_key = g4NVp0a29PTzlPdGczWk1vekVUcVBSa0o4NkhBWWMvR1NadDV4YVEy\nremote.s3.endpoint = https://s3-eu-west-2.amazonaws.com\n \n[salesdata1]\nremotePath = volume:msos_s2s3_vol/remotepath1\n\n[salesdata2]\nremotePath = volume:msos_s2s3_vol/remotepath2\n\n[salesdata3]\nremotePath = volume:msos_s2s3_vol/remotepath3\n","server.conf":""}}`)
+
+	// Missing Volume config should return an error
+	cr.Spec.SmartStore.VolList = nil
+	_, _, err = ApplySmartstoreConfigMap(client, &cr, &cr.Spec.SmartStore)
+	if err == nil {
```
akondur, line 209: Same as above.

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+					{Name: "msos_s2s3_vol_3", Endpoint: "https://s3-eu-west-2.amazonaws.com", Path: "testbucket-rs-london", SecretRef: "splunk-test-sec_3"},
+					{Name: "msos_s2s3_vol_4", Endpoint: "https://s3-eu-west-2.amazonaws.com", Path: "testbucket-rs-london", SecretRef: "splunk-test-sec_4"},
+				},
+			},
+		},
+	}
+
+	// S3 secret owner reference removal, with non-existing secret objects
+	err = DeleteOwnerReferencesForS3SecretObjects(client, secret, &cr.Spec.SmartStore)
+	if err == nil {
```
akondur, line 295: Same as above

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+		t.Errorf("Should report an error, when the secret object referenced in the volume config doesn't exist")
+	}
+
+	// Smartstore volume config with non-existing secret objects
+	err = DeleteOwnerReferencesForResources(client, &cr, &cr.Spec.SmartStore)
+	if err == nil {
```
akondur, line 301: Same as above

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+	}
+
+	_, err = splctrl.ApplySecret(client, secret)
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+
+	// Missing S3 access key should return error
+	_, _, _, err = GetSmartstoreRemoteVolumeSecrets(cr.Spec.SmartStore.VolList[0], client, &cr, &cr.Spec.SmartStore)
+	if err == nil {
```
akondur, line 336: Same as above

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+		t.Errorf("Missing S3 access key should return an error")
+	}
+
+	secret.Data[s3AccessKey] = []byte("abcdJDckRkxhMEdmSk5FekFRRzBFOXV6bGNldzJSWE9IenhVUy80aa")
+
+	// Missing S3 secret key should return error
+	_, _, _, err = GetSmartstoreRemoteVolumeSecrets(cr.Spec.SmartStore.VolList[0], client, &cr, &cr.Spec.SmartStore)
+	if err == nil {
```
akondur, line 344: Same as above

akondur (commented):
@sgontla All the comments are wrt a single change i.e when checking for an err, being deterministic vs checking **err == nil**. Looks good otherwise.

pkg/splunk/controller/configmap_test.go
sgontla: That is good to have, but, it has its own drawbacks if at all we want to cover for all the error conditions. For example, a returned error string can be variable in length, values, in which case, we need to do string comparisons like suffix, prefix, contains etc..., which will drastically reduce the performance. For every returned error, we are anyhow printing the error, so we should be good IMO, and all we need is a good descriptive comment above each test case.

Added to that, looking at the coverage report, we can make sure that there are no holes(to the maximum extent possible).

akondur: Performance, functionality and code coverage wise a comment describing the expected error sounds good @sgontla.

pkg/splunk/client/enterprise.go
sgontla: It was used Arjun. Now, with this change, the caller will not cover a UT for this case for success, and to address it holistically, I put a separate Jira cspl-581, to cover both UT and also introducing the pod manager to the CM CR.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -261,23 +269,33 @@ func TestPerformCmBundlePush(t *testing.T) {
 		Data: map[string]string{configToken: ""},
 	}
 
+	_, err = splctrl.ApplyConfigMap(client, &smartstoreConfigMap)
```
kashok-splunk, line 272: Check to err here again? If err in cm creation

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+					{Name: "msos_s2s3_vol_4", Endpoint: "https://s3-eu-west-2.amazonaws.com", Path: "testbucket-rs-london", SecretRef: "splunk-test-sec_4"},
+				},
+			},
+		},
+	}
+
+	// S3 secret owner reference removal, with non-existing secret objects
+	err = DeleteOwnerReferencesForS3SecretObjects(client, secret, &cr.Spec.SmartStore)
+	if err == nil {
+		t.Errorf("Should report an error, when the secret object referenced in the volume config doesn't exist")
```
kashok-splunk, line 296: The error message can be "Error in owner ref removal as SecretRef %s doesn't exist"

pkg/splunk/enterprise/util_test.go
```diff
@@ -260,4 +270,85 @@ func TestRemoveOwenerReferencesForSecretObjectsReferredBySmartstoreVolumes(t *te
...
+	if err == nil {
+		t.Errorf("Missing S3 access key should return an error")
+	}
+
+	secret.Data[s3AccessKey] = []byte("abcdJDckRkxhMEdmSk5FekFRRzBFOXV6bGNldzJSWE9IenhVUy80aa")
+
+	// Missing S3 secret key should return error
+	_, _, _, err = GetSmartstoreRemoteVolumeSecrets(cr.Spec.SmartStore.VolList[0], client, &cr, &cr.Spec.SmartStore)
+	if err == nil {
+		t.Errorf("Missing S3 secret key should return an error")
```
kashok-splunk, line 345: Can we have a different message here, easy to debug which condition are we failing at

akondur (approved):

pkg/splunk/enterprise/clustermaster_test.go
sgontla: Addressed it.

pkg/splunk/enterprise/configuration_test.go
sgontla: Getting an error is expected, hence the explicit 'else' check for the keysChanged

pkg/splunk/enterprise/util_test.go
sgontla: Here, we are expecting an error, but if the test doesn't return an error, that is when we are printing this message. So, I guess, the message in the current form is reasonable.

pkg/splunk/enterprise/util_test.go
sgontla: Here, we are filling only the s3AccessKey, and leaving the s3SecretKey missing. Hence the error "Missing S3 secret key should return an error", which is the condition we are testing. There is a comment also added for this test case.

kashok-splunk (approved):

smohan-splunk (approved):

PR #180, opened 2020-11-24
CSPL-588: This PR improves code coverage for clustermaster specific code

gaurav-splunk:


pkg/splunk/enterprise/clustermaster.go
```diff
@@ -58,9 +58,6 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 	if !reflect.DeepEqual(cr.Status.SmartStore, cr.Spec.SmartStore) ||
 		AreRemoteVolumeKeysChanged(client, cr, SplunkClusterMaster, &cr.Spec.SmartStore, cr.Status.ResourceRevMap, &err) {
 
-		if err != nil {
-			return result, err
-		}
```
sgontla, line 63: Any reason we are removing this error check? This was introduced to handle the case when somebody accidentally deletes the secret object consisting the S4 keys.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -222,7 +233,59 @@ func TestApplyClusterMasterWithSmartstore(t *testing.T) {
...
+	client.AddObject(ss)
+	if _, err := ApplyClusterMaster(client, &current); err == nil {
+		t.Errorf("ApplyClusterMaster() should have returned error")
+	}
+
+	ss.Status.ReadyReplicas = 1
+	*ss.Spec.Replicas = ss.Status.ReadyReplicas
+	objects := []runtime.Object{ss, pod}
+	client.AddObjects(objects)
+	current.Spec.CommonSplunkSpec.Mock = false
```
sgontla, line 283: I think we are in the flow of removing the mock flag, and fixing those dependencies? In my earlier changes, I removed the mock dependencies, and that even created some holes in the coverage, so I raised a couple of stories to fix them formally.

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: I knew this would come up :). If you look at the code for `AreRemoteVolumeKeysChanged`, there is no case in that function where we are returning `true` and also returning `err` as well.

pkg/splunk/enterprise/clustermaster_test.go
gaurav-splunk: As discussed this thing touches MC code, hence I left it as is as it might require some more code changes on MC side.

pkg/splunk/enterprise/clustermaster.go
sgontla: Good point Gaurav. That was a miss i guess. In that case, we need to handle it differently such that we still return for that case.

akondur (approved):
Looks good!

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: yes, you are right! we should error out in that case. So in short, we need to take care of case where it returns false with error. Let me fix it and push a change again.

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -75,6 +72,11 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 		cr.Status.SmartStore = cr.Spec.SmartStore
 	}
 
+	// This is to take care of case where AreRemoteVolumeKeysChanged returns an error if it returns false.
+	if err != nil {
```
gaurav-splunk, line 76: This should take care of above comment.

sgontla (approved):

smohan-splunk (approved):

PR #186, opened 2020-11-26
Feaure/support init containers

rfaircloth-splunk:


kumarajeet:
It looks like some files are showing diff sections w.r.t. some indentation? Just wanted to confirm.

patrykw-splunk:
stale PR, we can close it. In case of mistake we can reopen it

PR #190, opened 2020-12-01
Eks use existing vpc

pdhanoyasplunk:
Updated the eks cluster create script to use existing VPC to avoid resource bottleneck on AWS.
- Added 2 environment variable in env.sh for existing public and private subnet string.
- Updated eksctl create command to use existing subnets.

### DIFF
--- a/test/env.sh
+++ b/test/env.sh
@@ -8,6 +8,8 @@
 : "${NUM_NODES:=2}"
 : "${COMMIT_HASH:=}"
 : "${ECR_REGISTRY:=}"
+: "${VPC_PUBLIC_SUBNET_STRING:=}"
+: "${VPC_PRIVATE_SUBNET_STRING:=}"
 
 # Docker registry to use to push the test images to and pull from in the cluster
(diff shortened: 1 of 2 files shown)


test/env.sh
```diff
@@ -8,6 +8,8 @@
 : "${NUM_NODES:=2}"
 : "${COMMIT_HASH:=}"
 : "${ECR_REGISTRY:=}"
+: "${VPC_PUBLIC_SUBNET_STRING:=subnet-0921cea9bcffd7b77,subnet-0dbbc27abdf4a416e,subnet-0dec6ad34b32e791f}"
```
smohan-splunk, line 11: Can these read from command line arguments instead of the hardcode?

pdhanoyasplunk: Thanks Sirish. Thats a really good point. I am removing the subnets and they can be set locally like ECR Registry.

smohan-splunk (approved):

akondur (approved):

Issue #185, opened 2020-11-26
support mgmt (8089) without TLS


ryanfaircloth:
inside of a k8s network there are some valid reasons to disable ssl including performance when external network encryption such as weave is applied. The following options correctly use ansible to deconfigure ssl on this port however operator connections presume ssl.

  defaults: |-
    cert_prefix: http
    splunk:
      ssl:
        enable: false

Use cases TLS offload with external ingress and ISTIO mutal TLS auth (offloaded TLS)

akondur:
Hi @rfaircloth-splunk , could you try using **defaults** or **defaultsUrl** [common spec parameter](https://github.com/splunk/splunk-operator/blob/master/docs/CustomResources.md#common-spec-parameters-for-splunk-enterprise-resources) to pass in the SSL options?

rfaircloth-splunk:
the defaults portion works however operator is hard coded to connect by TLS so it breaks the operator if the instance is not TLS enabled

jkoby:
This should work now, after a fix in the Docker image: https://github.com/splunk/docker-splunk/pull/488

thooooooomas:
I don't think splunk-operator is supporting TLS off as the original poster intended..
The PR that adds support #194 seems to have never merged..

It is still confusing reading this issue if splunk-operator supports TLS off or not for mgmt port, is it or is it not supported?

thooooooomas:
> This should work now, after a fix in the Docker image: [splunk/docker-splunk#488](https://github.com/splunk/docker-splunk/pull/488)

I believe this does not fix the issue that @ryanfaircloth mentioned in comment 3: that splunk-operator is hard coded to connect via https

MaxRenaud:
@jkoby Can you please comment on @thooooooomas 's comment above? We are also interested in this and can create a new issue or reopen this one.

PR #191, opened 2020-12-01
DO NOT MERGE feat: Allow configuration of TLS off for all ports

ryanfaircloth:
This change allows for configuration off of TLS for all records managed operator. This will allow for istio to manage port level security in an untrusted network and eliminate the need to manage certs in the container. In Addition updated port name patterns for istio other mesh layers to "discover" protocol via the port name.

This PR depends on https://github.com/splunk/docker-splunk/pull/437 
ref  #185


akondur:
PR is addressing the following [github issue](https://github.com/splunk/splunk-operator/issues/185).

PR #193, opened 2020-12-03
CSPL-534: Fix unnecessary pod recycle during scale up/down

gaurav-splunk:
This PR fixes the problem of unnecessary pod resets/recycles during scale up/down of splunk resources. As part of this solution, as soon as we try to scale up/down pods, we now update the `pod-revision-hash` label on pods to match `updateRevision` of StatefulSet. This prevents the recycle of pods based on `pod-revision-hash` mismatch of pods.

### DIFF
--- a/pkg/splunk/controller/statefulset.go
+++ b/pkg/splunk/controller/statefulset.go
@@ -135,6 +135,7 @@ func UpdatePodRevisionHash(c splcommon.ControllerClient, statefulSet *appsv1.Sta
 	return nil
 }
 
+// isRevisionUpdateSuccessful checks if current revision is different from updated revision
 func isRevisionUpdateSuccessful(c splcommon.ControllerClient, statefulSet *appsv1.StatefulSet) bool {
 	scopedLog := log.WithName("isUpdateSuccessful").WithValues(
 		"name", statefulSet.GetObjectMeta().GetName(),
@@ -158,6 +159,7 @@ func isRevisionUpdateSuccessful(c splcommon.ControllerClient, statefulSet *appsv
 	return true
 }
 
+// checkAndUpdatePodRevision updates the pod revision hash labels on pods if statefulset update was successful
 func checkAndUpdatePodRevision(c splcommon.ControllerClient, statefulSet *appsv1.StatefulSet, readyReplicas int32, skipRecheckUpdate *bool) (bool, error) {
 	scopedLog := log.WithName("UpdateStatefulSetPods").WithValues(
 		"name", statefulSet.GetObjectMeta().GetName(),
--- a/pkg/splunk/controller/statefulset.go
+++ b/pkg/splunk/controller/statefulset.go
@@ -34,8 +34,9 @@ type DefaultStatefulSetPodManager struct{}
 // Update for DefaultStatefulSetPodManager handles all updates for a statefulset of standard pods
 func (mgr *DefaultStatefulSetPodManager) Update(client splcommon.ControllerClient, statefulSet *appsv1.StatefulSet, desiredReplicas int32) (splcommon.Phase, error) {
 	phase, err := ApplyStatefulSet(client, statefulSet)
+	skipRecheckUpdate := false
 	if err == nil && phase == splcommon.PhaseReady {
-		phase, err = UpdateStatefulSetPods(client, statefulSet, mgr, desiredReplicas)
+		phase, err = UpdateStatefulSetPods(client, statefulSet, mgr, desiredReplicas, &skipRecheckUpdate)
 	}
 	return phase, err
 }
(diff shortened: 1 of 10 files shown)


kashok-splunk (approved):

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,102 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+			scopedLog.Error(err, "Waiting for Pod to become ready", "podName", podName)
+			return err
+		}
+
+		labels := pod.GetLabels()
+		revisionHash := labels["controller-revision-hash"]
+		// update pod controller revision hash
+		if current.Status.UpdateRevision != "" && current.Status.UpdateRevision != revisionHash {
+			// update controller-revision-hash label for pod with statefulset update revision
+			labels["controller-revision-hash"] = current.Status.UpdateRevision
```
jryb, line 123: This "controller-revision-hash" param is set to "v0" at pod bringup then gets updated when a scale-up/down event occurs?  Where is this "controller-revision-hash" updated other than here?  Is this typically done only when the pod is spun up?

pkg/splunk/controller/statefulset.go
```diff
@@ -112,7 +221,23 @@ func UpdateStatefulSetPods(c splcommon.ControllerClient, statefulSet *appsv1.Sta
...
+		err := splutil.UpdateResource(c, statefulSet)
+		if err != nil {
+			scopedLog.Error(err, "Unable to update statefulset")
+			return splcommon.PhaseError, err
+		}
+
+		// Check if the revision was updated successfully for Statefulset.
+		// It so can happen that it may take few seconds for the update to be
+		// reflected in the resource. In that case, just return from here and
+		// check the status back in the next reconcile loop.
```
jryb, line 233: nit: "It can happen"
If this check can fail and its not big deal, we'll just check it at the next reconcile, do we need it?

pkg/splunk/controller/statefulset.go
```diff
@@ -139,6 +264,17 @@ func UpdateStatefulSetPods(c splcommon.ControllerClient, statefulSet *appsv1.Sta
 			return splcommon.PhaseError, err
 		}
 
+		// Check if the revision was updated successfully for Statefulset.
+		// It so can happen that it may take few seconds for the update to be
+		// reflected in the resource. In that case, just return from here and
+		// check the status back in the next reconcile loop.
+		var ret bool
+		ret, err = checkAndUpdatePodRevision(c, statefulSet, readyReplicas-1, skipRecheckUpdate)
+		if !ret || err != nil {
```
jryb, line 273: Can we move this check outside of the scale up/scale down checks since its common for both.  Though I guess the `readyReplicas` is different by 1.

jryb (commented):
With this change does that mean we don't recycle the running pods when we scale up/down? Or do we just recycle the pods less?

pkg/splunk/controller/statefulset.go
gaurav-splunk: I dont think so it is "v0" during bring up. I checked in my local env and it has definite hash value,probably similar to updateRevision of statefulset. And yes it gets updated during any update to statefulset, not necessarily scale up/down only( Check `ApplyStatefulSet`).

pkg/splunk/controller/statefulset.go
gaurav-splunk: During the next reconcile we'll landup either in line 196 or 208, since we have already updated `replicas` here in this reconcile.

pkg/splunk/controller/statefulset.go
gaurav-splunk: Yes readyReplicas was different in both cases hence did not move it out.

pkg/splunk/controller/statefulset.go
jryb: nm, I was looking at the Tester code.

jryb (approved):

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,102 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+			}
+			scopedLog.Info("Updated controller-revision-hash label for pod", "podName", podName, "revision", current.Status.UpdateRevision)
+		}
+	}
+
+	scopedLog.Info("Updated controller-revision-hash label for all pods")
+	return nil
+}
+
+func isRevisionUpdateSuccessful(c splcommon.ControllerClient, statefulSet *appsv1.StatefulSet) bool {
```
akondur, line 138: Add a comment here?

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,102 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+	// Check if current revision is different from update revision
+	if current.Status.CurrentRevision == current.Status.UpdateRevision {
+		scopedLog.Error(err, "Statefulset UpdateRevision not updated yet")
+		return false
+	}
+
+	return true
+}
+
+func checkAndUpdatePodRevision(c splcommon.ControllerClient, statefulSet *appsv1.StatefulSet, readyReplicas int32, skipRecheckUpdate *bool) (bool, error) {
```
akondur, line 161: Add a comment here?

akondur (approved):
Great PR @gaurav-splunk

pkg/splunk/controller/statefulset_test.go
```diff
@@ -57,6 +60,181 @@ func TestDefaultStatefulSetPodManager(t *testing.T) {
...
+	err := updatePodRevisionHashTester(t, statefulSet, replicas, statefulSet, pod)
+	if err != nil {
+		t.Errorf("UpdatePodRevisionHash should not have returned error")
+	}
+
+	// Test the negative case where phase != corev1.PodRunning
+	pod.Status.Phase = corev1.PodPending
+	err = updatePodRevisionHashTester(t, statefulSet, replicas, statefulSet, pod)
+	if err != nil {
+		t.Errorf("UpdatePodRevisionHash should not have returned error")
```
sgontla, line 226: You may want to print the error string also. Same comment for other places.

sgontla (changes requested):

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,104 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+		"namespace", statefulSet.GetObjectMeta().GetNamespace())
+
+	namespacedName := types.NamespacedName{Namespace: statefulSet.GetNamespace(), Name: statefulSet.GetName()}
+	var current appsv1.StatefulSet
+
+	err := c.Get(context.TODO(), namespacedName, &current)
+	if err != nil {
+		scopedLog.Error(err, "Unable to Get statefulset", "statefulset", statefulSet.GetName())
+	}
+
```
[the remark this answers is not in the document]
gaurav-splunk: Yeah good catch!

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,104 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+		}
+		if pod.Status.Phase != corev1.PodRunning || len(pod.Status.ContainerStatuses) == 0 || pod.Status.ContainerStatuses[0].Ready != true {
+			scopedLog.Error(err, "Waiting for Pod to become ready", "podName", podName)
+			return err
+		}
+
+		labels := pod.GetLabels()
+		revisionHash := labels["controller-revision-hash"]
+		// update pod controller revision hash
+		if current.Status.UpdateRevision != "" && current.Status.UpdateRevision != revisionHash {
```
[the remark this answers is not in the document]
gaurav-splunk: If we error out from line 98, I don't think so we would have `current` as nil, right?

pkg/splunk/controller/statefulset.go
```diff
@@ -85,8 +86,104 @@ func ApplyStatefulSet(c splcommon.ControllerClient, revised *appsv1.StatefulSet)
...
+	// update the controller-revision-hash label on pods to
+	// to avoid unnecessary recycle of pods
+	err = UpdatePodRevisionHash(c, statefulSet, readyReplicas)
+	if err != nil {
+		scopedLog.Error(err, "Unable to update pod-revision-hash for the pods")
+		*skipRecheckUpdate = false
+		return false, err
+	}
+	*skipRecheckUpdate = true
+	return true, nil
```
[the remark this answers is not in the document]
gaurav-splunk: We need to update `skipRecheckUpdate` here for further reconciles. Return value is independent of it.

pkg/splunk/controller/statefulset.go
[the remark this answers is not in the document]
sgontla: 'current' case is covered with the above return. Here the comment is for the revisionHash :-)

pkg/splunk/controller/statefulset.go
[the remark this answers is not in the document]
sgontla: Ahh ok. In that case, we can just use the 'skipRecheckUpdate' for the return value also?

pkg/splunk/controller/statefulset.go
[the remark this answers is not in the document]
gaurav-splunk: Ok, but controller-revision-hash is never empty anyways :)

sgontla (approved):

smohan-splunk (approved):

PR #194, opened 2020-12-04
feat: Allow configuration of default TLS in Splunk

ryanfaircloth:
By default Splunk will enable SSL/TLS for splunkd (8089) and hec (8088). It can be desirable
to offload encryption between pods to a the network layer using CNI and externally using ingress
or istio. The following environment variables can be added to the operator deployment to disable
default SSL/TLS

This PR depends on splunk/docker-splunk#437
ref #185


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -445,8 +445,8 @@ func (mgr *indexerClusterPodManager) getClient(n int32) *splclient.SplunkClient
 	// Get Pod Name
 	memberName := GetSplunkStatefulsetPodName(SplunkIndexer, mgr.cr.GetName(), n)
 
-	// Get Fully Qualified Domain Name
-	fqdnName := splcommon.GetServiceFQDN(mgr.cr.GetNamespace(),
+	// Get uri
```
mgaldino-splunk, line 448: I was wondering if the change from FQDN to URI here is required for disabling the default SSL, or if this change is an additional enhancement?

mgaldino-splunk (commented):
The PR looks good in my opinion, I would wait until other folks review it as well prior to approving it.

thooooooomas:
What was the outcome of this PR? Was it merged? Or is this not coming to splunk-operator?

Issue #189, opened 2020-12-01
Monitoring console broken in complex installations


ryanfaircloth:
Because the deployment for  MC is auto-generated the admin can not pass critical information into the template such as licensemasterref

Looking at the pod deployment it looks like this would not work correctly if there was more than one indexer cluster in the deployment in two ways the admin can not control which indexer cluster is used for output and it can not define multiple indexer clusters for search. The ask is as follows
Define a new CRD for Monitoring Console that also uses licenseref and indexerref
on each defined CR i.e cluster master add a mcref field to specify which monitoring console should monitor this deployment.
Ensure the ref syntax support namespace/object pattern such that MC can monitor across namepsaces

romain-bellanger:
Hello,
@rfaircloth-splunk I think this is related to the problem which you describe, but please let me know if you prefer this to be documented in a separate ticket.

### Problem description

This issue has other impacts: no volumes are defined on monitoring console pods, files created or change in etc and var are written directly in the container's layer. I think the impact depends on the used [Docker storage driver](https://docs.docker.com/storage/storagedriver/select-storage-driver/), but on our side the containers root FS is of type overlay, which doesn't seem to be supported by Splunk:

```
oc rsh splunk-splunk-dev-monitoring-console-0 df -hT var
Filesystem     Type     Size  Used Avail Use% Mounted on
overlay        overlay   60G  8.4G   52G  14% /
```

At startup Splunk fails to start and reports **unusable filesystem**:
```
STDERR:

homePath='/opt/splunk/var/lib/splunk/audit/db' of index=_audit on unusable filesystem.
Validating databases (splunkd validatedb) failed with code '1'.  If you cannot resolve the issue(s) above after consulting documentation, please file a case online at http://www.splunk.com/page/submit_issue
```

As described in this [Splunk answers ticket](https://community.splunk.com/t5/Getting-Data-In/Why-am-I-getting-quot-homePath-opt-splunk-var-lib-splunk-audit/m-p/220230), patching the splunk-launch.conf before Splunk starts allows the pod to start successfully.
```bash
echo 'echo "OPTIMISTIC_ABOUT_FILE_LOCKING = 1" >>etc/splunk-launch.conf' | oc rsh splunk-splunk-dev-monitoring-console-0
```

[...]

PR #145, opened 2020-10-08
[CSPL-486] Update MC configMap for  "SPLUNK_MULTISITE_MASTER" and "SPLUNK_SITE" env var

kashok-splunk:
**Scenario:**
 - Multisite-IDXC is running.
 - Create a standalone instance, this changes the env variables on the MC pod and MC loses the multi-site info

**Solution:**
- Append "SPLUNK_MULTISITE_MASTER" and "SPLUNK_SITE" env for MC in MC configMap
- Set multiple owners for the stateful set

Issue #126, opened 2020-09-22
Trigger app deployment on CRD update


romain-bellanger:
Hello,

* This ticket is partially related to issue [Operator deployment of apps](https://github.com/splunk/splunk-operator/issues/43)
* I also opened a ticket to [address some limitations of the app installation solution in splunk-ansible](https://github.com/splunk/splunk-ansible/issues/545)

The main issue which we experience at operator level with the deployment of Splunk apps is that the deployment, either on indexer cluster or search-head cluster, is not triggered automatically when applying a new version of the CRD blueprint. It requires to delete some Kubernetes resources, which complexify the integration to a CD pipeline (and is not supported at all in our CD pipeline). There are two situations:

**1. Configuring the apps through the ```defaults``` property of the CRD:**

The operator creates a ConfigMap with the content of the default properties and mount it to the pods. To trigger a new deployment, ansible needs to be triggered on the cluster-master or the deployer, which can currently be done by deleting the pods.

**2. Provide the URL of a config file in the ```defaultsUrl``` property:**

**This is our preferred method**, because this simplifies the CRD templates, and we can release a versioned ansible config file to our artifact repository within the apps build/CI pipeline together with the apps.
When using this method, the ```defaultsUrl``` property of the CRD is mapped to the environment variable ```SPLUNK_DEFAULTS_URL``` of the managed StatefulSets. The problems are that:
* The StatefulSets are not updated when modifying the ```defaultsUrl``` property of the CRD
* On SearchHeadClusters CRDs, the ```defaultsUrl``` property of the CRD is shared by the StatefulSets managing the deployer and the search-heads. If the StatefulSets were updated, this would trigger a rolling rebuild of the search-head pods, while the deployer would trigger an apply shcluster-bundle, potentially also triggering a rolling restart of the searchhead cluster in parallel.
**Note:** This problem does not really exist on IndexerCluster CRD as the cluster-master can have a dedicated CRD, so with different ```defaultsUrl``` from the indexers.

A possible solution would be to ensure the update of the ```SPLUNK_DEFAULTS_URL``` variable in StatefulSet, and to have a new property ```defaultsUrlDeployer``` in the SearchHeadClusters CRD to control the variable on the deployer StatefulSet independently from the searchheads StatefulSet, to avoid triggering a rebuild of SearchHead pods (if ```defaultsUrlDeployer``` not defined, the deployer could still use the value from ```defaultsUrl``` for backward compatibility).

[...]

PR #200, opened 2020-12-09
Update StatefulSets on environment variable change

romain-bellanger:
This piece of code was actually added by @kashok-splunk in PR #120 (diff [link](https://github.com/splunk/splunk-operator/pull/120/files#diff-f051fa8775a356c08c90ac411f88b50e7455ec17e8f3241a810b64eb7f00a747)), but removed in PR #145 (diff [link](https://github.com/splunk/splunk-operator/pull/145/files#diff-f051fa8775a356c08c90ac411f88b50e7455ec17e8f3241a810b64eb7f00a747)) because it was not required anymore for its initial purpose, according to the PR comments.

However this also brought significant improvements for application load as described in issue #126. Thanks to this change, it was not required anymore to delete StatefulSets manually to trigger app deployments after updating the ```defaultsUrl``` property of ClusterMaster or SearchHeadCluster CRDs, as the change of ```SPLUNK_DEFAULTS_URL``` triggered the update of the associated StatefulSets. We were able to put a continuous deployment pipeline in place thanks to this behavior, but it does not work anymore with the operator beta version 0.2.0.

**Possible impacts**:

We noticed some impact on the monitoring-console pod, when the ```defaultsUrl``` property is defined in some ClusterMaster or SearchHeadCluster resources. As the reconciliation of any Splunk custom resource calls the code building and applying the StatefulSet of the monitoring-console, and provides its own specs to build it, having a different ```defaultsUrl``` in some ClusterMaster and/or SearchHeadCluster resources seems to cause the monitoring-console StatefulSet to be constantly updated and switching its specs... This impact seems to be limited to the usage of ```defaultsUrl```, as the other variables of the MC StatefulSet are not impacted by the custom resource spec.

This code could be bypassed in the case of a monitoring console, but as the pod metadata is not passed to this method (labels not available), this would have to be done with a loop checking for environment variable ```SPLUNK_ROLE```, which would be a bit dirty. From our perspective this solution of monitoring-console inheriting its specs from a "random" resource has many other impacts and should be reviewed, as described in #189.

### DIFF
--- a/pkg/splunk/controller/util.go
+++ b/pkg/splunk/controller/util.go
@@ -149,6 +149,15 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name 
 				result = true
 			}
 
+			// check Env
+			if splcommon.CompareEnvs(current.Containers[idx].Env, revised.Containers[idx].Env) {
+				scopedLog.Info("Pod Container Env differ",
+					"current", current.Containers[idx].Env,
+					"revised", revised.Containers[idx].Env)
+				current.Containers[idx].Env = revised.Containers[idx].Env
+				result = true
+			}
+
 			// check VolumeMounts
 			if splcommon.CompareVolumeMounts(current.Containers[idx].VolumeMounts, revised.Containers[idx].VolumeMounts) {
 				scopedLog.Info("Pod Container VolumeMounts differ",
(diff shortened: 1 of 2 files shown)


gaurav-splunk:
@romain-bellanger We have a similar PR from @jryb for the same issue. Here is the link to the PR- https://github.com/splunk/splunk-operator/pull/199 :)

romain-bellanger:
Closed as duplicated of #199

PR #201, opened 2020-12-09
Monitoring-console: mount emptyDir to remove dependency on Docker storage driver

romain-bellanger:
This addresses one of the problems described in #189 (link to [the comment](https://github.com/splunk/splunk-operator/issues/189#issuecomment-740547576)).

On the monitoring console, unlike other resources, the ```/opt/splunk/etc``` and ```/opt/splunk/var``` directories are located on the root filesystem, so writes in the container layer. The problem is that the type of the root FS depends on the configured Docker storage driver. In the case of overlay, Splunk doesn't recognize the FS type and fails to start, reporting indexes being placed on unusable FS.

Mounting emptyDirs to these paths solves the problem (even though the long term solution would probably be to create a new CRD as discussed).


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -147,6 +147,10 @@ func getMonitoringConsoleStatefulSet(client splcommon.ControllerClient, cr splco
 		labels[k] = v
 	}
 
+	emptyVolumeSource := corev1.VolumeSource{
+		EmptyDir: &corev1.EmptyDirVolumeSource{},
+	}
+
```
jryb, line 153: It looks like this would happen for the MC if `spec.EphemeralStorage` was set to `true` in the common function `getSplunkStatefulSet()`.  Now since, as you pointed out, the MC doesn't have it's own CR, we cannot set this value for just the MC.  We could move this check to the common function:
```
func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, instanceType InstanceType, replicas int32, extraEnv []corev1.EnvVar) (*appsv1.StatefulSet, error) {
...
	// update template to include storage for etc and var volumes
	if spec.EphemeralStorage || type == SplunkMonitoringConsole {
		// add ephemeral volumes to the splunk pod for etc and opt
		emptyVolumeSource := corev1.VolumeSource{
			EmptyDir: &corev1.EmptyDirVolumeSource{},
		}
		statefulSet.Spec.Template.Spec.Volumes = []corev1.Volume{
			{Name: "mnt-splunk-etc", VolumeSource: emptyVolumeSource},
			{Name: "mnt-splunk-var", VolumeSource: emptyVolumeSource},
		}
```
But it's probably best served in the MC specific function until a MC CRD is created.

jryb (approved):

akondur (approved):
@romain-bellanger Looks good. 

Nit: I raised a [PR](https://github.com/splunk/splunk-operator/pull/219) revamping storageclass config for generic Splunk Enterprise CRDs. The nomenclature for **mnt-splunk-*** could be changed to **fmt.Sprintf(splcommon.SplunkMountNamePrefix, splcommon.EtcVolumeStorage)** & **fmt.Sprintf(splcommon.SplunkMountNamePrefix, splcommon.VarVolumeStorage)** in case you commit after my PR is merged.

pkg/splunk/enterprise/monitoringconsole.go
romain-bellanger: Hi @jryb, thanks a lot for your review. On top of this, at the moment the common function ```getSplunkStatefulSet()``` used by all the other resources is not called for the MC. For sure it would be nice to have a dedicated CRD for MC and to reuse the same code.

romain-bellanger:
Hi @akondur , thank you for the review. I've just checked your PR as well, and I don't think there is any risk of conflict, because for MC the storage class doesn't depend on any CRD property, it is not configurable. So far, it was not possible to mount any volume to /opt/splunk/etc and /opt/splunk/var (was always writing to the container layer), and with this PR there will always be an EmptyDir mounted there. It's hardcoded. I think that this would only change if a dedicated CRD is created for the MC.
But thank you for highlighting this, it's very good to know because we will have to adapt our custom resources before upgrading to the next version of the operator.

akondur:
> Hi @akondur , thank you for the review. I've just checked your PR as well, and I don't think there is any risk of conflict, because for MC the storage class doesn't depend on any CRD property, it is not configurable. So far, it was not possible to mount any volume to /opt/splunk/etc and /opt/splunk/var (was always writing to the container layer), and with this PR there will always be an EmptyDir mounted there. It's hardcoded. I think that this would only change if a dedicated CRD is created for the MC.
> But thank you for highlighting this, it's very good to know because we will have to adapt our custom resources before upgrading to the next version of the operator.

@romain-bellanger The changes look solid and I concur with you that there would be no conflicts with existing CRDs storage volumes. My only comment was regarding using templated strings for **mnt-splunk-***  vs hardcoding it for names of volumes, volumeMounts.

gaurav-splunk:
Hi @romain-bellanger , do you plan to write a unit test too for this change?

romain-bellanger:
Hi, I'm sorry for the late reply. The purpose of this PR was to fix a bug, not trying to improve the existing:

- @akondur: the code for the volumes and volume mounts is a copy/paste from [the generic function getSplunkStatefulSet()](https://github.com/splunk/splunk-operator/blob/develop/pkg/splunk/enterprise/configuration.go#L376) when CRD property spec.EphemeralStorage is defined. This function is called for all the other components, and the names are hardcoded there as well.
- @gaurav-splunk: there was no existing unit test for the StatefulSet generated for the monitoring console. I thought it would be a bit different from the other unit tests as the monitoring console has no dedicated CRD and its creation is triggered by other resources, so maybe a bit more tricky, and I understood that this is going to be redesigned with the creation of a dedicated CRD and using the common getSplunkStatefulSet() function, so I was not sure that it is worth to spend time on this... But you're right it's better to do it anyway, I just added a commit with a test.

smohan-splunk (approved):

PR #212, opened 2020-12-17
CSPL-564: Attach CM and Indexer Cluster to LM

gaurav-splunk:
**Problem**
Before this change, for an IndexerCluster to connect to LicenseMaster(LM), we had to explicitly specify `licenseMasterRef` in the spec section of IndexerCluster CRD to point it to LM. And similarly for CM too. But, ideally, if indexer cluster is connected to CM, then it should get the LM info from CM and connect it from there.

**Solution**
The above problem is tacked by following changes in code -  
1. Add ClusterMaster as a resource to watch for in IndexerCluster controller.
2. Make IndexerCluster as owner of CM, so that if CM is now connected to LM, indexer cluster will watch for that change and now it will also connect to LM.


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -154,6 +154,13 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		cr.Status.IdxcPasswordChangedSecrets = make(map[string]bool)
 
 		result.Requeue = false
+		// Set indexer cluster CR as owner reference for clustermaster
+		namespacedName = types.NamespacedName{Namespace: cr.GetNamespace(), Name: GetSplunkStatefulsetName(SplunkClusterMaster, cr.Spec.ClusterMasterRef.Name)}
+		err = splctrl.SetStatefulSetOwnerRef(client, cr, namespacedName)
```
akondur, line 159: NIT: Should we rename this to SetStatefulSetOwnerRef?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -154,6 +154,13 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterprisev1.Ind
 		cr.Status.IdxcPasswordChangedSecrets = make(map[string]bool)
 
 		result.Requeue = false
+		// Set indexer cluster CR as owner reference for clustermaster
+		namespacedName = types.NamespacedName{Namespace: cr.GetNamespace(), Name: GetSplunkStatefulsetName(SplunkClusterMaster, cr.Spec.ClusterMasterRef.Name)}
```
akondur, line 158: Why don't we set the owner reference of ClusterMaster CR vs the ClusterMaster statefulSet? Not sure how its working now.

I thought we are adding ClusterMaster as a watch type & if the ClusterMaster CR is owned by the indexercluster CR we reconcile the indexercluster CR on a change in the Clustermaster CR.

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: isn't it already `SetStatefulSetOwnerRef`? :)

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: I think any change in CM CR is eventually a change in CM statefulset which will trigger the reconcile of indxc. So, I think it should be ok, right? Or am I missing something basic here?

pkg/splunk/enterprise/configuration.go
```diff
@@ -639,7 +640,29 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		if spec.ClusterMasterRef.Namespace != "" {
 			clusterMasterURL = splcommon.GetServiceFQDN(spec.ClusterMasterRef.Namespace, clusterMasterURL)
 		}
+		//Check if CM is connected to a LicenseMaster
```
smohan-splunk, line 643: Does this also work with [External License Master](https://splunk.github.io/splunk-operator/Examples.html#using-an-external-license-master)?

pkg/splunk/enterprise/indexercluster.go
akondur: @gaurav-splunk Sorry my bad. Corrected a typo. I meant **SetStatefulOwnerRef**. I only meant to avoid the word set twice.

pkg/splunk/enterprise/indexercluster.go
akondur: If we are adding the IDXC CR as an OwnerReference to the **CM statefulSet** then we don't need to add the ClusterMaster watch type for IDXC CR as it is already watching all statefulSets(by the virtue of statefulSet as a watchtype) and will get reconciled for any changes in a StatefulSet owned by it. 

A change in the CM CR doesn't necessarily lead to a change in the CM statefulSet? Inverse of which might be true as well i.e there can be a change in CM statefulSet without a change in the CM CR. 

IMO we should review both scenarios and do either or both of these:

1. Add CM CR as a watch type and add IDXC CR as an OwnerReference to the CM CR
2. Add IDXC CR as an owner reference to the CM CR's statefulSet.

For the LM ref case, point 1 above should suffice.

pkg/splunk/enterprise/configuration.go
gaurav-splunk: TBH, I haven't tried with that. This PR just focusses on LM created with splunk operator as of now.

smohan-splunk: Ok, in case we find it not working, I suggest we create a separate bug which can then be picked up once CSPL-507 is fixed

pkg/splunk/enterprise/indexercluster.go
gaurav-splunk: As discussed offline, we are going with option 2, as we cannot set a CR as owner ref unless it is a default k8s resoruce.

pkg/splunk/enterprise/indexercluster.go
akondur: Will be taken up later

akondur (approved):

smohan-splunk (approved):

Issue #213, opened 2020-12-19
Only first pod of StatefulSet updated on modification of SearchHeadCluster or IndexerCluster (e.g. image, resources)


romain-bellanger:
Hello,

This is an impact of #193: on change of SearchHeadCluster or IndexerCluster custom resource properties such as image or resource quotas, only the first pod of the associated StatefulSet is updated. The purpose of this PR was precisely to update the ```controller-revision-hash``` label of all the pods to prevent their restart.

We are currently working on the upgrade of clusters in Splunk 8.0.6 deployed with the operator alpha version 0.1.0, to Splunk 8.1.0.1 with the operator beta version 0.2.1. This breaks a lot of automation in the upgrade process. When changing the docker image, pods were previously restarted sequentially with a proper offline (for indexers), which was matching the requirements for a clean upgrade (with just a couple of additional API calls to initialize / finalize the process). With this change only the first pod is restarted automatically. The upgrade of the remaining pods must be done by running offline from inside the container and deleting the pod manually.

The same impact probably exists on all the specs from custom resources inherited by the pods (I tested for instance modifying the CPU limit, again only the first pod was updated).

For now, we patched the operator by commenting the [first call to checkAndUpdatePodRevision](https://github.com/splunk/splunk-operator/pull/193/files#diff-28693ad8a99405feca6ef7afadb73ea0c728cc225b46becfbf109e784a9cf4f8R198) from method ```UpdateStatefulSetPods``` (case of ```readyReplicas < replicas```), which fixed the problem. But we don't understand the impact of this modification, as we don't understand the initial purpose of the PR. We never observed cases of unnecessary pod recycle on scaling. Could you please describe these scenarios, and look for a way to address them without causing such an impact on change of CRD specs?

gaurav-splunk:
Hi @romain-bellanger Thanks for finding the issue. Let me check it at my end and I'll let you know about the findings and fix at my end.
> But we don't understand the impact of this modification, as we don't understand the initial purpose of the PR. We never observed cases of unnecessary pod recycle on scaling.

We observed in our environment that whenever we tried to scale up/down indexer cluster, it recycled the indexer pods since the `controller-revision-hash` has changed and in our code, we were recycling the pods if there is a change in that label. I am not sure how you tried scaling. We tried using `kubectl scale idxc <idxc-name> --replicas=<new replica count>`.

gaurav-splunk:
[...]

PR #214, opened 2020-12-22
CSPL-694: Fix skipping of pod recycle during changes to CR spec

gaurav-splunk:
This PR targets the [bug](https://github.com/splunk/splunk-operator/issues/213) introduced during the implementation of [193](https://github.com/splunk/splunk-operator/pull/193), where if you change the CR spec(such as `image` in the case of an upgrade), then only the latest pod gets recycled to come up with the updated version but not the other replicas.

To fix this we have now introduced another field `IsScaling` in `Status` field of indexercluster and search head cluster statefulsets to track if we are in the process of scaling and only then update the pod-revision-hash label for the pods, else just go through with the recycling of pods.


smohan-splunk:
Closing this PR, updated implementation merged as part of https://github.com/splunk/splunk-operator/pull/244

PR #218, opened 2021-01-05
Added option to pass test selection env variable through CIRCLE CI

pdhanoyasplunk:
* Added keyword to test cases so that they can be filtered by ginkgo
* Added a nightly build workflow to cirlce CI config
* Change integration test to smoke test to represent the running test cases
* Modified test selection logic in run_test.sh

### DIFF
--- a/test/example/example_suite_test.go
+++ b/test/example/example_suite_test.go
@@ -14,7 +14,7 @@ import (
 
 var (
 	testenvInstance *testenv.TestEnv
-	testSuiteName   = "example-suite-" + testenv.RandomDNSName(6)
+	testSuiteName   = "example-" + testenv.RandomDNSName(3)
 )
 
 func init() {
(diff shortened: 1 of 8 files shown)


test/example/example_suite_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	testenvInstance *testenv.TestEnv
-	testSuiteName   = "example-suite-" + testenv.RandomDNSName(6)
+	testSuiteName   = "example-" + testenv.RandomDNSName(3)
```
gaurav-splunk, line 17: why change from 6 to 3?

pdhanoyasplunk: Ginkgo throws an error total length kubernetes artifacts go above 24 chars. This wasn't a problem before as we were simply skipping this suite. Now we are initializing and skipping hence an error is thrown. I am also inclining towards removing this suite all together as it serves no purpose.

smohan-splunk (approved):

gaurav-splunk (approved):

PR #219, opened 2021-01-06
CSPL-660: Storage class revamp

akondur:
- Changed the way storage class specifications are configured. Introduced two new fields **etcVolumeStorageConfig(for /opt/splunk/etc)** and **varVolumeStorageConfig(for /opt/splunk/var)** with **storageClassName**, **storageCapacity** & **ephemeralStorage** nested under them. 
- Changed docs to reflect the same
- Unit test to support
- Functional tests done to verify different combinations of storageCapacity, storageClassName combinations for the two volumes.


deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -660,17 +660,27 @@ spec:
...
+                until the SHC Deployer has its own CRD.  The defaults listed here
+                will only be installed on a standalone, or a SHC Deployer or IDC CM
+                to be pushed to SH/IDXs in a cluster. This parameter is ignored on
+                individual SHs or IDXs within a cluster
               type: string
+            etcVolumeStorageConfig:
+              description: Storage configuration for /opt/splunk/etc volume
+              properties:
+                ephemeralStorage:
+                  description: If true, ephemeral (emptyDir) storage will be used
```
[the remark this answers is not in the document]
akondur: The default for any boolean property is **false**. Do we still need to mention it?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -660,17 +660,27 @@ spec:
...
+                  type: boolean
+                storageCapacity:
+                  description: Storage capacity to request persistent volume claims
+                    (default=”10Gi” for etc and "100Gi" for var)
+                  type: string
+                storageClassName:
```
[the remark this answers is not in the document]
akondur: It is mentioned in the [documentation](https://github.com/splunk/splunk-operator/blob/develop/docs/StorageClass.md#setting-up-a-persistent-storage-for-splunk) as **"If no storageClassName is provided, the default Storage Class for your Kubernetes cluster will be used."**

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -1134,10 +1141,20 @@ spec:
...
-            varStorage:
-              description: Storage capacity to request for /opt/splunk/var persistent
-                volume claims (default=”50Gi”)
-              type: string
+            varVolumeStorageConfig:
+              description: Storage configuration for /opt/splunk/var volume
+              properties:
+                ephemeralStorage:
+                  description: If true, ephemeral (emptyDir) storage will be used
+                  type: boolean
```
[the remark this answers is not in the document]
akondur: Same comment as above.

docs/StorageClass.md
```diff
@@ -11,10 +11,28 @@ deployments. For each pod, the following two volumes will be mounted:
...
+metadata:
+  name: example
+  finalizers:
+  - enterprise.splunk.com/delete-pvc
+spec:
+  etcVolumeStorageConfig:
+    storageClassName: gp2
+    storageCapacity: 15Gi
+  varVolumeStorageConfig:
+    storageClassName: gp2
```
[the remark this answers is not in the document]
akondur: IMO it is good to provide flexibility at the cost of configuring the storageClassName twice(if the same one is used for etc and var). If the default storageClass for the cluster is used, there is no need to configure it which makes it redundant config.

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -660,17 +660,27 @@ spec:
...
-              type: string
-            ephemeralStorage:
-              description: If true, ephemeral (emptyDir) storage will be used for
-                /opt/splunk/etc and /opt/splunk/var volumes
-              type: boolean
-            etcStorage:
-              description: Storage capacity to request for /opt/splunk/etc persistent
-                volume claims (default=”1Gi”)
+              description: Full path or URL for one or more default.yml files specific
+                to App install, separated by commas This is meant as a temporary fix
```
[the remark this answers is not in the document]
jryb: That works with me.  So this would be something like:
```
              description: Full path or URL for one or more default.yml files specific
                to App install, separated by commas.  The defaults listed here
                will be installed on the CM instance.
```

deploy/crds/enterprise.splunk.com_searchheadclusters_crd.yaml
```diff
@@ -682,17 +682,27 @@ spec:
...
-              type: string
-            ephemeralStorage:
-              description: If true, ephemeral (emptyDir) storage will be used for
-                /opt/splunk/etc and /opt/splunk/var volumes
-              type: boolean
-            etcStorage:
-              description: Storage capacity to request for /opt/splunk/etc persistent
-                volume claims (default=”1Gi”)
+              description: Full path or URL for one or more default.yml files specific
+                to App install, separated by commas This is meant as a temporary fix
```
jryb, line 686: This would change to:
```
              description: Full path or URL for one or more default.yml files specific
                to App install, separated by commas.  The defaults listed here
                will only be installed the SHC Deployer and pushed to SHs in a cluster.
                This parameter is ignored on individual SH members.
```

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
```diff
@@ -681,14 +681,28 @@ spec:
...
                 by commas
               type: string
-            ephemeralStorage:
-              description: If true, ephemeral (emptyDir) storage will be used for
-                /opt/splunk/etc and /opt/splunk/var volumes
-              type: boolean
-            etcStorage:
-              description: Storage capacity to request for /opt/splunk/etc persistent
-                volume claims (default=”1Gi”)
+            defaultsUrlApps:
```
jryb, line 684: I think we czan remove this from the indexercluster CRD as it is ignored there.

deploy/crds/enterprise.splunk.com_licensemasters_crd.yaml
```diff
@@ -665,17 +665,27 @@ spec:
...
-                by commas
-              type: string
-            ephemeralStorage:
-              description: If true, ephemeral (emptyDir) storage will be used for
-                /opt/splunk/etc and /opt/splunk/var volumes
-              type: boolean
-            etcStorage:
-              description: Storage capacity to request for /opt/splunk/etc persistent
-                volume claims (default=”1Gi”)
+              description: Full path or URL for one or more default.yml files specific
```
jryb, line 668: Change to:
```
              description: Full path or URL for one or more defaults.yml files specific
                to App install, separated by commas.  The defaults listed here
                will be installed on the LM instance.
```

deploy/crds/enterprise.splunk.com_standalones_crd.yaml
```diff
@@ -675,17 +675,27 @@ spec:
...
-                by commas
-              type: string
-            ephemeralStorage:
-              description: If true, ephemeral (emptyDir) storage will be used for
-                /opt/splunk/etc and /opt/splunk/var volumes
-              type: boolean
-            etcStorage:
-              description: Storage capacity to request for /opt/splunk/etc persistent
-                volume claims (default=”1Gi”)
+              description: Full path or URL for one or more default.yml files specific
```
jryb, line 678: This would change to:
```
              description: Full path or URL for one or more defaults.yml files specific
                to App install, separated by commas.  The defaults listed here 
                will be installed on a standalone.
```

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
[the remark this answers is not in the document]
smohan-splunk: I see what you are saying, there are many booleans already in the file. I was looking for some place where we can clarify what the default is, and I found [Custom Resources docs](https://splunk.github.io/splunk-operator/CustomResources.html) already has it, so I think we are good.

pkg/splunk/enterprise/configuration.go
```diff
@@ -315,6 +308,76 @@ func addSplunkVolumeToTemplate(podTemplateSpec *corev1.PodTemplateSpec, name str
 	}
 }
 
+// addPVCVolumes adds pvc volumes to statefulSet
+func addPVCVolumes(cr splcommon.MetaObject, spec *enterprisev1.CommonSplunkSpec, statefulSet *appsv1.StatefulSet, labels map[string]string, volumeType string) error {
+	// prepare and append persistent volume claims if storage is not ephemeral
```
smohan-splunk, line 313: What are the usecases where we may need to 'append' multiple PersistentVolumeClaims to the same pods? What if a user is attempting to replace one PersistentVolume with another?

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -845,3 +846,71 @@ func TestAreRemoteVolumeKeysChanged(t *testing.T) {
...
+			},
+		},
+	}
+
+	// Define valid common spec
+	spec := &enterprisev1.CommonSplunkSpec{}
+	err := addStorageVolumes(&cr, spec, statefulSet, labels)
+	if err != nil {
+		t.Errorf("Unable to add storage volumes, error: %s", err.Error())
+	}
```
smohan-splunk, line 892: Any tests where we confirm if the (default?) Volumes are indeed mounted? Or can we peek into statefulset spec template to confirm this? Looks like at this point we just go by whether the API is returning success or error, without checking if the intended effect has been taken

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
akondur: As discussed this is auto-generated code and comes as a part of changes to common_types.go. Leaving this as is for now. In the future we can move defaultsUrlApps to CRD specific structures to avoid this.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -845,3 +846,71 @@ func TestAreRemoteVolumeKeysChanged(t *testing.T) {
...
+
```
smohan-splunk, line 893: Some combinations which need testing
 - When storage class is not explicitly specified, statefulsets are setup with defaults for both etc & var
 - etc with explicit Storage class, var is empty(& hence uses default)
 - var with explicit Storage class, etc is empty(& hence uses default)
 - etc & var with explicit but same Storage classes
 - etc & var with explicit but different Storage classes
 - Changing etc & var's storage class after they have been setup
 - Deleting explicit storage class on etc/var after they have been setup

@pdhanoyasplunk Do you intend to cover these as part of automation tests?

docs/StorageClass.md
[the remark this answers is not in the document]
akondur: Will change this.

pkg/splunk/enterprise/configuration_test.go
akondur: I tested most of these scenarios. Dynamic change of storage class name/capacity might not be supported by K8S. The pod might need to be deleted & redeployed for the same.

akondur: Added UT as well to test the same.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -845,3 +852,129 @@ func TestAreRemoteVolumeKeysChanged(t *testing.T) {
...
+	spec = &enterprisev1.CommonSplunkSpec{
+		VarVolumeStorageConfig: enterprisev1.StorageClassSpec{
+			StorageCapacity: "----",
+		},
+	}
+	err = addStorageVolumes(&cr, spec, statefulSet, labels)
+	if err == nil {
+		t.Errorf("Unable to idenitfy incorrect VarVolumeStorageConfig resource quantity")
+	}
+
```
smohan-splunk, line 979: It _may_ be good to have `test` here to confirm the previous values have not been wrongly/mistakenly overwritten when invalid configs are applied.

smohan-splunk (approved):

pkg/splunk/enterprise/configuration_test.go
akondur: For now, this test was only for code coverage purposes of an error leg. The erroneous config test might not help here. Good idea though, will keep it in mind for the future though.

jryb (approved):

PR #220, opened 2021-01-08
CSPL-720: Adding support to configure a custom serviceaccount per CRD

akondur:
- Added support to configure custom service account per CRD. 
- Added util apis for service accounts
- Added UT for code coverage & config testing
- Changed documentation as necessary


pkg/splunk/controller/serviceaccount.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2018-2020 Splunk Inc. All rights reserved.
```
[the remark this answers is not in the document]
akondur: Resolved in a different [PR](https://github.com/splunk/splunk-operator/pull/228)

pkg/splunk/controller/serviceaccount_test.go
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2018-2020 Splunk Inc. All rights reserved.
```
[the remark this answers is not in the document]
akondur: Same as above

smohan-splunk (approved):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -752,6 +752,10 @@ spec:
               description: Name of Scheduler to use for pod placement (defaults to
                 “default-scheduler”)
               type: string
+            serviceAccount:
```
jryb, line 755: Does the `serviceAccount` default to anything if not set?  Should we document that here?

pkg/splunk/controller/serviceaccount.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// ApplyServiceAccount creates or updates a Kubernetes serviceAccount
+func ApplyServiceAccount(client splcommon.ControllerClient, serviceAccount *corev1.ServiceAccount) error {
+	namespacedName := types.NamespacedName{Namespace: serviceAccount.GetNamespace(), Name: serviceAccount.GetName()}
+	var current corev1.ServiceAccount
+
+	err := client.Get(context.TODO(), namespacedName, &current)
+	if err == nil {
+		if !reflect.DeepEqual(serviceAccount, &current) {
```
jryb, line 35: Do we need to throw an error or print a log here if the Apply fails?

pkg/splunk/enterprise/configuration.go
```diff
@@ -441,6 +441,16 @@ func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaOb
 		return statefulSet, err
 	}
 
+	// add serviceaccount if configured
+	if spec.ServiceAccount != "" {
+		namespacedName := types.NamespacedName{Namespace: statefulSet.GetNamespace(), Name: spec.ServiceAccount}
+		_, err := splctrl.GetServiceAccount(client, namespacedName)
+		if err == nil {
+			// serviceAccount exists
```
jryb, line 449: Do we want to bail here if we have an error?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
akondur: If no custom serviceaccount is attached to a pod, a default service account assigned to the cluster is used by the pod. Let me document that.

pkg/splunk/controller/serviceaccount.go
akondur: `If err != nil`  it means that the object was found and hence we don't need to error out/print log?

pkg/splunk/enterprise/configuration.go
akondur: **GetServiceAccount** throws an error as per this [log](https://github.com/splunk/splunk-operator/pull/220/files#diff-b0b10241f20d210433da861923f8978a724ec51c2d73c668b3163951eb3dccedR53).

pkg/splunk/controller/serviceaccount.go
jryb: Ok so the cases here are: The context doesn't exists and we create it.  Or it exists and we check if its the same and if not, call `UpdateResource`.  If there is an error in either its returned to the caller.  Do we want a log stateing we're updating it?

akondur: Got it. Making this change as well now.

jryb (approved):

PR #221, opened 2021-01-08
CSPL-637 update port names to conform with istio convention

vebken-splunk:
When using istio ingress, istio uses a port naming convention "<protocol>[-<suffix>]" to help determine port protocol. This PR updates our port names to conform with this convention.


mgaldino-splunk (approved):

pkg/splunk/enterprise/configuration.go
```diff
@@ -246,27 +246,27 @@ func prepareSplunkSmartstoreConfigMap(identifier, namespace string, crKind strin
 // getSplunkPorts returns a map of ports to use for Splunk instances.
 func getSplunkPorts(instanceType InstanceType) map[string]int {
 	result := map[string]int{
-		"splunkweb": 8000,
-		"splunkd":   8089,
+		"http-splunkweb": 8000,
```
akondur, line 249: Can we add string templates for the names of the ports in names.go? Intention would be to maintain a single place in case of further changes too.

akondur (commented):
Minor comment, but looks good otherwise.

pkg/splunk/enterprise/configuration.go
```diff
@@ -246,27 +246,27 @@ func prepareSplunkSmartstoreConfigMap(identifier, namespace string, crKind strin
...
-		result["datareceive"] = 19000
-		result["dfsmaster"] = 9000
+		result[GetPortName(dfccontrolPort, "tcp")] = 17000
+		result[GetPortName(datareceivePort, "tcp")] = 19000
+		result[GetPortName(dfsmasterPort, "tcp")] = 9000
 	case SplunkIndexer:
-		result["hec"] = 8088
-		result["s2s"] = 9997
+		result[GetPortName(hecPort, "http")] = 8088
+		result[GetPortName(s2sPort, "tcp")] = 9997
```
akondur, line 269: @vebken-splunk The change looks great. 

Nit: Does it make sense to create string constants for the protocol strings as well?

vebken-splunk: Yeah I guess it's probably worth having them for consistency and potential future usage by other modules. Will update that.

smohan-splunk (approved):

akondur (approved):

jryb (approved):

PR #225, opened 2021-01-12
CSPL-578: Adding testcase lm with standalone

jambrosiano:


smohan-splunk (approved):
Nice PR! We should extend this to other (non-standalone) scenarios too in the future

pdhanoyasplunk (approved):

PR #244, opened 2021-02-02
CSPL-694: Avoid pod recycle while scaling for all CRs except SHC

akondur:
Changes to avoid pod recycle while scaling up/down for all CRs except SHC.

- Reverted changes made for [CSPL-534 ](https://github.com/splunk/splunk-operator/pull/193)
- Sorting the container ports, tolerations, volumes, volume mounts and env variables before creation of a statefulSet.
- Avoiding the use of SPLUNK_INDEXER_URL for the indexer CR as it is used only for non-clustering scenarios by standalone, shc CR pod containers
- Fix/add UT.
- **"make generate"** for the develop branch also generated permissions for **serviceaccount** manipulation for the operator. Reason for this being that **"make generate"** was not run in this already merged [PR](https://github.com/splunk/splunk-operator/pull/220) for service account configuration.

Short description:

In every reconcile, the statefulSet spec created by the operator is compared against the one stored in etcd. While comparing the two specs, for the fields represented by slices(ports, volume mounts etc..) the order of the elements is important i.e any change in order followed by an update of statefulSet will cause a change in the UpdatedRevision field in the StatefulSpec. This in turn triggers a pod recycle unnecessarily. To avoid the same, sort the slices during the statefulSet creation.

Note: During the update scenario, MergePodUpdates takes care of sorting.


pkg/splunk/common/util.go
```diff
@@ -424,6 +424,21 @@ func sortAndCompareSlices(a interface{}, b interface{}, keyName string) bool {
...
 
+	// Sort slices
+	SortSlice(a, keyName)
+	SortSlice(b, keyName)
+
+	return !reflect.DeepEqual(a, b)
+}
+
+// SortSlice sorts a slice of any kind by keyName
+func SortSlice(a interface{}, keyName string) {
```
sgontla, line 435: Seems like needs a return type?

pkg/splunk/controller/util.go
```diff
@@ -189,6 +189,32 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name
...
+
+		// Sort VolumeMounts
+		splcommon.SortSlice(current.Containers[idx].VolumeMounts, splcommon.SortFieldName)
+
+		// Sort env variables
+		splcommon.SortSlice(current.Containers[idx].Env, splcommon.SortFieldName)
+	}
+	scopedLog.Info("Successfully sorted slices in statefulSet")
+
+	return nil
```
sgontla, line 215: In the above comment, if SortSlice is returning a value, then needs to be populated here with a corresponding error?

smohan-splunk (approved):

pkg/splunk/common/util.go
akondur: I didn't find a need to have a return type here. Any errors experienced here leads to a panic.

pkg/splunk/controller/util.go
akondur: Same comment as above. Didn't find the need to return a value.

pkg/splunk/common/util.go
sgontla: My bad, I didn't notice that return @ 463 is part of the anonymous function. Mya be better readable :-) ?

sgontla (approved):

pkg/splunk/controller/util.go
```diff
@@ -189,6 +189,32 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name
 	return result
 }
 
+// SortSlicesStatefulSet sorts required slices in a statefulSet
+func SortSlicesStatefulSet(current *corev1.PodSpec, name string) error {
```
gaurav-splunk, line 193: nit: maybe rename it to `SortStatefulSetSlices`?

gaurav-splunk (approved):

PR #246, opened 2021-02-04
CSPL-817 Removes collapsible blocks for Github Pages compatibility

mgaldino-splunk:
We used collapsible blocks to keep the documentation short, but that caused problems with Github pages vs GitHub repo rendering.

As a solution, we are removing the collapsible blocks to ensure the documentation is consistent on all platforms.

No changes in the content were made in this PR.


vebken-splunk (approved):

smohan-splunk (approved):

akondur (approved):
I would do one final check if all the occurrences are covered(If not done already)

PR #254, opened 2021-02-11
CSPL-812: Test Automation to verify LM with IDXC

jambrosiano:
This PR adds a test to verify that an indexer cluster (with licenseMasterRef only on the CM, not on the peers) can be configured with a LM


smohan-splunk (approved):

tpereirasplunk (approved):

akondur (approved):
Looks good

test/delete_cr/deletecr_test.go
```diff
@@ -54,7 +54,8 @@ var _ = Describe("DeleteCR test", func() {
 	Context("Multisite cluster deployment (M13 - Multisite indexer cluster, Search head cluster)", func() {
 		It("deletecr: can deploy indexers and search head cluster", func() {
 
-			err := deployment.DeploySingleSiteCluster(deployment.GetName(), 3)
+			shc := true
```
akondur, line 57: Nit: Instead of using a variable here. You could directly call the function as follows to indicate that the bool is to indicated whether you want a SHC:

`err := deployment.DeploySingleSiteCluster(deployment.GetName(), 3, /*shc*/true)`

test/testenv/deployment.go
```diff
@@ -280,9 +280,12 @@ func (d *Deployment) DeploySingleSiteCluster(name string, indexerReplicas int) e
 		return err
 	}
 
-	_, err = d.DeploySearchHeadCluster(name+"-shc", name, licenseMaster, "")
-	if err != nil {
-		return err
+	// Deploy the SH cluster
+	if shc == true {
```
akondur, line 284: Nit: `if shc` vs `if shc == true` ?

test/smoke/smoke_test.go
```diff
@@ -203,4 +204,38 @@ var _ = Describe("Smoke test", func() {
...
+			licenseFilePath, err := testenv.DownloadFromS3Bucket()
+			Expect(err).To(Succeed(), "Unable to downlaod license file")
+
+			// Create License Config Map
+			testenvInstance.CreateLicenseConfigMap(licenseFilePath)
+
+			// Create Cluster Master with LicenseMasterRef, IndexerCluster without LicenseMasterRef
+			idxCount := 3
+			shc := false
+			err = deployment.DeploySingleSiteCluster(deployment.GetName(), idxCount, shc)
```
akondur, line 220: Same as above.

akondur (approved):
Small comments but looks good

PR #257, opened 2021-02-18
CSPL-547: Test Automation to Verify MC configured with LM

tpereirasplunk:


test/testenv/mcutil.go
```diff
@@ -177,3 +177,33 @@ func CheckStandalonePodOnMC(ns string, podName string) bool {
 	}
 	return found
 }
+
+// CheckMontoringConsoleConfigured checks if mc is configured on given pod
+func CheckMontoringConsoleConfigured(ns string, deployment *Deployment, podName string) bool {
```
jambrosiano, line 182: Nit: typo in 'monitoring'

smohan-splunk (approved):

jambrosiano (approved):

test/testenv/mcutil.go
```diff
@@ -177,3 +177,33 @@ func CheckStandalonePodOnMC(ns string, podName string) bool {
...
+		logf.Log.Error(err, "Failed to execute command", "command", cmd)
+		return false
+	}
+
+	for _, line := range strings.Split(string(output), "\n") {
+		// Check for empty lines to prevent an error in logic below
+		if len(line) == 0 {
+			continue
+		}
+		// Check for specifc logs that indicate that MC was configured with LM
```
akondur, line 198: Isn't there a way to check the MC pod/container for information rather than relying on logs?

test/testenv/verificationutils.go
```diff
@@ -272,3 +272,11 @@ func VerifyIndexExistsOnS3(deployment *Deployment, podName string, indexName str
 		return indexFound
 	}, deployment.GetTimeout(), PollInterval).Should(gomega.Equal(true))
 }
+
+//VerifyMCConfiguredWithLM Verify that MC is configured with License Master
```
akondur, line 276: verifies vs verify?

test/testenv/mcutil.go
tpereirasplunk: Modified

test/testenv/verificationutils.go
tpereirasplunk: Just followed the syntax followed for other functions.

test/testenv/mcutil.go
tpereirasplunk: There are two other options the way i see.....

Option1: 
Check for peers in 'splunk_monitoring_console_assets.conf' and see if License Master is one of them.

Option2:
Check 'apps/splunk_monitoring_console/local/app.conf' and see if  'is_configured = 1' is set. We can do these.

Ideally we can do all 3. Let me know

akondur: I would go with **option 1 + option 2** & avoid the logging check to be safe. Logging with ansible can be changed anytime and configuration checks on the PVC volumes would be a good verification mechanism for the same.

tpereirasplunk: Added Option 1 + 2. And removed the log assertion

PR #258, opened 2021-02-18
CSPL-827: Improve automation to include License Master on several deployments

jambrosiano:
This PR adds a License Master test folder with several tests which verify indexers and SHs are well configured with  LM.
LM tests are added for following deployments:
- Standalone deployment (S1) with LM
- Clustered deployment (C3 - clustered indexer) with LM
- Clustered deployment (C3 - clustered indexer, search head cluster)
- Multisite cluster deployment (M13 - Multisite indexer cluster, Search head cluster)
- Multisite cluster deployment (M1 - multisite indexer cluster)

This PR also removes all tests with LM from smoke, so it can be run on customers environment without need of S3 access.

### DIFF
--- a/test/smoke/smoke_test.go
+++ b/test/smoke/smoke_test.go
@@ -143,38 +143,6 @@ var _ = Describe("Smoke test", func() {
 		})
 	})
 
-	Context("Standalone deployment (S1) with LM", func() {
-		It("smoke: can deploy a standalone instance and a License Master", func() {
-			// Download License File
-			licenseFilePath, err := testenv.DownloadFromS3Bucket()
-			Expect(err).To(Succeed(), "Unable to downlaod license file")
-
-			// Create License Config Map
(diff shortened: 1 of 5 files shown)


tpereirasplunk (approved):

test/smoke/smoke_test.go
```diff
@@ -89,14 +96,37 @@ var _ = Describe("Smoke test", func() {
 
 			// Verify RF SF is met
 			testenv.VerifyRFSFMet(deployment, testenvInstance)
+
```
pdhanoyasplunk, line 99: Considering we are giving customers like HP ability to run these test cases on their local environment we should move LM test into its own suite as they are reliant on s3 access. Please create a new suite for LM Test.

pdhanoyasplunk (approved):

smohan-splunk (approved):

PR #267, opened 2021-03-01
CSPL-841 Upgrade aws-sdk-go version to v1.34.1 or later

vebken-splunk:
A Whitesource scan of the operator branch revealed it is vulnerable to CVE-2020-8911 and CVE-2020-8912

CVE-2020-8911 is fixed by upgrading the aws-sdk-go library to v1.34.1 or later.

CVE-2020-8912 is fixed by upgrading the aws-sdk-go library to v1.28.3 or later.

PR #268, opened 2021-03-01
CSPL-836 Mitigate vulnerabilities from github.com/coreos/prometheus-operator/pkg/client/versioned/scheme-v0.34.0 library

vebken-splunk:
Upgrade prometheus-operator from v0.34.0 to v0.38.3 to address vulnerabilities CVE-2020-8565, CVE-2019-11253, CVE-2019-11254, CVE-2020-14040, and CVE-2020-15113


go.mod
```diff
@@ -3,15 +3,15 @@ module github.com/splunk/splunk-operator
 go 1.13
 
 require (
-	github.com/aws/aws-sdk-go v1.17.7
+	github.com/aws/aws-sdk-go v1.30.12
```
smohan-splunk, line 6: Do we need this line as this is already being covered as part of https://github.com/splunk/splunk-operator/pull/267/files ?

go.mod
```diff
@@ -3,15 +3,15 @@ module github.com/splunk/splunk-operator
...
+	github.com/coreos/prometheus-operator v0.38.3 // indirect
```
smohan-splunk, line 7: Question similar to aws-sdk-go [discussion](https://github.com/splunk/splunk-operator/pull/267#discussion_r585112183) .. does updating to latest version  of prometheus operator cause build failures or other problems?

vebken-splunk: Yes, upgrading to the latest version of prometheus operator causes dependency issues and build issues. v0.38.3 is the latest version that will work without further changes such as upgrading the operator sdk version.

go.mod
vebken-splunk: This version is updated automatically by requiring prometheus-operator v0.38.3 so it is necessary for this standalone fix. When merging this and CSPL-841 we will need to ensure that the latest version of the awk-sdk-go library is retained (depending on the order the tickets are merged in).

smohan-splunk (approved):

akondur (approved):

PR #284, opened 2021-03-19
correct example YAML in README.md

jkoby:
The file mentioned on line 60 doesn't have v1 versions of resources - recommend changing this to v1beta1 until the operator install script is updated.


smohan-splunk:
@jkoby Thanks for the PR. We just made a release (1.0.0-RC) yesterday to fix the CRDs & docs, hence this change will not be necessary anymore.

PR #288, opened 2021-03-20
CSPL-931: Fix failing secret test

pdhanoyasplunk:
Fixed the following issues with the secret test:
* GetSecretObject throwing parsing error on unmarshall. Modify the get secret command to get json string output that can be unmarshalled.
* Multiple Log.Info issues (Odd keys value pairs provided resulting in errors during runtime)
* No assertion in Verification methods ( Test are not failing on key mismatch)
* No Update Phase check after new secret are applied resulting in failures due to incorrect phase.
* Incorrect Secret response struct.
* Incorrect parameter names defined for method **VerifySecretObjectUpdate**
* Removed redundant phase check method ( Not related to secret test)
* Added support for verification for **password, idxc_support, pass4symmkey, shc_secret**
* Added proper checks for Splunk Environment to go to Ready Phase

### DIFF
--- a/test/testenv/secretutil.go
+++ b/test/testenv/secretutil.go
@@ -15,22 +15,35 @@
 package testenv
 
 import (
+	// "bytes"
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os/exec"
-	"strings"
+
+	// "strings"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 //SecretResponse Secret object struct
 type SecretResponse struct {
-	HecToken     string `json:"hec_token"`
-	IdxcSecret   string `json:"idxc_secret"`
-	Pass4SymmKey string `json:"pass4SymmKey"`
-	Password     string `json:"password"`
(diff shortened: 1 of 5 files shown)


test/testenv/secretutil.go
```diff
@@ -50,52 +63,52 @@ func EncodeBase64(str string) string {
...
-		value := DecodeBase64(restResponse.ShcSecret)
-		return value
-	default:
-		return "Invalid Key"
+	if restResponse == (SecretResponse{}) {
+		return "Not Found"
+	}
+	logf.Log.Info("Get secret object encoded value", "Secret Name", secretName, "Key", key)
+	value := "Invalid Key"
+	if key == "hec_token" {
```
akondur, line 91: Nit: Use else if or switch? Can take it up later as well.

akondur (approved):

tpereirasplunk (approved):

test/testenv/secretutil.go
pdhanoyasplunk: Will do.

smohan-splunk (approved):

PR #309, opened 2021-04-06
CSPL-979: Updates regarding current state of namespace and review on docs per markdown compatibility

kumarajeet:


eddikay (commented):
This is a dilemma we see in docs all the time. There's a bug. It's in a release note. If people are looking for issues, they can find it. However, sometimes the broken functionality is important enough that a release note alone is deemed insufficient (usually based on case count and severity, real or potential.) Then there's a desire to update the docs directly, in addition to having a RN. 

If you do this, it should be stated in the note what is broken, and it should state clearly if the customers should avoid using the functionality. If there is a workaround, the workaround should be stated clearly in the note. Then you can add a "the bug is tracked here" link.

I also suggest that any doc references be added to the bug ticket as a post-fix task so the references can be reviewed and updated with the new info (or just the note removed, as needed.)

docs/CustomResources.md
```diff
@@ -76,7 +76,8 @@ configuration parameters:
...
-| resources             | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#resourcerequirements-v1-core) | The settings for allocating [compute resource requirements](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/) to use for each pod instance. |
+| resources             | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#resourcerequirements-v1-core) | The settings for allocating [compute resource requirements](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/) to use for each pod instance. The default settings  should be considered for demo/test purposes.  Please see [Hardware Resource Requirements](https://github.com/splunk/splunk-operator/blob/develop/docs/README.md#hardware-resources-requirements) for production values.
```
akondur, line 79: Extra space between the words **settings** and **should**

akondur (approved):

docs/README.md
```diff
@@ -41,17 +41,12 @@ Review the [Change Log](ChangeLog.md) page for a history of changes in each rele
...
-|  |   |
-
-
-
-
+<table>
+<tr><td> Splunk Development & Testing Platforms </td><td> Amazon Elastic Kubernetes Service (EKS), Google Kubernetes Engine (GKE) </td></tr>
+<tr><td> Customer Reported Platforms </td><td> Microsoft Azure Kubernetes Service (AKS), Red Hat OpenShift </td></tr>
+<tr><td> Partner Tested Platforms</td><td> HPE Ezmeral</td></tr>
+<tr><td> Other Platforms </td><td>CNCF certified distribution</td></tr>
```
smohan-splunk, line 48: The hyperlink needs to be added back?

kumarajeet: I removed it as we already have CNCF with link in the section above.

eddikay (approved):

smohan-splunk (approved):

PR #325, opened 2021-04-30
CSPL-633 Support extraENV in CR spec

jryb:
### Problem
Splunk operator developers need a method to add additional env variables to the Splunk instance pods for debugging and Splunk app installation.  The operator needs to provide them with a method of defining these environment variables in the CRD so that they are propagated to all created Splunk instances when deployed.

### Solution
Allow users to specify additional env variables to be passed down to the Splunk instance
pods via the CRD with the following config:
```
  kind: Standalone
  metadata:
    name: sa
  spec:
    extraEnv:
    - name: SOME_COOL_ENV_VAR
      value: "stuff"
  - name: ANSIBLE_EXTRA_FLAGS
    value: "-vvv"
```
These environment variables will be passed and set in the env of each Splunk instances
created by applying this CRD.

Added additional unit tests to validate that the `extraEnv` parameters are passed to the
`env` of the created Splunk instance containers.

### Tests

Along with adding unit tests, this functionality can be verified functionally as well.  Validate that when adding `ANSIBLE_EXTRA_FLAGS` to the `extraEnv:` section, Ansible prints out verbose debugs:
```
apiVersion: enterprise.splunk.com/v1beta1
kind: Standalone
metadata:
  name: sa
  finalizers:
spec:
  replicas: 1
  extraEnv:
  - name: SOME_COOL_ENV_VAR
    value: "stuff"
  - name: ANSIBLE_EXTRA_FLAGS
    value: "-vvv"
```
Check kubectl describe:
```
$ kdes pod/splunk-sa-standalone-0
Name:         splunk-sa-standalone-0
Namespace:    default
...
Controlled By:  StatefulSet/splunk-sa-standalone
Containers:
  splunk:
...
    Environment:
      ANSIBLE_EXTRA_FLAGS:                -vvv
      SOME_COOL_ENV_VAR:                  stuff

```
Validate with ansible logs:
```
<snip>
included: /opt/ansible/roles/splunk_common/tasks/licenses/add_license.yml for localhost
Friday 30 April 2021  16:37:10 +0000 (0:00:00.114)       0:00:51.036 ********** 
Friday 30 April 2021  16:37:10 +0000 (0:00:00.108)       0:00:51.145 ********** 
Using module file /usr/lib/python3.7/site-packages/ansible/modules/stat.py
Pipelining is enabled.
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: splunk
<localhost> EXEC /bin/sh -c '/usr/bin/python && sleep 0'

TASK [splunk_common : Ensure license path] *************************************
task path: /opt/ansible/roles/splunk_common/tasks/licenses/add_license.yml:15
ok: [localhost] => {
    "changed": false,
    "invocation": {
        "module_args": {
            "checksum_algorithm": "sha1",
            "follow": false,
            "get_attributes": true,
            "get_checksum": true,
            "get_md5": false,
            "get_mime": true,
            "path": "splunk.lic"
        }
    },
    "stat": {
        "exists": false
    }
}
Friday 30 April 2021  16:37:10 +0000 (0:00:00.752)       0:00:51.897 ********** 
Friday 30 April 2021  16:37:10 +0000 (0:00:00.083)       0:00:51.981 ********** 
META: ran handlers
<snip>
```


pkg/splunk/enterprise/indexercluster.go
```diff
@@ -589,7 +589,7 @@ func getIndexerStatefulSet(client splcommon.ControllerClient, cr *enterprisev1.I
 	// 1. Introduce the new env variables in the function getIndexerExtraEnv
 	// 2. Avoid SPLUNK_INDEXER_URL in getIndexerExtraEnv for idxc CR
 	// 3. Re-introduce the call to getIndexerExtraEnv here.
-	return getSplunkStatefulSet(client, cr, &cr.Spec.CommonSplunkSpec, SplunkIndexer, cr.Spec.Replicas, make([]corev1.EnvVar, 0))
+	return getSplunkStatefulSet(client, cr, &cr.Spec.CommonSplunkSpec, SplunkIndexer, cr.Spec.Replicas, cr.Spec.CommonSplunkSpec.ExtraEnv)
```
akondur, line 592: Is the default value of `cr.Spec.CommonSplunkSpec.ExtraEnv` an slice of type `corev1.EnvVar` and size 0?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -164,6 +164,9 @@ func validateClusterMasterSpec(cr *enterprisev1.ClusterMaster) error {
 // getClusterMasterStatefulSet returns a Kubernetes StatefulSet object for a Splunk Enterprise license master.
 func getClusterMasterStatefulSet(client splcommon.ControllerClient, cr *enterprisev1.ClusterMaster) (*appsv1.StatefulSet, error) {
 	var extraEnvVar []corev1.EnvVar
+	for _, envVar := range cr.Spec.CommonSplunkSpec.ExtraEnv {
```
akondur, line 167: Since the extraEnv parameter is a part of `CommonSplunkSpec`(which is available to use in the function `getSplunkStatefulSet`), could we add the logic to adding to the env variables [here](https://github.com/splunk/splunk-operator/blob/develop/pkg/splunk/enterprise/configuration.go#L652) in the following manner?

```
for _, envVar := range spec.CommonSplunkSpec.ExtraEnv {
    	extraEnvVar = append(env, envVar)
}
```

This would involve making a change in a single place in the repo vs CR specific changes.

Just a suggestion, the changes look good otherwise.

jryb: Ya good point.  I had put the code in the `extraEnv` param for `getSplunkStatefulSet()` since it already had the extraEnv variable that was NULL'ed out for most CRDs.  Though since its common it makes a lot more sense to put it in the common area for all types.  I'll adjust the diff to reflect this.

pkg/splunk/enterprise/indexercluster.go
jryb: Yes.

vebken-splunk:
I believe you would need to run "make generate" so that the CRDs themselves are updated with extraEnv in this PR as well, right?

pkg/splunk/enterprise/clustermaster.go
akondur: Looks great. Thanks Jeff

akondur (approved):

jryb:
> I believe you would need to run "make generate" so that the CRDs themselves are updated with extraEnv in this PR as well, right?

Yeah I need to run `make generate`.  I did however it created a lot of white space/spacing diffs.  So I'm trying a few things to make sure the CRD changes that are generated are isolated to these changes only.

jryb:
Squashed commits

jryb:
> > I believe you would need to run "make generate" so that the CRDs themselves are updated with extraEnv in this PR as well, right?
> 
> Yeah I need to run `make generate`. I did however it created a lot of white space/spacing diffs. So I'm trying a few things to make sure the CRD changes that are generated are isolated to these changes only.

Fixed by using yq v3.2.1

pkg/splunk/enterprise/configuration.go
```diff
@@ -708,6 +708,11 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		})
 	}
 
+	// Add extraEnv from the CommonSplunkSpec config to the extraEnv variable list
+	for _, envVar := range spec.ExtraEnv {
+		extraEnv = append(extraEnv, envVar)
+	}
+
```
sgontla, line 715: Looking at the CRD changes seems like `[]corev1.EnvVar`, is supporting multiple ways of specifying the env variables to source from(configMap, secret object, etc..).  Wondering if we need to deal with them the same way, like reading those resource objects, and populating them as part of the extraEnv here?

jryb: Since the extraEnv is the same `[]corev1.EnvVar` object, it will behave the same as all the other `EnvVar` objects.  Here we are just appending the `ExtraEnv` section to the env vars already created.  If `EnvVarSource` is used and the source doesn't exist, this should throw an error when creating the pod.
Applied CR:
```
apiVersion: enterprise.splunk.com/v1beta1
kind: Standalone
metadata:
  name: sa
  finalizers:
spec:
  replicas: 1
  extraEnv:
  - name: ANSIBLE_EXTRA_FLAGS
    value: "-vvv"
  - name: FROM_SOURCE
    valueFrom:
      secretKeyRef:
        name: doesntexistsecret
        key: awsSecretAccessKey
```
Error:
```
splunk-sa-standalone-0             0/1     CreateContainerConfigError   0          101s   10.244.0.22   kind-control-plane   <none>           <none>
...
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  13s               default-scheduler  Successfully assigned default/splunk-sa-standalone-0 to kind-control-plane
  Normal   Pulled     1s (x3 over 13s)  kubelet            Container image "docker.io/splunk/splunk:8.1.3" already present on machine
  Warning  Failed     1s (x3 over 13s)  kubelet            Error: secret "doesntexistsecret" not found
```
I'm not sure we really need to this prior to apply since its already handled by kubectl.

sgontla: Thanks Jeff. 
Doing it prior may help in the case of incremental config change, where we destroy an existing Pod, before starting a pod with the missing resource, which may be a rare case :-)

sgontla (approved):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -943,7 +1047,6 @@ spec:
                               x-kubernetes-int-or-string: true
                           required:
                           - port
-                          - protocol
```
smohan-splunk, line 946: Is this removed because of the newer SDK?

smohan-splunk (approved):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
jryb: Yes, I assume so.  I didn't touch this config, just ran `make generate` and it threw out this diff.

jryb:
Manually readded https://github.com/splunk/splunk-operator/pull/331/commits back, updated to latest and squashed.

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
vebken-splunk: Having to re-add the protocol field is a limitation of the workaround to https://github.com/kubernetes/kubernetes/issues/91395 at the moment that we have to do anytime we generate CRDs.
 I have been looking into a better solution to automate this process or workaround the issue altogether, I can open a CSPL to formally track this.

Other projects that have hit the issue (kubernetes-sigs/controller-tools#444) mention upgrading their controller-gen version works around the issue, but I don't believe we use controller-gen to generate the crds. There are some other ideas mentioned here kubernetes/kubernetes#91395, but the automated ones seem to require modifying k8s api code.

vebken-splunk (approved):

pdhanoyasplunk (approved):

PR #329, opened 2021-05-03
CSPL1023: Add configuration support(CRD Spec changes) for App framework

sgontla:


sgontla:
Sample config:
kind: Standalone
metadata:
  name: s1
  finalizers:
spec:
  replicas: 1
  appRepo:
    defaults:
      volumeName: volume_repo_2
      scope: local
    appSources:
      - name: ciscoApps
        location: ciscoAppsRepo
        volumeName: volume_repo_2
      - name: juniperApps
        location: juniperAppsRepo
        volumeName: volume_repo_2
      - name: genericNetworkApps
        location: genericNetworkAppsRepo
      - name: devApps
        location: devAppsRepo
        scope: local
    volumes:
      - name: volume_repo_2
        path: sgontla-s2-bucket/bucketCorruption
        secretRef: s3-secret-1

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -178,36 +188,49 @@ type IndexAndCacheManagerCommonSpec struct {
...
+	// List of App sources on remote Volumes
+	AppSources []AppSourceSpec `json:"appSources,omitempty"`
+}
+
+// AppDeploymentContext for storing the App deployment state
+type AppDeploymentContext struct {
+	// App Framework version info for future use
+	Version uint16 `json:"version"`
+
+	// false: Remote Storage can be probed.
```
smohan-splunk, line 231: Comment needs update?

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +783,125 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
 
 	return nil
 }
 
-// ValidateSplunkSmartstoreSpec checks and validates the smartstore config
-func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error {
-	var err error
+//  isAppFrameworkConfigured checks and returns true if App Framework is configured
+//  App Repo config without any App sources will not cause any App Framework activity
+func isAppFrameworkConfigured(appFramework *enterprisev1.AppFrameworkSpec) bool {
```
jryb, line 851: This is basically our feature flag, correct?  If both `appRepo` and `appSources` are not configured, App framework is effectively disabled?

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +783,125 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+
+	// Todo: sgontla: Enable later. For now, allowing multiple volumes and app sources for UT purpose
+	if len(appFramework.VolList) > 1 || len(appFramework.AppSources) > 2 {
+		//return fmt.Errorf("Invalid App Framework config. Only one Volume, and a max. of two App sources supported")
+	}
+
+	scopedLog.Info("Validating app framework spec")
+
+	//ToDo: sgontla: Do we need to return error, if the poll interval is too small? TBD
+	if appFramework.AppsRepoPollInterval < minAppsRepoPollInterval {
```
jryb, line 873: Should we specify the units of the polling interval in the CRD and the internal variable names?

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +783,125 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+		_, err = checkIfVolumeExists(appFramework.VolList, appFramework.Defaults.VolName)
+		if err != nil {
+			return fmt.Errorf("Invalid Volume Name for Defaults. Error: %s", err)
+		}
 	}
 
+	return err
+}
+
+func validateRemoteVolumeSpec(volList []enterprisev1.VolumeSpec) error {
```
jryb, line 901: `func ValidateAppFrameworkSpec()` just validates the config vs. the spec, right?  Not the validity of the configured S3 parameters (i.e. if the S3 endpoint or bucket actually exist, or the creds are correct, etc)

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -71,6 +71,16 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
+	// 1. Spec update
+	// 2. S3 credentails change
+	// 3. Probe times expired
+	if !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
+		// TBD: Probe the remote store, and handle any changes on remote store
+
+		cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
+	}
+
 	// This is to take care of case where AreRemoteVolumeKeysChanged returns an error if it returns false.
```
smohan-splunk, line 84: Err check block related to AreRemoteVolumeKeysChanged should preferably be moved prior to the AppFrameworkConfig spec check

pkg/splunk/enterprise/configuration.go
```diff
@@ -33,6 +32,12 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+//ToDo sgontla: Move it to a common place
+const (
+	minAppsRepoPollInterval uint = 60           // one minute
+	maxAppsRepoPollInterval uint = 60 * 60 * 24 // one day
```
smohan-splunk, line 38: Can we add a default (1 hour)?

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +783,125 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+		"localScope: %t", localScope)
+
+	// Todo: sgontla: Enable later. For now, allowing multiple volumes and app sources for UT purpose
+	if len(appFramework.VolList) > 1 || len(appFramework.AppSources) > 2 {
+		//return fmt.Errorf("Invalid App Framework config. Only one Volume, and a max. of two App sources supported")
+	}
+
+	scopedLog.Info("Validating app framework spec")
+
+	//ToDo: sgontla: Do we need to return error, if the poll interval is too small? TBD
```
smohan-splunk, line 872: Yes, rather than silently defaulting to min/max, we should instead throw error, so that Admin knows they need to change the spec

pkg/splunk/enterprise/configuration.go
smohan-splunk, line 873: Is AppsRepoPollInterval mandatory? If not, by not providing the spec param, can we default (to say 1 hour)?

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +783,125 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+		}
+		duplicateChecker[vol+appSrc.Location] = true
 
-	if appFramework.S3Endpoint == "" {
-		return fmt.Errorf("Apps Remote Storage S3Endpoint is missing")
 	}
 
-	if appFramework.S3Bucket == "" {
-		return fmt.Errorf("Apps Remote Storage S3Bucket is missing")
+	if localScope && appFramework.Defaults.Scope != "" && appFramework.Defaults.Scope != "local" {
```
smohan-splunk, line 831: What happens if Defaults.Scope = "" and scope has not been explicitly specified? Should we default to local/cluster based on the CR?

pkg/apis/enterprise/v1/common_types.go
sgontla: Addressed

pkg/splunk/enterprise/clustermaster.go
sgontla: Addressed

pkg/splunk/enterprise/configuration.go
sgontla: addressed

pkg/splunk/enterprise/configuration.go
sgontla: Good catch. adding extra check for returning error in that case.

sgontla: Also, added a UT to catch the same.

pkg/splunk/enterprise/configuration.go
sgontla: Correct. To be more clear, the config is not effective, unless the App sources are specified.

pkg/splunk/enterprise/configuration.go
sgontla: OK, addressing it.

pkg/splunk/enterprise/configuration.go
sgontla: Addressed

pkg/splunk/enterprise/configuration.go
sgontla: That is correct. We are doing that when we are actually using it, which is kind of bad idea, but, just fitting the existing flow, as we need the cr and the controller client to pull any K8 resources, which we are not doing in this flow. Example:-

`func validateLicenseMasterSpec(spec *enterprisev1.LicenseMasterSpec) error {

	err := ValidateAppFrameworkSpec(&spec.AppFrameworkConfig, true)
	if err != nil {
		return err
	}

	return validateCommonSplunkSpec(&spec.CommonSplunkSpec)`

	All we have is application spec. Its not so difficult to pass on the K8 info either :-)

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -614,39 +614,73 @@ spec:
...
-                type:
-                  description: App Package Remote Store type. The currently supported
-                    type is s3 only.
-                  type: string
+                defaults:
+                  description: Defines default configuration settings for App sources
+                  properties:
+                    scope:
+                      description: 'Scope of the app deployment: cluster, local. Scope
+                        determines the App deployment path on the splunk enterprise
```
kashok-splunk, line 654: should we explain what cluster/local means?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -614,39 +614,73 @@ spec:
...
+                      type: string
+                    volumeName:
+                      description: Remote Storage Volume name
+                      type: string
+                  type: object
+                volumes:
+                  description: List of remote storage volumes
+                  items:
+                    description: VolumeSpec defines remote volume name and remote
+                      volume URI
```
kashok-splunk, line 665: Nit: URI/URL

pkg/apis/enterprise/v1/clustermaster_types.go
```diff
@@ -56,6 +54,9 @@ type ClusterMasterStatus struct {
 
 	// Resource Revision tracker
 	ResourceRevMap map[string]string `json:"resourceRevMap"`
+
+	// App Framework Context
```
kashok-splunk, line 58: Nit: Can we make this comment more descriptive

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -595,134 +595,204 @@ func TestValidateSplunkSmartstoreSpec(t *testing.T) {
...
+		t.Errorf("Unsupported app scope should be cause error, but failed to detect")
 	}
 
-	err = ValidateAppFrameworkSpec(&AppFrameWorkAllParamsValid)
-	if err != nil {
-		t.Errorf("Expected no errors as all good params provided but got error: %s", err)
+	// If the CR supports only local apps, and default is configured with "cluster" scope, that should be detected
+	AppFramework.AppSources[0].Scope = "local"
+	AppFramework.AppSources[1].Scope = "local"
+	AppFramework.AppSources[2].Scope = "local"
```
kashok-splunk, line 769: Super nit: Future we can have one more negative test case where anyone AppSources can have "cluster" scope defined and everything else is set to "local", for a CR that supports only local apps

pkg/splunk/enterprise/configuration.go
```diff
@@ -873,44 +931,72 @@ func ValidateSplunkSmartstoreSpec(smartstore *enterprisev1.SmartStoreSpec) error
...
-	if defaults.VolName != "" {
-		_, err = checkIfVolumeExists(volList, defaults.VolName)
-		if err != nil {
-			return fmt.Errorf("Invalid configuration for defaults volume. %s", err)
-		}
-	}
+	return nil
+}
+
+func validateSplunkIndexesSpec(smartstore *enterprisev1.SmartStoreSpec) error {
```
kashok-splunk, line 937: add a comment for the function definition

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +784,128 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+		_, err = checkIfVolumeExists(appFramework.VolList, appFramework.Defaults.VolName)
+		if err != nil {
+			return fmt.Errorf("Invalid Volume Name for Defaults. Error: %s", err)
+		}
+	}
+
+	return err
+}
+
+func validateRemoteVolumeSpec(volList []enterprisev1.VolumeSpec) error {
```
kashok-splunk, line 905: add a comment for the function definition

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +784,128 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
 	return false
 }
 
-// ValidateAppFrameworkSpec checks and validates the Apps Frame Work config
-func ValidateAppFrameworkSpec(appFramework *enterprisev1.AppFrameworkSpec) error {
+func validateSplunkAppSources(appFramework *enterprisev1.AppFrameworkSpec, localScope bool) error {
```
kashok-splunk, line 787: add a comment for the function definition

jryb (approved):

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -178,36 +188,49 @@ type IndexAndCacheManagerCommonSpec struct {
 	HotlistBloomFilterRecencyHours uint `json:"hotlistBloomFilterRecencyHours,omitempty"`
 }
 
+// AppSourceDefaultSpec defines config common for defaults and App Sources
+type AppSourceDefaultSpec struct {
+	// Remote Storage Volume name
+	VolName string `json:"volumeName,omitempty"`
+
+	// Scope of the app deployment: cluster, local. Scope determines the App deployment path on the splunk enterprise pod. The path is dependent upon the Kind defined in the CRD
```
smohan-splunk, line 196: // Scope of the App deployment: cluster, local. Scope determines whether the App(s) are installed locally or cluster-wide

smohan-splunk (approved):

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +785,127 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+	return !(appFramework == nil || appFramework.AppSources == nil)
+}
 
-	// Smartstore is an optional config (at least) for now
-	if !isSmartstoreConfigured(smartstore) {
+// ValidateAppFrameworkSpec checks and validates the Apps Frame Work config
+func ValidateAppFrameworkSpec(appFramework *enterprisev1.AppFrameworkSpec, localScope bool) error {
+	var err error
+	if !isAppFrameworkConfigured(appFramework) {
 		return nil
```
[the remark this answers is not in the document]
sgontla: ya, we are not logging errors here. This is to avoid unnecessary logging on every reconcile if the customer is not using this feature.

pkg/splunk/enterprise/configuration.go
```diff
@@ -778,76 +785,127 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+
+	// Todo: sgontla: Enable later. For now, allowing multiple volumes and app sources for UT purpose
+	if len(appFramework.VolList) > 1 || len(appFramework.AppSources) > 2 {
+		//return fmt.Errorf("Invalid App Framework config. Only one Volume, and a max. of two App sources supported")
 	}
 
+	scopedLog.Info("Validating app framework spec")
+
+	// Todo: sgontla: Best place to store the final value as part of the status, so that the config change detection will be easy
+	if appFramework.AppsRepoPollInterval == 0 {
```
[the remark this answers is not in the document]
sgontla: For composite configs like app sources, and volumes, we have separate functions. For all, standalone config(even, if something added in the future), we can have it inline here, in at one place.

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -128,7 +130,7 @@ type IndexConfDefaultsSpec struct {
 	IndexAndGlobalCommonSpec `json:",inline"`
 }
 
-// VolumeSpec defines remote volume name and remote volume URI
+// VolumeSpec defines remote volume config
```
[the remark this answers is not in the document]
sgontla: Good point. Can you add this in your changes. Also, for the S3, if we need aws/miniio kind of info, based on your first-hand info on what exactly needed here?

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -141,6 +143,14 @@ type VolumeSpec struct {
 
 	// Secret object name
 	SecretRef string `json:"secretRef"`
+
+	// Remote Storage type.
+	Type string `json:"storageType"`
```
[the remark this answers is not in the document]
sgontla: I know we discussed this last time. For now, I did not add any hard checks on this. But, please add any as you get clarity on this.

Also, I guess this is something we need to think about how we handle the init/sidecar containers as well, particularly, if it turns out that there are separate utilities to be called based on the implementation of S3.

gaurav-splunk (approved):

PR #334, opened 2021-05-14
CSPL-765 Changes relates to framework for s3 client APIs

gaurav-splunk:
1. This PR targets framework, which will be used by operator to connect to remote storage(typically s3) to get the list of apps as specified in the CRs and which can be used later on to deploy and install the apps from remote storage.

2. This PR is the first part of the changes. Second part/PR will target UTs and code coverage for the related code.


pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +343,47 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+		}
+
+		//get the corresponding S3 client
+		s3Client, err := GetRemoteStorageClient(client, cr, appFrameworkRef, &vol, appSource.Location)
+		if err != nil {
+			return nil, err
+		}
+
+		// Now, get the apps list from remote storage
+		s3Response, err = s3Client.GetAppsList()
```
jryb, line 380: What is the format of the s3Response?

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,86 @@
...
+	svc := s3.New(sess)
+
+	options := &s3.ListObjectsV2Input{
+		Bucket:     aws.String(awsclient.BucketName),
+		Prefix:     aws.String(awsclient.Prefix),
+		StartAfter: aws.String(awsclient.StartAfter), //exclude the directory itself from listing
+	}
+
+	// List the bucket contents
+	resp, err := svc.ListObjectsV2(options)
```
jryb, line 66: Is this a recursive listing of everything from the `path` defined in the `appSource` or only the `path` level?

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,86 @@
...
+		Credentials: credentials.NewStaticCredentials(
+			awsclient.AWSAccessKeyID,     // id
+			awsclient.AWSSecretAccessKey, // secret
+			"")},                         //token
+	)
+
+	// Create S3 service client
+	svc := s3.New(sess)
+
+	options := &s3.ListObjectsV2Input{
```
jryb, line 59: Do we need to add filtering options to only grab the `.tgz` and `.spl` files here or is that filtering handled in the operator code?

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -146,6 +146,10 @@ type VolumeSpec struct {
 
 	// Remote Storage type.
 	Type string `json:"storageType"`
+
+	// App Package Remote Store provider.
+	// For e.g. AWS, Azure, Minio, etc.
```
jryb, line 151: Should we list the currently supported types here as well?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -76,23 +76,36 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
-
 	defer func() {
 		err = client.Status().Update(context.TODO(), cr)
 		if err != nil {
 			scopedLog.Error(err, "Status update failed")
 		}
 	}()
 
+	//check if the apps need to be downloaded from remote storage
+	if cr.Spec.CommonSplunkSpec.Mock != true && !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
```
jryb, line 87: This is only executed when a CRD is applied and the AppFrameworkConfig changes, right?

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+//RegisterAWSS3Client will add the corresponding function pointer to the map
+func RegisterAWSS3Client() {
+	S3Clients["aws"] = GetAWSS3Client
+}
+
+// GetAppsList get the list of apps from remote storage
+func (awsclient *AWSS3Client) GetAppsList() (*S3Response, error) {
+	scopedLog := log.WithName("GetRemoteStorageClient")
```
akondur, line 44: Should this be `scopedLog := log.WithName("GetAppsList")`?

pkg/splunk/enterprise/util.go
```diff
@@ -32,6 +34,55 @@ import (
 // kubernetes logger used by splunk.enterprise package
 var log = logf.Log.WithName("splunk.enterprise")
 
+// regex to extract the region from the s3 endpoint
+var regionRegex = ".*.s3[-,.](?P<region>.*).amazonaws.com"
+
+//getRegion extracts the region from the endpoint field
+func getRegion(endpoint string) string {
```
akondur, line 41: Can we use the `endpoint` variable directly?

```
func getRegion(endpoint string) string {
	pattern := regexp.MustCompile(regionRegex)
	return pattern.FindStringSubmatch(endpoint)[1]
}
```

pkg/splunk/enterprise/util.go
```diff
@@ -32,6 +34,55 @@ import (
...
+// GetRemoteStorageClient returns the corresponding S3Client
+func GetRemoteStorageClient(client splcommon.ControllerClient, cr splcommon.MetaObject, appFrameworkRef *enterprisev1.AppFrameworkSpec, vol *enterprisev1.VolumeSpec, location string) (splclient.S3Client, error) {
+
+	scopedLog := log.WithName("GetRemoteStorageClient").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	//use the provider name to get the corresponding function pointer
+	getClient, _ := splclient.S3Clients[vol.Provider]
+
+	appSecretRef := vol.SecretRef
+	namespaceScopedSecret, err := splutil.GetSecretByName(client, cr, appSecretRef)
```
akondur, line 56: Nit: Is the secret per CR or per S3 client? If it is per S3 client can we call it S3ClientSecret?

pkg/splunk/enterprise/util.go
```diff
@@ -32,6 +34,55 @@ import (
...
+	return pattern.FindStringSubmatch(appS3Endpoint)[1]
+}
+
+// GetRemoteStorageClient returns the corresponding S3Client
+func GetRemoteStorageClient(client splcommon.ControllerClient, cr splcommon.MetaObject, appFrameworkRef *enterprisev1.AppFrameworkSpec, vol *enterprisev1.VolumeSpec, location string) (splclient.S3Client, error) {
+
+	scopedLog := log.WithName("GetRemoteStorageClient").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	//use the provider name to get the corresponding function pointer
+	getClient, _ := splclient.S3Clients[vol.Provider]
```
akondur, line 53: Do we need to error out here if the entry for `vol.Provider` does not exist?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -76,23 +76,36 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
+		var sourceToAppsList map[string]*splclient.S3Response
+
+		for _, vol := range cr.Spec.AppFrameworkConfig.VolList {
```
akondur, line 90: For ClusterMaster are we relying on other CR's(Indexer Cluster) to register S3 clients?

pkg/splunk/client/s3client.go
```diff
@@ -0,0 +1,40 @@
...
+	Key          *string
+	LastModified *time.Time
+	Size         *int64
+	StorageClass *string
+}
+
+//RegisterS3Client registers the respective Client
+func RegisterS3Client(provider string) {
+	switch provider {
+	case "aws":
```
smohan-splunk, line 36: Should we worry about case sensitivity here, or just make it clearer in common_types.go, as Jeff mentioned

pkg/splunk/client/awss3client.go
smohan-splunk: Brings up a bigger question - do we want to support apps in sub-buckets?

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +343,47 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+		if appSource.AppSourceDefaultSpec.VolName != "" {
+			vol = volMap[appSource.AppSourceDefaultSpec.VolName]
+		} else {
+			vol = volMap[appFrameworkRef.Defaults.VolName]
+		}
+
+		//get the corresponding S3 client
+		s3Client, err := GetRemoteStorageClient(client, cr, appFrameworkRef, &vol, appSource.Location)
+		if err != nil {
+			return nil, err
```
smohan-splunk, line 376: Needs further thought when we actually start supporting multiple sources per CR - do we really want to quit right after failure of one source?

docs/MultisiteExamples.md
```diff
@@ -115,7 +115,9 @@ EOF
 ```
 Create IndexerCluster CR for each required site with zone affinity specified as needed
 
-Note: The value of label for zone i.e. `zone-1a` for label `failure-domain.beta.kubernetes.io/zone` is specific to each cloud provider and should be changed based on the cloud provider you are using.
+Note:
+* The value of label for zone i.e. `zone-1a` for label `failure-domain.beta.kubernetes.io/zone` is specific to each cloud provider and should be changed based on the cloud provider you are using
```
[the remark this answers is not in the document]
gaurav-splunk: Yeah need to fix this, I accidentally rebased develop branch on top of this branch. Need to take those commits out of this branch.

pkg/apis/enterprise/v1/common_types.go
gaurav-splunk: Yup sure, will do it.

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,86 @@
...
+
+// GetAWSS3Client returns an AWS S3 client
+func GetAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
+	return &AWSS3Client{
+		Region:             region,
+		BucketName:         bucketName,
+		AWSAccessKeyID:     accessKeyID,
+		AWSSecretAccessKey: secretAccessKey,
+		Prefix:             prefix,
+		StartAfter:         startAfter,
```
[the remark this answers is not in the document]
gaurav-splunk: I say let's start with the default values. Even 1000 is too much. Never seen a customer case with 1000s of apps. If the need arises we can always change it to a higher value.

pkg/splunk/client/awss3client.go
gaurav-splunk: lol yeah, good catch :)

pkg/splunk/client/awss3client.go
gaurav-splunk: I am not sure if there is any such filtering option in the `ListObjectsV2Input`, this probably should be handled in the operator code itself. @sgontla do you want me to handle it or are you handling it already?

pkg/splunk/client/awss3client.go
gaurav-splunk: This will be a recursive listing of everything under the `location` defined in each `appSource`.

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: yup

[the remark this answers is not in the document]
gaurav-splunk: Unfortunately as of now, yes or else the UTs will fail. I will try to create a mock client to mimic the S3 request/response and hopefully, we won't need this field then.

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: Nope, this is just for CM. IndexerCluster CR does not have this in its spec anyways.

pkg/splunk/enterprise/util.go
gaurav-splunk: Yes, we can! I think I left some code as part of the code clean-up. Thanks for catching it :)

pkg/splunk/enterprise/util.go
gaurav-splunk: I believe we are already checking that as part of the validation of cr and if this field is empty we won't even come this far.

pkg/splunk/enterprise/util.go
gaurav-splunk: Its this - 
```
// S3Response struct contains list of RemoteObject objects as part of S3 responsetype 
S3Response struct {	
Objects []*RemoteObject}

// RemoteObject struct contains contents returned as part of S3 response
type RemoteObject struct {	
Etag         *string	
Key          *string	
LastModified *time.Time	
Size         *int64	
StorageClass *string}

For reference, it is in s3client.go

pkg/splunk/enterprise/util.go
gaurav-splunk: Hmm good point. We are actually supporting multiple appSources, so this might bite us. We can just log an error and continue with the next appSource. Thoughts @sgontla and @jryb ?

pkg/splunk/client/s3client.go
gaurav-splunk: I think it's better to make it clearer in common_types.go itself.

pkg/splunk/client/awss3client.go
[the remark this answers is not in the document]
gaurav-splunk: As per our discussion, we will now return up to 4k keys.

gaurav-splunk:
@akondur @smohan-splunk @jryb and @sgontla I have created another fresh PR with all the review comments incorporated. This is the link to the fresh PR: https://github.com/splunk/splunk-operator/pull/344
Apologies for all the mess. I will close and decline this PR soon.

PR #337, opened 2021-05-14
Manually fix nightly build after Operator-SDK v0.18.2 upgrade

jryb:
After upgrading operator-sdk to v0.18.2 the nightly builds started to fail due to https://github.com/kubernetes/kubernetes/issues/91395 which is an issue with how Kubernetes 1.18 interprets the operator-sdk v0.18.2 generated CRDs.

The issue manifests as :
"The CustomResourceDefinition “clustermasters.enterprise.splunk.com” is invalid: spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[serviceTemplate].properties[spec].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property"

### DIFF
--- a/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
@@ -1009,6 +1009,7 @@ spec:
                               x-kubernetes-int-or-string: true
                           required:
                           - port
+                          - protocol
                           type: object
                         type: array
                         x-kubernetes-list-map-keys:
--- a/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
@@ -967,6 +967,7 @@ spec:
                               x-kubernetes-int-or-string: true
                           required:
                           - port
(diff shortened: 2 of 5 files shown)


vebken-splunk (approved):

sgontla (approved):

gaurav-splunk (approved):

akondur:
@jryb Does an operator-sdk generate break this? i.e we run the command operator-sdk generate for every CR spec change and will that remove this line?

smohan-splunk (approved):

vebken-splunk:
> @jryb Does an operator-sdk generate break this? i.e we run the command operator-sdk generate for every CR spec change and will that remove this line?

Yes this will be the case, I have been looking into a better solution to automate this process or workaround the issue altogether. 

Other people who have hit the issue (https://github.com/kubernetes-sigs/controller-tools/issues/444)  mention upgrading their controller-gen version works around the issue, but I don't believe we use controller-gen to generate the crds. There are some other ideas mentioned here https://github.com/kubernetes/kubernetes/issues/91395, but the automated ones seem to require modifying k8s api code.

PR #338, opened 2021-05-16
CSPL-1004: Create operator component to vett Apps in S3 & Maintain Ap…

sgontla:
App Framework context state machine support.


deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -1009,7 +1009,6 @@ spec:
                               x-kubernetes-int-or-string: true
                           required:
                           - port
-                          - protocol
```
sgontla, line 1012: :-)

pkg/splunk/client/enterprise.go
```diff
@@ -28,6 +28,26 @@ import (
...
+// BucketEntry struct of index data response returned by /data/indexes endpoint
+type RemoteObject struct {
+	Etag *string //`json:"hash"`
+	Key  *string //`json:"name"`
+	//	LastModified *time.Time //`json:"modified"`
+	Owner        *Owner  //`json:"owner"`
+	Size         *int64  //`json:"size"`
+	StorageClass *string //`json:"class"`
+}
+
```
sgontla, line 50: To be removed, once after Gaurav's changes are in.

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
gaurav-splunk: welcome to the party ;)

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+	var found bool
+	var appName string
+	var newAppInfoList []enterprisev1.AppDeploymentInfo
+	var appChangesDetected bool
+	var appDeployInfo enterprisev1.AppDeploymentInfo
+
+	for _, remoteObj := range remoteS3ObjList {
+		receivedKey := *remoteObj.Key
+		if !isAppExtentionValid(receivedKey) {
+			scopedLog.Info("App name Parsig", " Ignoring the key: %s, with invalid extention for the app", receivedKey)
```
smohan-splunk, line 459: Nit: "App name **Parsing**"

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+		// Now update App status as seen in the remote listing
+		found = false
+		for _, appDeployInfo = range appSrcDeploymentInfo.AppDeploymentInfoList {
+			if appDeployInfo.AppName == appName {
+				found = true
+				if appDeployInfo.ObjectHash != *remoteObj.Etag {
+					appDeployInfo.ObjectHash = *remoteObj.Etag
+					appDeployInfo.DeployStatus = enterprisev1.DeployStatusPending
+					appChangesDetected = true
+				} else if appDeployInfo.RepoState == enterprisev1.RepoStateDeleted {
```
smohan-splunk, line 479: How about this scenario:
1. App is removed from remoteStore -  so the `RepoState` is set to `RepoStateDeleted`. I guess rest of the App metadata (such as `ObjectHash`) is preserved
2. App is updated offline & then copied back to remoteStore - as per this logic `DeployStatus` is set to `DeployStatusPending` but `RepoState` will never be set to `RepoStateActive`. Also `Size` isn't updated either

One way of handling I can think of:
```
			if appDeployInfo.AppName == appName {
				found = true
				// App has either been updated or enabled back
				if appDeployInfo.ObjectHash != *remoteObj.Etag ||  appDeployInfo.RepoState == enterprisev1.RepoStateDeleted {
					scopedLog.Info("App %s has either been updated or enabled back", appName)
					appDeployInfo.ObjectHash = *remoteObj.Etag
					appDeployInfo.Size = uint64(*remoteObj.Size)
					appDeployInfo.DeployStatus = enterprisev1.DeployStatusPending
					appDeployInfo.RepoState = enterprisev1.RepoStateActive
					appChangesDetected = true
				} 
				break
```

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+			appDeployInfo.RepoState = enterprisev1.RepoStateActive
+			appDeployInfo.DeployStatus = enterprisev1.DeployStatusPending
+			appChangesDetected = true
+
+			// Add it to a seperate list so that we don't loop through the newly added entries
+			newAppInfoList = append(newAppInfoList, appDeployInfo)
+		}
+	}
+
+	// Add the newly discovered Apps to the App source group
```
smohan-splunk, line 505: May be do this only when `newAppInfoList` is not empty?

pkg/splunk/enterprise/util_test.go
```diff
@@ -331,3 +335,265 @@ func TestGetSmartstoreRemoteVolumeSecrets(t *testing.T) {
...
+	var S3Response splclient.S3Response
+
+	// Test-1: Empty remoteObjectList Map should return an error
+	retStatus, err = handleAppRepoChanges(client, &cr, &appDeployContext, remoteObjListMap, &appFramworkConf)
+
+	if err != nil {
+		t.Errorf("Empty remote Object list should not trigger an error, but got error : %v", err)
+	}
+
+	// Test-2: Valid remoteObjectList should not cause an errror
```
smohan-splunk, line 411: Nit: error

pkg/splunk/enterprise/util_test.go
```diff
@@ -331,3 +335,265 @@ func TestGetSmartstoreRemoteVolumeSecrets(t *testing.T) {
...
+		t.Errorf("Could not handle a valid remote listing. Error: %v", err)
+	}
+
+	err = validateAppSrcDeployInfoByStateAndStatus(appFramworkConf.AppSources[0].Name, appDeployContext.AppsSrcDeployStatus, enterprisev1.RepoStateActive, enterprisev1.DeployStatusPending)
+	if err != nil {
+		t.Errorf("Unable to detect the change, when the object changed. Error: %v", err)
+	}
+
+	// Test-8:  For an AppSrc, when all the Apps are deleted on remote store and re-introduced, should modify the state to active and pending
+	setStateAndStatusForAppDeployInfoList(appDeployContext.AppsSrcDeployStatus[appFramworkConf.AppSources[0].Name].AppDeploymentInfoList, enterprisev1.RepoStateDeleted, enterprisev1.DdeployStatusComplete)
```
smohan-splunk, line 498: A test like this might be helpful
// Test-8.1: For an AppSrc, when an App is deleted on remote store, updated offline and re-introduced, should modify the App state to active and Deployment status to pending

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+func checkIfAppSrcExistsWithRemoteListing(appSrc string, remoteObjListingMap map[string]splclient.S3Response) bool {
+	if _, ok := remoteObjListingMap[appSrc]; ok {
+		return true
+	}
+
+	return false
+}
+
+// changeAppSrcDeployInfoStatus sets the new status to all the apps in an AppSrc if the given repo state and deploy status matches
+// primarly used in Phase-3
```
smohan-splunk, line 319: How will this be used in Phase 3?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -1009,7 +1010,6 @@ spec:
                               x-kubernetes-int-or-string: true
                           required:
                           - port
```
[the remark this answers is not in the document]
sgontla: We will address manually, OR, through a sed command somewhere in scripts, till we have a fix. So that it will not affect others.

pkg/splunk/enterprise/util.go
sgontla: high-level idea is, when we ask the sidecar with App source info(s3 info, and secrets), and when the deployment work is done, we change all the apps under that app source to the complete state. We still have some unknowns at this time how we handle a single app failure case, otherwise, we are handling the App source as a single unit.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+	}
+
+	// ToDo: sgontla: Ideally, this check should go to the reconcile entry point once the glue logic in place.
+	if appDeployContext.AppsSrcDeployStatus == nil {
+		appDeployContext.AppsSrcDeployStatus = make(map[string]enterprisev1.AppSrcDeployInfo)
+	}
+
+	// 1. Go through each AppSrc in my local list, and look for any changes
+	for appSrc, appSrcDeploymentInfo := range appDeployContext.AppsSrcDeployStatus {
+		// 1.1 Check if the AppSrc is deleted in latest config, is so, mark them for deletion
```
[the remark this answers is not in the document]
sgontla: initially, i had two different cases, later merged. will address it.

pkg/splunk/enterprise/util.go
sgontla: Line 481 sets the repoState to Active again. Size, Ownership, last modified fields we are intentionally not adding, for better scaling. But these are printed as part of the logs and can be checked on remote stores at any time.

pkg/splunk/enterprise/util.go
sgontla: Go auto covers in case of nil list. i.e. the existing list is not re-allocated as long as the existing length doesn't cross its capacity.

pkg/splunk/enterprise/util_test.go
sgontla: I think test-7 is the same as the scenario.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+				if !checkIfAnAppIsActiveOnRemoteStore(currentList[appIdx].AppName, s3Response.Objects) {
+					setStateAndStatusForAppDeployInfo(currentList[appIdx], enterprisev1.RepoStateDeleted, enterprisev1.DeployStatusPending)
+					appDeployContext.IsDeploymentInProgress = true
+				}
+			}
+		}
+
+		// 2.2 Check for any App changes(Ex. A new App source, a new App added/updated)
+		if createOrUpdateAppSrcDeploymentInfo(&appSrcDeploymentInfo, s3Response.Objects) {
+			appDeployContext.IsDeploymentInProgress = true
```
jryb, line 417: So is it correct to say that if the `IsDeploymentInProgress` flag is true, then theres a change in the Apps we need to deployment and we can use it as a trigger to kick off App installation (via Ansible/pod restart)?

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+	case "tgz":
+		return true
+
+	default:
+		return false
+	}
+}
+
+// createOrUpdateAppSrcDeploymentInfo  modifies the App deployment status as perceived from the remote object listing
+func createOrUpdateAppSrcDeploymentInfo(appSrcDeploymentInfo *enterprisev1.AppSrcDeployInfo, remoteS3ObjList []*splclient.RemoteObject) bool {
```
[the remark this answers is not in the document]
sgontla: Good point. Yes, since it is part of CR, describing the CR(-o json)  should give us the details in a readable format.

pkg/splunk/enterprise/util.go
sgontla: Sorry, I overlooked the scenario. Will address this case more wholistically.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+
+	return appChangesDetected
+}
+
+// markAppsStatusToComplete sets the required status for a given state.
+// Gets called from glue logic based on how we want to hand-off to init/side car, and look for the return status
+// For now, two possible cases:
+// 1. Completing the changes for Deletes. Called with state=AppStateDeleted, and status=DeployStatusPending
+// 2. Completing the changes for Active(Apps newly added, apps modified, Apps previously deleted, and now active).
+// Note:- Used in only for Phase-2
```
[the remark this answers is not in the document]
sgontla: Correct. In our glue logic(yet to be put), the caller should call this if and only if the Pod comes up successfully, and when the IsDeploymentInProgress flag is 'true'.

jryb (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+// clinet and cr are used when we put the glue logic to hand-off to the side car
+func handleAppRepoChanges(client splcommon.ControllerClient, cr splcommon.MetaObject,
+	appDeployContext *enterprisev1.AppDeploymentContext, remoteObjListingMap map[string]splclient.S3Response, appFrameworkConfig *enterprisev1.AppFrameworkSpec) (bool, error) {
+	crKind := cr.GetObjectKind().GroupVersionKind().Kind
+	scopedLog := log.WithName("handleAppRepoChanges").WithValues("kind", crKind, "name", cr.GetName(), "namespace", cr.GetNamespace())
+	var err error
+
+	scopedLog.Info("App Listing", " received App listing for %d app sources", len(remoteObjListingMap))
+	if remoteObjListingMap == nil || len(remoteObjListingMap) == 0 {
+		scopedLog.Error(nil, "remoteObjectList is empty")
```
kashok-splunk, line 368: won't throw err here? Just logging it?

sgontla: Ya, we intentionally not returning an error, to cover the case in which the customer wants to disable/delete all the apps that were deployed.

kashok-splunk (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -395,3 +395,241 @@ func GetAppListFromS3Bucket(client splcommon.ControllerClient, cr splcommon.Meta
...
+// 1. Completing the changes for Deletes. Called with state=AppStateDeleted, and status=DeployStatusPending
+// 2. Completing the changes for Active(Apps newly added, apps modified, Apps previously deleted, and now active).
+// Note:- Used in only for Phase-2
+func markAppsStatusToComplete(appSrcDeplymentStatus map[string]enterprisev1.AppSrcDeployInfo) error {
+	var err error
+	scopedLog := log.WithName("markAppsStatusToComplete")
+
+	// ToDo: sgontla: Passing appSrcDeplymentStatus is redundant, but this function will go away in phase-3, so ok for now.
+	for appSrc := range appSrcDeplymentStatus {
+		changeAppSrcDeployInfoStatus(appSrc, appSrcDeplymentStatus, enterprisev1.RepoStateActive, enterprisev1.DeployStatusInProgress, enterprisev1.DdeployStatusComplete)
```
gaurav-splunk, line 627: nit: s/enterprisev1.DdeployStatusComplete/enterprisev1.DeployStatusComplete

pkg/splunk/enterprise/util.go
```diff
@@ -395,3 +395,241 @@ func GetAppListFromS3Bucket(client splcommon.ControllerClient, cr splcommon.Meta
...
+		changeAppSrcDeployInfoStatus(appSrc, appSrcDeplymentStatus, enterprisev1.RepoStateDeleted, enterprisev1.DeployStatusInProgress, enterprisev1.DdeployStatusComplete)
```
gaurav-splunk, line 628: nit: s/enterprisev1.DdeployStatusComplete/enterprisev1.DeployStatusComplete

gaurav-splunk (changes requested):

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -229,14 +257,32 @@ type AppFrameworkSpec struct {
...
 
-	// IsDeploymentInProgress indicates if the App deployment in progress
+	// IsDeploymentInProgress indicates if the Apps deployment is in progress
 	IsDeploymentInProgress bool `json:"isDeploymentInProgress"`
 
 	// List of App package (*.spl, *.tgz) locations on remote volume
 	AppFrameworkConfig AppFrameworkSpec `json:"appRepo,omitempty"`
+
+	// Represents the Apps deployment status
+	AppsSrcDeployStatus map[string]AppSrcDeployInfo `json:"appSrcDeployStatus,omitempty"`
```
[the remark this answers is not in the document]
sgontla: Initial days, when I was playing CRD generation didn't allow this format(not sure if it was due to SDK limitation), and IIRC, there were some known issues with the SDK, so had to go with this. 

On the other side, IMO, it is a cleaner way to accommodate any other App src level context for the future, if we want to track the progress etc... at App src level as the feature evolves.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
 
 	return err
 }
+
+// checkIfAnAppIsActive checks if the App is listed as part of the AppSrc listing
+func checkIfAnAppIsActiveOnRemoteStore(appName string, list []*splclient.RemoteObject) bool {
```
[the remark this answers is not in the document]
sgontla: As we are handling a unique need(prefix on a struct member) here, I guess, this is fine for now.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+		if strings.HasSuffix(*list[i].Key, appName) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// checkIfAppSrcExistsWithRemoteListing checks if a given AppSrc is part of the remote listing
+func checkIfAppSrcExistsWithRemoteListing(appSrc string, remoteObjListingMap map[string]splclient.S3Response) bool {
```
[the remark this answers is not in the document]
sgontla: I just tried this, and it seems like the caller also needs to pass the type as interface{}. Am keeping it this way for now, but we can take it later if that is something we can achieve easily without too much re-org.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+// handleAppRepoChanges parses the remote storage listing and updates the repoState and deployStatus accordingly
+// clinet and cr are used when we put the glue logic to hand-off to the side car
+func handleAppRepoChanges(client splcommon.ControllerClient, cr splcommon.MetaObject,
+	appDeployContext *enterprisev1.AppDeploymentContext, remoteObjListingMap map[string]splclient.S3Response, appFrameworkConfig *enterprisev1.AppFrameworkSpec) (bool, error) {
+	crKind := cr.GetObjectKind().GroupVersionKind().Kind
+	scopedLog := log.WithName("handleAppRepoChanges").WithValues("kind", crKind, "name", cr.GetName(), "namespace", cr.GetNamespace())
+	var err error
+
+	scopedLog.Info("App Listing", " received App listing for %d app sources", len(remoteObjListingMap))
+	if remoteObjListingMap == nil || len(remoteObjListingMap) == 0 {
```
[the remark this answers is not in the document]
sgontla: Its simple sanity logging, and is good to have in between the package hand-offs, as the code evolves over a period of time.

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +294,236 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+			!checkIfAppSrcExistsWithRemoteListing(appSrc, remoteObjListingMap) {
+			scopedLog.Info("App Listing:", " deleting/disabling all the apps for App source: %s. Reason: App source deleted", appSrc)
+			curAppDeployList := appSrcDeploymentInfo.AppDeploymentInfoList
+			if setStateAndStatusForAppDeployInfoList(curAppDeployList, enterprisev1.RepoStateDeleted, enterprisev1.DeployStatusPending) {
+				appDeployContext.IsDeploymentInProgress = true
+			}
+		}
+
+		// Finally update the Map entry with latest info
+		appDeployContext.AppsSrcDeployStatus[appSrc] = appSrcDeploymentInfo
```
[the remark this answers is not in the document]
sgontla: As discussed, lists are always through the reference during the assignments, so we are good here. List memory allocation will not change unless the length crosses the capacity.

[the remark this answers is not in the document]
sgontla: Probably, May be a better improvement we can do is to return the list. Added a comment, so that i don't lose it :-)

pkg/splunk/enterprise/util.go
```diff
@@ -395,3 +395,241 @@ func GetAppListFromS3Bucket(client splcommon.ControllerClient, cr splcommon.Meta
...
+	for appSrc, s3Response := range remoteObjListingMap {
+		// 2.1 Mark Apps for deletion if they are missing in remote listing
+		appSrcDeploymentInfo, appSrcExistsLocally := appDeployContext.AppsSrcDeployStatus[appSrc]
+
+		if appSrcExistsLocally {
+			currentList := appSrcDeploymentInfo.AppDeploymentInfoList
+			for appIdx := range currentList {
+				if !checkIfAnAppIsActiveOnRemoteStore(currentList[appIdx].AppName, s3Response.Objects) {
+					scopedLog.Info("App change", "deleting/disabling the App: ", currentList[appIdx].AppName, "as it is missing in the remote listing")
+					setStateAndStatusForAppDeployInfo(currentList[appIdx], enterprisev1.RepoStateDeleted, enterprisev1.DeployStatusPending)
```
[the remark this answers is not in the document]
sgontla: Line 492 is taking care of the AppSrc level handling. Here it is the case, in which AppSrc exists, but there is a missing delta from the last remote listing.

gaurav-splunk (approved):

smohan-splunk (approved):

PR #344, opened 2021-05-17
Initial commit for S3 framework APIs

gaurav-splunk:
1.This PR targets framework, which will be used by operator to connect to remote storage(typically s3) to get the list of apps as specified in the CRs and which can be used later on to deploy and install the apps from remote storage.

2.This PR is the first part of the changes. Second part/PR will target UTs and code coverage for the related code.

### DIFF
--- a/pkg/apis/enterprise/v1/common_types.go
+++ b/pkg/apis/enterprise/v1/common_types.go
@@ -146,6 +146,12 @@ type VolumeSpec struct {
 
 	// Remote Storage type.
 	Type string `json:"storageType"`
+
+	// App Package Remote Store provider.
+	// For e.g. aws, azure, minio, etc.
+	// Currently we are only supporting aws.
+	// TODO: Support minio as well.
+	Provider string `json:"provider"`
 }
 
 // VolumeAndTypeSpec used to add any custom varaibles for volume implementation
--- a/pkg/splunk/client/awss3client.go
+++ b/pkg/splunk/client/awss3client.go
@@ -22,8 +22,8 @@ type AWSS3Client struct {
 	StartAfter         string
 }
 
-// GetAWSS3Client returns an AWS S3 client
-func GetAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
+// NewAWSS3Client returns an AWS S3 client
+func NewAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
 	return &AWSS3Client{
 		Region:             region,
 		BucketName:         bucketName,
@@ -36,12 +36,12 @@ func GetAWSS3Client(region string, bucketName string, accessKeyID string, secret
 
 //RegisterAWSS3Client will add the corresponding function pointer to the map
 func RegisterAWSS3Client() {
-	S3Clients["aws"] = GetAWSS3Client
+	S3Clients["aws"] = NewAWSS3Client
 }
 
 // GetAppsList get the list of apps from remote storage
 func (awsclient *AWSS3Client) GetAppsList() (S3Response, error) {
-	scopedLog := log.WithName("GetRemoteStorageClient")
+	scopedLog := log.WithName("GetAppsList")
 
 	scopedLog.Info("Getting Apps list", "AWS S3 Bucket", awsclient.BucketName)
 	s3Resp := S3Response{}
--- a/pkg/splunk/client/awss3client.go
+++ b/pkg/splunk/client/awss3client.go
@@ -1,0 +1,87 @@
+package client
+
+import (
+	"encoding/json"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+// blank assignment to verify that AWSS3Client implements S3Client
+var _ S3Client = &AWSS3Client{}
+
+// AWSS3Client is a client to implement S3 specific APIs
+type AWSS3Client struct {
+	Region             string
+	BucketName         string
+	AWSAccessKeyID     string
+	AWSSecretAccessKey string
+	Prefix             string
+	StartAfter         string
+}
+
+// GetAWSS3Client returns an AWS S3 client
+func GetAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
+	return &AWSS3Client{
+		Region:             region,
+		BucketName:         bucketName,
+		AWSAccessKeyID:     accessKeyID,
+		AWSSecretAccessKey: secretAccessKey,
+		Prefix:             prefix,
+		StartAfter:         startAfter,
+	}
+}
+
+//RegisterAWSS3Client will add the corresponding function pointer to the map
+func RegisterAWSS3Client() {
+	S3Clients["aws"] = GetAWSS3Client
+}
+
+// GetAppsList get the list of apps from remote storage
+func (awsclient *AWSS3Client) GetAppsList() (S3Response, error) {
+	scopedLog := log.WithName("GetRemoteStorageClient")
+
+	scopedLog.Info("Getting Apps list", "AWS S3 Bucket", awsclient.BucketName)
+
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(awsclient.Region),
+		Credentials: credentials.NewStaticCredentials(
+			awsclient.AWSAccessKeyID,     // id
+			awsclient.AWSSecretAccessKey, // secret
+			"")},                         //token
+	)
+
+	// Create S3 service client
+	svc := s3.New(sess)
+
+	options := &s3.ListObjectsV2Input{
+		Bucket:     aws.String(awsclient.BucketName),
+		Prefix:     aws.String(awsclient.Prefix),
+		StartAfter: aws.String(awsclient.StartAfter), //exclude the directory itself from listing
+		MaxKeys:    aws.Int64(4000),                  // return upto 4K keys from S3
+	}
+
+	s3Resp := S3Response{}
+	// List the bucket contents
+	resp, err := svc.ListObjectsV2(options)
+	if err != nil {
+		scopedLog.Error(err, "Unable to list items in bucket ", awsclient.BucketName)
+		return s3Resp, err
+	}
+
+	tmp, err := json.Marshal(resp.Contents)
+	if err != nil {
+		scopedLog.Error(err, "Failed to marshal s3 response")
+		return s3Resp, err
+	}
+
+	err = json.Unmarshal(tmp, &(s3Resp.Objects))
+	if err != nil {
+		scopedLog.Error(err, "Failed to unmarshal s3 response")
+		return s3Resp, err
+	}
+
+	return s3Resp, nil
+}
--- a/pkg/splunk/client/awss3client.go
+++ b/pkg/splunk/client/awss3client.go
@@ -44,14 +44,19 @@ func (awsclient *AWSS3Client) GetAppsList() (S3Response, error) {
 	scopedLog := log.WithName("GetRemoteStorageClient")
 
 	scopedLog.Info("Getting Apps list", "AWS S3 Bucket", awsclient.BucketName)
+	s3Resp := S3Response{}
 
 	sess, err := session.NewSession(&aws.Config{
 		Region: aws.String(awsclient.Region),
 		Credentials: credentials.NewStaticCredentials(
 			awsclient.AWSAccessKeyID,     // id
 			awsclient.AWSSecretAccessKey, // secret
-			"")},                         //token
+			"")},                         // token
 	)
+	if err != nil {
+		scopedLog.Error(err, "Unable to create a new aws s3 session", "AWS S3 Bucket", awsclient.BucketName)
+		return s3Resp, err
+	}
 
 	// Create S3 service client
 	svc := s3.New(sess)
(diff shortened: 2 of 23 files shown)


gaurav-splunk:
Sorry for the mess but this is the fesh PR and also has initial review comments from original PR(https://github.com/splunk/splunk-operator/pull/334)

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -146,6 +146,12 @@ type VolumeSpec struct {
 
 	// Remote Storage type.
 	Type string `json:"storageType"`
+
+	// App Package Remote Store provider.
+	// For e.g. aws, azure, minio, etc.
+	// Currently we are only supporting aws.
+	// TODO: Support minio as well.
+	Provider string `json:"provider"`
```
sgontla, line 154: Nit: As discussed earlier, unless we really need to differentiate a particular client implementation, storage type(above) should be good enough to differentiate?

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +342,53 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+
+	scopedLog := log.WithName("GetAppListFromS3Bucket").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	sourceToAppListMap := make(map[string]splclient.S3Response)
+
+	volMap := make(map[string]enterprisev1.VolumeSpec)
+
+	for _, vol := range appFrameworkRef.VolList {
+		volMap[vol.Name] = vol
+	}
```
sgontla, line 357: Nit: checkIfVolumeExists returns the index of the volume, if you want to use the slice directly, instead of going to a map.

jryb (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -292,3 +342,53 @@ func DeleteOwnerReferencesForS3SecretObjects(client splcommon.ControllerClient,
...
+			continue
+		}
+
+		// Now, get the apps list from remote storage
+		s3Response, err = s3Client.GetAppsList()
+		if err != nil {
+			// move on to the next appSource if we are not able to get apps list
+			scopedLog.Error(err, "Unable to get apps list", "appSource", appSource.Name)
+			continue
+		}
```
sgontla, line 388: Wondering if we need to have a retry OR a carry-forward retry based on some flag?, otherwise, silently returning from this function can't trigger the retry. Can be an issue with a large probing interval.s

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+
+// GetAWSS3Client returns an AWS S3 client
+func GetAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
+	return &AWSS3Client{
+		Region:             region,
+		BucketName:         bucketName,
+		AWSAccessKeyID:     accessKeyID,
+		AWSSecretAccessKey: secretAccessKey,
+		Prefix:             prefix,
+		StartAfter:         startAfter,
```
sgontla, line 33: Is there a way, that we can end after the current level? If not, I think I need to take care of adding a check for this when consuming the response.

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+	scopedLog.Info("Getting Apps list", "AWS S3 Bucket", awsclient.BucketName)
+
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(awsclient.Region),
+		Credentials: credentials.NewStaticCredentials(
+			awsclient.AWSAccessKeyID,     // id
+			awsclient.AWSSecretAccessKey, // secret
+			"")},                         //token
+	)
+
```
sgontla, line 55: Do we need to handle the error here?

pkg/splunk/client/s3client.go
```diff
@@ -0,0 +1,40 @@
...
+	StorageClass *string
+}
+
+//RegisterS3Client registers the respective Client
+func RegisterS3Client(provider string) {
+	switch provider {
+	case "aws":
+		RegisterAWSS3Client()
+	default:
+	}
```
sgontla, line 39: Nit: default: can we log an error to avoid any future surprises :-)

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -76,23 +76,36 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
+		if err != nil {
+			scopedLog.Error(err, "Unable to get apps list from remote storage")
+			return result, err
+		}
+
+		for _, appSource := range cr.Spec.AppFrameworkConfig.AppSources {
+			scopedLog.Info("Apps List retrieved from remote storage", "App Source", appSource.Name, "Content", sourceToAppsList[appSource.Name].Objects)
+		}
+
+		cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
```
sgontla, line 106: Since GetAppListFromS3Bucket is not returning the error, and here at the end, we are updating the config. So, that way, we will not re-try in error case. I think we ONLY need to update the config unless all the app sources are listed properly OR having some additional tracking at the App source level.

pkg/splunk/enterprise/licensemaster.go
```diff
@@ -37,19 +38,31 @@ func ApplyLicenseMaster(client splcommon.ControllerClient, cr *enterprisev1.Lice
...
+		sourceToAppsList, err = GetAppListFromS3Bucket(client, cr, &cr.Spec.AppFrameworkConfig)
+		if err != nil {
+			scopedLog.Error(err, "Unable to get apps list from remote storage")
+			return result, err
+		}
+
+		for _, appSource := range cr.Spec.AppFrameworkConfig.AppSources {
+			scopedLog.Info("Apps List retrieved from remote storage", "App Source", appSource.Name, "Content", sourceToAppsList[appSource.Name].Objects)
+		}
 		cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
```
sgontla, line 66: Same as earlier: Do not update the config unless all the App sources are probed successfully.

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -47,12 +47,22 @@ func ApplySearchHeadCluster(client splcommon.ControllerClient, cr *enterprisev1.
...
+			scopedLog.Error(err, "Unable to get apps list from remote storage")
+			return result, err
+		}
+
+		for _, appSource := range cr.Spec.AppFrameworkConfig.AppSources {
+			scopedLog.Info("Apps List retrieved from remote storage", "App Source", appSource.Name, "Content", sourceToAppsList[appSource.Name].Objects)
+		}
 
 		cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
 	}
```
sgontla, line 68: same as above comment.

pkg/splunk/enterprise/standalone.go
```diff
@@ -67,12 +69,22 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
...
+		if err != nil {
+			scopedLog.Error(err, "Unable to get apps list from remote storage")
+			return result, err
+		}
+
+		for _, appSource := range cr.Spec.AppFrameworkConfig.AppSources {
+			scopedLog.Info("Apps List retrieved from remote storage", "App Source", appSource.Name, "Content", sourceToAppsList[appSource.Name].Objects)
+		}
 
 		cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
```
sgontla, line 89: same as the above comment.

pkg/splunk/enterprise/util.go
```diff
@@ -32,6 +34,54 @@ import (
...
+
+	scopedLog := log.WithName("GetRemoteStorageClient").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	//use the provider name to get the corresponding function pointer
+	getClient, _ := splclient.S3Clients[vol.Provider]
+
+	appSecretRef := vol.SecretRef
+	s3ClientSecret, err := splutil.GetSecretByName(client, cr, appSecretRef)
+	if err != nil {
+		return nil, err
```
sgontla, line 57: Nit: seems like a generic error. Probably, the best place to add an error is in GetSecretByName itself, I don't know why we didn't do it before :-).

sgontla (changes requested):

pkg/apis/enterprise/v1/common_types.go
gaurav-splunk: Yes subba, we need to differentiate based on the provider. storageType alone won't be enough as we need provider specific SDK libraries at the minimum to make it work.

pkg/splunk/client/awss3client.go
gaurav-splunk: I did not find any option to end at the current level.

gaurav-splunk: Ok, I just researched about it and there is actually a way to limit the listing to the current level only. I will take care of it in my code.

pkg/splunk/enterprise/standalone.go
```diff
@@ -67,12 +69,22 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
...
-	// ToDo sgontla: Handle
-	// 1. Spec update
-	// 2. S3 credentails change
-	// 3. Probe times expired
-	if !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
-		// TBD: Probe the remote store, and handle any changes on remote store
+	if cr.Spec.CommonSplunkSpec.Mock != true && !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
+		var sourceToAppsList map[string]splclient.S3Response
+
+		for _, vol := range cr.Spec.AppFrameworkConfig.VolList {
```
akondur, line 75: For clustermaster we have the following snippet of code:

```
for _, vol := range cr.Spec.AppFrameworkConfig.VolList {			
    if _, ok := splclient.S3Clients[vol.Provider]; !ok  {
				splclient.RegisterS3Client(vol.Provider)			
     }
}
```

Are we skipping the check on the entries in the map `splclient.S3Clients`

gaurav-splunk: Udpated on other CRs as well!

pkg/splunk/client/awss3client.go
gaurav-splunk: Done

pkg/splunk/enterprise/standalone.go
akondur: @gaurav-splunk , I still see a difference with clustermaster.go. Am I missing something here?

gaurav-splunk: Yeah you are missing my update commit which I haven't pushed yet :)

akondur: I'll resolve it now and approve it then :)

akondur (approved):
Looks great!

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+//RegisterAWSS3Client will add the corresponding function pointer to the map
+func RegisterAWSS3Client() {
+	S3Clients["aws"] = GetAWSS3Client
+}
+
+// GetAppsList get the list of apps from remote storage
+func (awsclient *AWSS3Client) GetAppsList() (S3Response, error) {
+	scopedLog := log.WithName("GetRemoteStorageClient")
```
kashok-splunk, line 44: Change this to "scopedLog := log.WithName("GetAppsList")"

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+	BucketName         string
+	AWSAccessKeyID     string
+	AWSSecretAccessKey string
+	Prefix             string
+	StartAfter         string
+}
+
+// GetAWSS3Client returns an AWS S3 client
+func GetAWSS3Client(region string, bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string) S3Client {
+	return &AWSS3Client{
```
kashok-splunk, line 27: Nit: Can we change this to NewAWSS3Client() to keep it consistent with NewSplunkClient()?

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+	s3Resp := S3Response{}
+	// List the bucket contents
+	resp, err := svc.ListObjectsV2(options)
+	if err != nil {
+		scopedLog.Error(err, "Unable to list items in bucket ", awsclient.BucketName)
+		return s3Resp, err
+	}
+
+	tmp, err := json.Marshal(resp.Contents)
+	if err != nil {
```
kashok-splunk, line 80: why do we need to Marshal the resp here first and then again unmarshal it?

pkg/splunk/enterprise/util.go
gaurav-splunk: So I will be checking the size of `sourceToAppListMap` returned from here and the size of appSources, if they are not same, we will reconcile. So, I think the reconcile will take care of the retry mechanism itself, right?

pkg/splunk/enterprise/standalone.go
```diff
@@ -38,6 +39,7 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
 		RequeueAfter: time.Second * 5,
 	}
 
+	scopedLog := log.WithName("ApplyStandalone").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
```
kashok-splunk, line 42: Nice catch of missing logs :)

pkg/splunk/client/awss3client.go
gaurav-splunk: This is taken care of at line 69.

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: Done!

pkg/splunk/enterprise/licensemaster.go
gaurav-splunk: Done!

pkg/splunk/enterprise/searchheadcluster.go
gaurav-splunk: Done!

pkg/splunk/client/awss3client.go
```diff
@@ -0,0 +1,87 @@
...
+		scopedLog.Error(err, "Failed to marshal s3 response")
+		return s3Resp, err
+	}
+
+	err = json.Unmarshal(tmp, &(s3Resp.Objects))
+	if err != nil {
+		scopedLog.Error(err, "Failed to unmarshal s3 response")
```
kashok-splunk: Can we use "fmt.Errorf" here to maintain the consistency in client

pkg/splunk/enterprise/standalone.go
gaurav-splunk: Done!

pkg/splunk/enterprise/standalone.go
gaurav-splunk: Done!

pkg/splunk/client/awss3client.go
gaurav-splunk: I tried unmarshalling it straight away but it turns out structs can't be unmarshalled directly. Hence I had to marshal it first.

kashok-splunk: You need to read the response body first and then unmarshal output. Something like 
      `data, _ := ioutil.ReadAll(resp.Body)
	err= json.Unmarshal(data, &(s3Resp.Objects))`

gaurav-splunk: Yeah there is nothing in `resp` that is of the format `io.Reader`. We could do that if the returned response is of type `http.Response`, but not in this case.

sgontla (approved):

pkg/splunk/enterprise/util.go
gaurav-splunk: Ok, i'll add an error log in GetSecretByName

smohan-splunk (approved):

PR #348, opened 2021-05-19
CSPL-1063-fix-nightly-run-failure

pdhanoyasplunk:
Reduced parallel test run to avoid nightly failure on eks cluster 
Changed cluster name for eks cluster to circumvent existing resource issue

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,8 +105,9 @@ executors:
       # Set cluster workers to 5 (5 cluster nodes)
       # NUM_NODES represent number of parallel test executions
       # NUM_WORKERS represent number of nodes in a k8 cluster
-      NUM_NODES: 8
-      NUM_WORKERS: 10
+      NUM_NODES: 5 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -105,8 +105,9 @@ executors:
       # Set cluster workers to 5 (5 cluster nodes)
       # NUM_NODES represent number of parallel test executions
       # NUM_WORKERS represent number of nodes in a k8 cluster
-      NUM_NODES: 8
-      NUM_WORKERS: 10
+      NUM_NODES: 5
```
akondur, line 108: Curious on how this made a difference? Do we not have access to enough nodes/workers?

smohan-splunk: The value was lower earlier, and we were experimenting to see if bumping up the nodes would reduce the run time of tests. Apparently this is now causing resource cleanup issues

smohan-splunk (approved):

PR #353, opened 2021-05-25
CSPL-1068 common utils for app framework automation.

pdhanoyasplunk:
Added common utils details below:

* Util to get directories under given path on a pod
* Util to check if a string exists in a given string slice
* Util to check two string slices match

### DIFF
--- a/test/testenv/util.go
+++ b/test/testenv/util.go
@@ -20,6 +20,8 @@ import (
 	"math/rand"
 	"os/exec"
 	"path"
+	"reflect"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jryb (approved):
Interesting that go doesn't have lib functions for this.  Looks good to me.

test/testenv/util.go
```diff
@@ -551,3 +553,38 @@ func ExecuteCommandOnPod(deployment *Deployment, podName string, stdin string) (
 	logf.Log.Info("Command executed on pod", "pod", podName, "command", command, "stdin", stdin, "stdout", stdout, "stderr", stderr)
 	return stdout, nil
 }
+
+// GetSubDirsOnPod returns subdirectory under given path on the given POD
+func GetSubDirsOnPod(deployment *Deployment, podName string, path string) ([]string, error) {
```
tpereirasplunk, line 558: Do we need to check if there are files in the directory?  Since the current result would be a mix of files and folders

smohan-splunk (approved):

jambrosiano (approved):

tpereirasplunk (approved):

PR #357, opened 2021-05-26
CSPL-1081: Automation for app framework phase 2

tpereirasplunk:
Added Test case utils details below:

* Utility function to get the install status and app version for an app

* Verification util to iterate and check multiple apps

* Verification to validate apps are copied/downloaded to location

### DIFF
--- a/test/testenv/appframework_utils.go
+++ b/test/testenv/appframework_utils.go
@@ -1,7 +1,11 @@
 package testenv
 
 import (
+	"fmt"
+	"strings"
+
 	enterprisev1 "github.com/splunk/splunk-operator/pkg/apis/enterprise/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // GenerateAppSourceSpec return AppSourceSpec struct with given values
@@ -11,4 +15,65 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
 		Location:             appSourceLocation,
 		AppSourceDefaultSpec: appSourceDefaultSpec,
 	}
+}
+
+// GetPodAppStatus Get the app install status and version number
+func GetPodAppStatus(deployment *Deployment, podName string, ns string, appname string) (string, string, error) {
+	output, _ := GetPodAppInstallStatus(deployment, podName, ns, appname)
+	status := strings.Fields(output)[1]
+	version, err := GetPodInstalledAppVersion(deployment, podName, ns, appname)
+	return status, version, err
+
+}
+
+// GetPodInstalledAppVersion Get the version of the app installed on pod
+func GetPodInstalledAppVersion(deployment *Deployment, podName string, ns string, appname string) (string, error) {
+	path := "etc/apps"
+	if strings.Contains(podName, "-indexer-") {
+		path = "etc/slave-apps/_cluster/default"
+	}
+	filePath := fmt.Sprintf("/opt/splunk/%s/%s/default/app.conf", path, appname)
+
+	confline, err := GetConfLineFromPod(podName, filePath, ns, "Version", "", false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to get Version from pod", "Pod Name", podName)
+		return "", err
+	}
+	version := strings.TrimSpace(strings.Split(confline, " ")[2])
+	return version, err
+
+}
+
+// GetPodAppInstallStatus Get the app install status
+func GetPodAppInstallStatus(deployment *Deployment, podName string, ns string, appname string) (string, error) {
+	stdin := fmt.Sprintf("/opt/splunk/bin/splunk display app '%s'", appname)
+	command := []string{"/bin/sh"}
+	stdout, stderr, err := deployment.PodExecCommand(podName, command, stdin, false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to execute command on pod", "pod", podName, "command", command, "stdin", stdin)
+		return "", err
+	}
+	logf.Log.Info("Command executed on pod", "pod", podName, "command", command, "stdin", stdin, "stdout", stdout, "stderr", stderr)
+
+	return strings.TrimSuffix(stdout, "\n"), nil
+}
+
+// GetPodAppbtoolStatus Get the app btool status
+func GetPodAppbtoolStatus(deployment *Deployment, podName string, ns string, appname string) (string, error) {
+	stdin := fmt.Sprintf("/opt/splunk/bin/splunk btool %s --app=%s --debug", appname, appname)
+	command := []string{"/bin/sh"}
+	stdout, stderr, err := deployment.PodExecCommand(podName, command, stdin, false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to execute command on pod", "pod", podName, "command", command, "stdin", stdin)
+		return "", err
+	}
+	logf.Log.Info("Command executed on pod", "pod", podName, "command", command, "stdin", stdin, "stdout", stdout, "stderr", stderr)
+
+	if len(stdout) > 0 {
+		if strings.Contains(strings.Split(stdout, "\n")[0], "Application is disabled") {
+			return "DISABLED", nil
+		}
+		return "ENABLED", nil
+	}
+	return "", err
 }
--- a/test/testenv/s3utils.go
+++ b/test/testenv/s3utils.go
@@ -2,6 +2,7 @@ package testenv
 
 import (
 	"os"
+	"path"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -99,8 +100,8 @@ func UploadFileToS3(dataBucket string, filename string, filepath string, file *o
 		uploader := s3manager.NewUploader(sess)
 		numBytes, err := uploader.Upload(&s3manager.UploadInput{
 			Bucket: aws.String(dataBucket),
-			Key:    aws.String(filename), // Name of the file to be saved
-			Body:   file,                 // File
+			Key:    aws.String(path.Join(filepath, filename)), // Name of the file to be saved
+			Body:   file,                                      // File
 		})
 		if err != nil {
 			logf.Log.Error(err, "Error in file upload")
--- a/test/testenv/verificationutils.go
+++ b/test/testenv/verificationutils.go
@@ -569,3 +569,39 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
 	}
 
 }
+
+// VerifyAppInstall verify that app of specific version is installed
+func VerifyAppInstall(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps map[string]string, versionCheck bool, statusCheck string) {
+	for appName, version := range apps {
+		status, versionInstalled, err := GetPodAppStatus(deployment, podName, ns, appName)
+		logf.Log.Info("App info returned for app", "App-name", appName, "status", status, "versionInstalled", versionInstalled, "error", err)
+		gomega.Expect(err).To(gomega.Succeed(), "Unable to get app status on pod ")
+		gomega.Expect(strings.ToLower(status)).Should(gomega.Equal(strings.ToLower(statusCheck)))
+		if versionCheck {
+			gomega.Expect(versionInstalled).Should(gomega.Equal(version))
+		}
+	}
+}
+
+// VerifyAppsCopied verify that apps are copied to correct location based on POD
+func VerifyAppsCopied(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps []string) {
+	path := "etc/apps"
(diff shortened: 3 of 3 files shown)


test/testenv/verificationutils.go
```diff
@@ -569,3 +569,12 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
 	}
 
 }
+
+// VerifyAppInstall verify that app is installed
+func VerifyAppInstall(deployment *Deployment, testenvInstance *TestEnv, ns string, podname string, apps []string) {
+	for _, appName := range apps {
+		status, _, err := GetAppStatus(deployment, podname, ns, appName)
+		gomega.Expect(status).Should(gomega.Equal("enabled"))
+		gomega.Expect(err).Should(gomega.Equal(""))
```
pdhanoyasplunk, line 578: Nit: Check for error before `Status`

pdhanoyasplunk (approved):

test/testenv/appframework_utils.go
```diff
@@ -12,3 +16,22 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
 		AppSourceDefaultSpec: appSourceDefaultSpec,
 	}
 }
+
+// GetAppStatus Get the app install status and enabled status
+func GetAppStatus(deployment *Deployment, podName string, ns string, appname string) (string, string, error) {
+	filePath := fmt.Sprintf("/opt/splunk/etc/apps/%s/default/app.conf", appname)
```
jryb, line 22: What if there is a `local/app.conf` that disables the app?  For ansible in the clustered cases the app is installed locally on the CM/Deployer, copied to the `manager-apps`/`shcluster`, pushed to the cluster, then disabled locally via `local/app.conf`:
```
[splunk@splunk-shc1-deployer-0 splunk]$ cat etc/apps/application1/local/app.conf 
[ui]

[launcher]

[install]
state = disabled
```

test/testenv/appframework_utils.go
```diff
@@ -12,3 +16,22 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
 		AppSourceDefaultSpec: appSourceDefaultSpec,
 	}
 }
+
+// GetAppStatus Get the app install status and enabled status
```
jryb, line 20: This would only be applicable for locally installed apps and not clustered apps, correct?

tpereirasplunk: Yes.... Wokring on the part to include clustered apps as well

pdhanoyasplunk: Can we rename the method to reflect that ?

gaurav-splunk (approved):

jambrosiano (approved):

smohan-splunk (approved):

test/testenv/appframework_utils.go
tpereirasplunk: Updated. Using splunk command and btool to check for app status

kashok-splunk (approved):

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,37 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
+		status, versionInstalled, err := GetPodAppStatus(deployment, podName, ns, appName)
+		gomega.Expect(err).Should(gomega.Equal(""))
+		gomega.Expect(status).Should(gomega.Equal(statusCheck))
+		if versionCheck {
+			gomega.Expect(versionInstalled).Should(gomega.Equal(version))
+		}
+	}
+}
+
+// VerifyAppsCopied verify that apps are copied to correct location based on CR spec
```
pdhanoyasplunk, line 585: NIT: is this based on POD Name or CR Spec ? Might need to refine comment.

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,37 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
+	if strings.Contains(podName, "-deployer-") {
+		path = "etc/shcluster/apps"
+	}
+	VerifyAppsInFolder(deployment, testenvInstance, ns, podName, apps, path)
+}
+
+// VerifyAppsInFolder verify that apps are present in folder
+func VerifyAppsInFolder(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps []string, path string) {
+	appList, err := GetSubDirsOnPod(deployment, podName, path)
+	gomega.Expect(err).To(gomega.Succeed(), "Apps not found in path", "Pod", podName)
```
pdhanoyasplunk, line 600: Nit: Refine the comment.... As it's not necessary that we failed to get apps but maybe failed to list dirs ?

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,37 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
 	}
 
 }
+
+// VerifyAppInstall verify that app of specific version is installed
+func VerifyAppInstall(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps map[string]string, versionCheck bool, statusCheck string) {
+	for appName, version := range apps {
+		status, versionInstalled, err := GetPodAppStatus(deployment, podName, ns, appName)
+		gomega.Expect(err).Should(gomega.Equal(""))
```
pdhanoyasplunk, line 577: Nit: This can be failed like this `gomega.Expect(err).To(Succeed(), "Unable to get app status on pod ")`
This way we fail with meaningful error.

test/testenv/verificationutils.go
pdhanoyasplunk: This can be something like `Unable to get apps on pod`

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,37 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
+		gomega.Expect(status).Should(gomega.Equal(statusCheck))
```
pdhanoyasplunk, line 578: Nit: Since its a string match can we add a comment for acceptable/expected return for `status`. Also might be a good idea to change both values to lower/upper case before comparison. We should also add log before assertion for meaningful error.

test/testenv/appframework_utils.go
```diff
@@ -12,3 +16,64 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
...
+	stdin := fmt.Sprintf("/opt/splunk/bin/splunk display app '%s'", appname)
+	command := []string{"/bin/sh"}
+	stdout, stderr, err := deployment.PodExecCommand(podName, command, stdin, false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to execute command on pod", "pod", podName, "command", command, "stdin", stdin)
+		return "Failed"
+	}
+	logf.Log.Info("Command executed on pod", "pod", podName, "command", command, "stdin", stdin, "stdout", stdout, "stderr", stderr)
+
+	return strings.TrimSuffix(stdout, "\n")
```
pdhanoyasplunk, line 58: Would be a good idea to return bool. Ignore my below comments if you decide to do that. 

My reasoning being the consumers of verification methods would not have to know the output of `"/opt/splunk/bin/splunk display app '%s'`

test/testenv/appframework_utils.go
```diff
@@ -12,3 +16,64 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
...
+}
+
+// GetPodAppInstallStatus Get the app install status
+func GetPodAppInstallStatus(deployment *Deployment, podName string, ns string, appname string) string {
+	stdin := fmt.Sprintf("/opt/splunk/bin/splunk display app '%s'", appname)
+	command := []string{"/bin/sh"}
+	stdout, stderr, err := deployment.PodExecCommand(podName, command, stdin, false)
+	if err != nil {
+		logf.Log.Error(err, "Failed to execute command on pod", "pod", podName, "command", command, "stdin", stdin)
+		return "Failed"
```
pdhanoyasplunk, line 54: Nit: Return error..so that the consumer can differentiate between failed to run the command vs app not present ? Doing via strings is the not most efficient way. To keep the code styling consistent we should return errors explicitly.

pdhanoyasplunk (changes requested):

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,37 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
+			gomega.Expect(versionInstalled).Should(gomega.Equal(version))
+		}
+	}
+}
+
+// VerifyAppsCopied verify that apps are copied to correct location based on CR spec
+func VerifyAppsCopied(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps []string) {
+	path := "etc/apps"
+	if strings.Contains(podName, "cluster-master") {
+		path = "etc/master-apps/_cluster/default"
```
jryb, line 589: Should this be `etc/master-apps` instead of `_cluster/default`?

tpereirasplunk: when running locally it seemed to be under this path. We can modify based on tests

jryb (approved):

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,38 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
+	VerifyAppsInFolder(deployment, testenvInstance, ns, podName, apps, path)
+}
+
+// VerifyAppsInFolder verify that apps are present in folder
+func VerifyAppsInFolder(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps []string, path string) {
+	appList, err := GetSubDirsOnPod(deployment, podName, path)
+	gomega.Expect(err).To(gomega.Succeed(), "Unable to get apps on pod", "Pod", podName)
+	for _, app := range apps {
+		found := CheckStringInSlice(appList, app)
+		gomega.Expect(found).Should(gomega.Equal(true))
```
pdhanoyasplunk, line 604: Nit: Add a log line here. It will be really hard to debug the failures without any logging.

test/testenv/verificationutils.go
```diff
@@ -569,3 +569,38 @@ func VerifyAppFrameworkDeployment(standalone *enterprisev1.Standalone) {
...
 
 }
+
+// VerifyAppInstall verify that app of specific version is installed
+func VerifyAppInstall(deployment *Deployment, testenvInstance *TestEnv, ns string, podName string, apps map[string]string, versionCheck bool, statusCheck string) {
+	for appName, version := range apps {
+		status, versionInstalled, err := GetPodAppStatus(deployment, podName, ns, appName)
+		logf.Log.Info("App info returned for app", "App-name", appName, "status", status, "versionInstalled", versionInstalled)
+		gomega.Expect(err).To(gomega.Succeed(), "Unable to get app status on pod ")
+		gomega.Expect(strings.ToLower(status)).Should(gomega.Equal(strings.ToLower(statusCheck)))
```
pdhanoyasplunk, line 579: Add logs before the checks for help with debugging.

test/testenv/s3utils.go
```diff
@@ -99,8 +100,8 @@ func UploadFileToS3(dataBucket string, filename string, filepath string, file *o
 		uploader := s3manager.NewUploader(sess)
 		numBytes, err := uploader.Upload(&s3manager.UploadInput{
 			Bucket: aws.String(dataBucket),
-			Key:    aws.String(filename), // Name of the file to be saved
-			Body:   file,                 // File
+			Key:    aws.String(path.Join(filepath, filename)), // Name of the file to be saved
```
pdhanoyasplunk, line 103: Thanks for fixing this.

test/testenv/appframework_utils.go
```diff
@@ -12,3 +16,64 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
...
+	if err != nil {
+		logf.Log.Error(err, "Failed to get Version from pod", "Pod Name", podName)
+		return "", err
+	}
+	version := strings.TrimSpace(strings.Split(confline, " ")[2])
+	return version, err
+
+}
+
+// GetPodAppInstallStatus Get the app install status
```
pdhanoyasplunk, line 47: Would be a good idea to return err as well. That way caller decides how to fail. Return `false` when command fails is incorrect.

pdhanoyasplunk (changes requested):

test/testenv/verificationutils.go
tpereirasplunk: logs added in line just after status and version are retrieved.

test/testenv/appframework_utils.go
tpereirasplunk: modified to return err instead of boolean

pdhanoyasplunk (approved):

PR #366, opened 2021-06-08
CSPL-767: Operator trigger Instance pod to download app package from S3

sgontla:
- Added new configMap to update the latest apps status to the pod locations

### DIFF
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -83,15 +83,13 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 		}
 	}()
 
-	//check if the apps need to be downloaded from remote storage
+	initAppFrameWorkContext(&cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
+
+	// check if the apps need to be downloaded from remote storage
+	// ToDo: sgontla: Once after we have the flow, move the following logic into a seperate function(Note:- all errors should not cause a return in that function)
+	// ToDo: also take care of the Mock
 	if cr.Spec.CommonSplunkSpec.Mock != true && !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
 		var sourceToAppsList map[string]splclient.S3Response
-
-		for _, vol := range cr.Spec.AppFrameworkConfig.VolList {
-			if _, ok := splclient.S3Clients[vol.Provider]; !ok {
-				splclient.RegisterS3Client(vol.Provider)
-			}
-		}
 
 		sourceToAppsList, err = GetAppListFromS3Bucket(client, cr, &cr.Spec.AppFrameworkConfig)
 		if len(sourceToAppsList) != len(cr.Spec.AppFrameworkConfig.AppSources) {
@@ -106,6 +104,11 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 		err = handleAppRepoChanges(client, cr, &cr.Status.AppContext, sourceToAppsList, &cr.Spec.AppFrameworkConfig)
 		if err != nil {
 			scopedLog.Error(err, "Unable to use the App list retrieved from the remote storage")
+			return result, err
+		}
+
+		_, _, err := ApplyAppListingConfigMap(client, cr, &cr.Spec.AppFrameworkConfig, cr.Status.AppContext.AppsSrcDeployStatus)
+		if err != nil {
 			return result, err
 		}
 
@@ -208,9 +211,9 @@ func getClusterMasterStatefulSet(client splcommon.ControllerClient, cr *enterpri
 	if err != nil {
 		return ss, err
 	}
-	_, exists := getSmartstoreConfigMap(client, cr, SplunkClusterMaster)
+	smartStoreConfigMap := getSmartstoreConfigMap(client, cr, SplunkClusterMaster)
 
-	if exists {
+	if smartStoreConfigMap != nil {
 		setupInitContainer(&ss.Spec.Template, cr.Spec.Image, cr.Spec.ImagePullPolicy, commandForCMSmartstore)
 	}
 
@@ -233,9 +236,9 @@ func CheckIfsmartstoreConfigMapUpdatedToPod(c splcommon.ControllerClient, cr *en
 		return fmt.Errorf("Failed to check config token value on pod. stdout=%s, stderror=%s, error=%v", stdOut, stdErr, err)
 	}
 
-	configMap, exists := getSmartstoreConfigMap(c, cr, SplunkClusterMaster)
-	if exists {
-		tokenFromConfigMap := configMap.Data[configToken]
+	smartStoreConfigMap := getSmartstoreConfigMap(c, cr, SplunkClusterMaster)
+	if smartStoreConfigMap != nil {
+		tokenFromConfigMap := smartStoreConfigMap.Data[configToken]
 		if tokenFromConfigMap == stdOut {
 			scopedLog.Info("Token Matched.", "on Pod=", stdOut, "from configMap=", tokenFromConfigMap)
 			return nil
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -17,6 +17,7 @@ package enterprise
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -26,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	enterprisev1 "github.com/splunk/splunk-operator/pkg/apis/enterprise/v1"
+	splclient "github.com/splunk/splunk-operator/pkg/splunk/client"
 	splcommon "github.com/splunk/splunk-operator/pkg/splunk/common"
 	splctrl "github.com/splunk/splunk-operator/pkg/splunk/controller"
 	splutil "github.com/splunk/splunk-operator/pkg/splunk/util"
@@ -231,19 +233,6 @@ func getSplunkDefaults(identifier, namespace string, instanceType InstanceType, 
 			"default.yml": defaults,
 		},
 	}
-}
-
-// prepareSplunkSmartstoreConfigMap returns a K8 ConfigMap containing Splunk smartstore config in INI format
-func prepareSplunkSmartstoreConfigMap(identifier, namespace string, crKind string, dataIniMap map[string]string) *corev1.ConfigMap {
-	configMapIni := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      GetSplunkSmartstoreConfigMapName(identifier, crKind),
-			Namespace: namespace,
-		},
-	}
-	configMapIni.Data = dataIniMap
-
-	return configMapIni
 }
 
 // getSplunkPorts returns a map of ports to use for Splunk instances.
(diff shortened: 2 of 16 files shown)


pkg/splunk/enterprise/configuration.go
```diff
@@ -642,6 +669,20 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
...
+		for key := range appListingConfigMap.Data {
+			if key != configToken {
+				appListingFiles = append(appListingFiles, key)
+			}
+		}
+		// Always sort the slice, so that map entries are ordered, to avoid pod resets
+		sort.Strings(appListingFiles)
+
+		for _, fileName := range appListingFiles {
+			splunkDefaults = fmt.Sprintf("%s%s,%s", "/mnt/app-listing/", fileName, splunkDefaults)
```
jryb, line 683: Since the Deployer and the SHC Member have the same CR, does this get added to the defaults for the SHC member as well?

jryb, line 683: Should we put "/mnt/app-listing" into a constant string?

pkg/splunk/enterprise/util.go
```diff
@@ -211,6 +211,85 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterprisev1.VolumeSpec, client spl
...
+	var err error
+	var crKind string
+	var configMapDataChanged bool
+	crKind = cr.GetObjectKind().GroupVersionKind().Kind
+
+	scopedLog := log.WithName("ApplyAppListingConfigMap").WithValues("kind", crKind, "name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	mapAppListing := make(map[string]string)
+
+	// ToDo: sgontla/Jeff: Adapt to any Ansible changes(if required) for differentiating local vs. cluster scope (?)
```
jryb, line 229: Yes this will need to be addressed.  Currently if we put both the cluster and the local in both, for CM/Deployer all apps regardless of scope will have be pushed to the cluster and disabled locally. 

Sidenote: do we block use of `scope: cluster` for non clustered Types (anything other than SHC and CM)?

pkg/splunk/enterprise/util.go
```diff
@@ -211,6 +211,85 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterprisev1.VolumeSpec, client spl
...
+
+	for appSrc, appSrcDeployInfo := range appsSrcDeployStatus {
+		appDeployList := appSrcDeployInfo.AppDeploymentInfoList
+
+		switch scope := getAppSrcScope(appConf, appSrc); scope {
+		case "local":
+			for idx := range appDeployList {
+				if appDeployList[idx].DeployStatus == enterprisev1.DeployStatusPending {
+					localAppsConf = fmt.Sprintf(`%s
+    - "/init-apps/%s"`, localAppsConf, appDeployList[idx].AppName)
```
jryb, line 245: Does this create the correct yml?  Ideally the yml for each scope should look like this:
```
splunk:
  apps_location:
    - "/init-apps/appSrc1/app1.tgz"
    - "/init-apps/appSrc1/app2.tgz"
    - "/init-apps/appSrc2/app3.tgz"
```
Did this format of string work for ansible to install the apps?

pkg/splunk/enterprise/configuration.go
```diff
@@ -785,6 +846,33 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
...
+			splclient.RegisterS3Client(vol.Provider)
+		}
+	}
+}
+
+// getAppSrcScope returns the scope of a given appSource
+func getAppSrcScope(appFrameworkConf *enterprisev1.AppFrameworkSpec, appSrcName string) string {
+	for _, appSrc := range appFrameworkConf.AppSources {
+		if appSrc.Name == appSrcName {
+			if appSrc.Scope != "" {
```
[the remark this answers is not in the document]
jryb: Thats fine.  We don't need additional checks if we know that it's either local or cluster already.

pkg/splunk/enterprise/util.go
jryb: I believe you, was just making sure :)

sgontla: I have to correct my last statement. My logic assumes we write to "/init-apps/" as a flat directory. addressing the change to append the app source name.

pkg/splunk/enterprise/util.go
```diff
@@ -211,6 +217,85 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterprisev1.VolumeSpec, client spl
...
+				}
+			}
+
+		case "cluster":
+			for idx := range appDeployList {
+				if appDeployList[idx].DeployStatus == enterprisev1.DeployStatusPending {
+					clusterAppsConf = fmt.Sprintf(`%s
+    - "/init-apps/%s/%s"`, clusterAppsConf, appSrc, appDeployList[idx].AppName)
+				}
+			}
```
gaurav-splunk, line 261: From the initial look, it seems like the content of both `localAppsConf` and `clusterAppsConf` is of a ### similar style and has nothing to do with scope, right? All they contain is something like - 
``` 
splunk:
  apps_location:
      - "/init-apps/admin_app/admin_app.tgz"
```
If this is the case, then probably make the above thing in a function and pass clusterAppsConf/localAppsConf as a arg to that function?

pkg/splunk/enterprise/configuration.go
```diff
@@ -472,15 +474,32 @@ func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaOb
 	return statefulSet, nil
 }
 
+// getAppListingConfigMap returns the App listing configMap, if it exists and applicable for that instanceType
+func getAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject, instanceType InstanceType) (*corev1.ConfigMap, bool) {
+
+	// ToDo: sgontla: Exclude MC, once it's own CR is available
+	if instanceType != SplunkIndexer && instanceType != SplunkSearchHead && instanceType != SplunkMonitoringConsole {
```
gaurav-splunk, line 481: Why is it only applicable for CM and Deployer?

pkg/splunk/enterprise/configuration.go
```diff
@@ -584,7 +603,15 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		if instanceType == SplunkStandalone {
 			podTemplateSpec.ObjectMeta.Annotations[smartStoreConfigRev] = smartstoreConfigMap.ResourceVersion
 		}
+	}
+
+	appListingConfigMap, needToMountAppListingCm := getAppListingConfigMap(client, cr, instanceType)
```
gaurav-splunk, line 608: Do we actually need `needToMountAppListingCm` or just checking if `appListingConfigMap!=nil` should do the trick?

pkg/splunk/enterprise/util.go
```diff
@@ -68,6 +68,12 @@ func GetRemoteStorageClient(client splcommon.ControllerClient, cr splcommon.Meta
 
 	//Get the prefix from the "path" field
 	basePrefix := strings.TrimPrefix(vol.Path, bucket+"/")
+	// if vol.Path just containes bucket name(i.e without ending "/"), TrimPrefix returns the vol.Path
```
gaurav-splunk, line 71: nit: contains

pkg/splunk/enterprise/configuration.go
```diff
@@ -642,6 +669,20 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 	if spec.Defaults != "" {
 		splunkDefaults = fmt.Sprintf("%s,%s", "/mnt/splunk-defaults/default.yml", splunkDefaults)
 	}
+	if needToMountAppListingCm {
```
gaurav-splunk, line 672: ditto

pkg/splunk/enterprise/configuration.go
sgontla: Applicable to any pod other than the one we are checking here.

pkg/splunk/enterprise/configuration.go
sgontla: Both work. getAppListingConfigMap internally returns both the values, and we are using one of that :-)

pkg/splunk/enterprise/configuration.go
sgontla: same as above.

pkg/splunk/enterprise/util.go
sgontla: Ya, we can do that. Will address this as part of addressing Jeff's ansible changes.

jryb: This is a current limitation of the ansible app installation process.  The `apps_location` parameter installs apps locally on non-cluster instances (Standalone/LM) and to the cluster for CM/Deployer.  Once changes are in ansible to add the local scope to CM/Deployer, we'll have to change the `scope:local` case for CM/Deployer here to use the new local install option.

pkg/splunk/enterprise/configuration.go
```diff
@@ -233,17 +235,17 @@ func getSplunkDefaults(identifier, namespace string, instanceType InstanceType,
 	}
 }
 
-// prepareSplunkSmartstoreConfigMap returns a K8 ConfigMap containing Splunk smartstore config in INI format
-func prepareSplunkSmartstoreConfigMap(identifier, namespace string, crKind string, dataIniMap map[string]string) *corev1.ConfigMap {
-	configMapIni := &corev1.ConfigMap{
+// prepareConfigMap prepares and returns a K8 ConfigMap object for the given data
+func prepareConfigMap(configMapName, namespace string, dataMap map[string]string) *corev1.ConfigMap {
```
kashok-splunk, line 239: can we move this to configmap.go? and something with CreateConfigMap()?

pkg/splunk/enterprise/configuration.go
```diff
@@ -472,15 +474,32 @@ func getSplunkStatefulSet(client splcommon.ControllerClient, cr splcommon.MetaOb
...
+		return getConfigMap(client, cr, smartStoreConfigMapName)
 	}
 
-	if smartStoreConfigMapName != "" {
-		namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: smartStoreConfigMapName}
+	return nil, false
+}
+
+// getConfigMap returns configMap for a given name
+func getConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject, configMapName string) (*corev1.ConfigMap, bool) {
```
kashok-splunk, line 500: This can be confused with GetConfigMap() API. What we are doing here is retrieving the configmap for the specific instanceType and checking if that needs to be mounted? OR we need to use the configmap Revision as the trigger?

pkg/splunk/enterprise/configuration.go
```diff
@@ -785,6 +846,33 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
 	return false
 }
 
+func initAppFrameWorkContext(appFrameworkConf *enterprisev1.AppFrameworkSpec, appStatusContext *enterprisev1.AppDeploymentContext) {
```
kashok-splunk, line 849: Add function definition line

pkg/splunk/enterprise/names.go
```diff
@@ -133,6 +145,11 @@ func GetSplunkSmartstoreConfigMapName(identifier string, crKind string) string {
 	return fmt.Sprintf(smartstoreTemplateStr, identifier, strings.ToLower(crKind))
 }
 
+// GetSplunkAppsConfigMapName uses a template to name a Kubernetes ConfigMap for a SplunkEnterprise resource.
+func GetSplunkAppsConfigMapName(identifier string, crKind string) string {
```
kashok-splunk, line 149: nit: can we add UT for this?

pkg/splunk/enterprise/util.go
```diff
@@ -211,6 +217,85 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterprisev1.VolumeSpec, client spl
...
+	SplunkAppListingConfigMap.SetOwnerReferences(append(SplunkAppListingConfigMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
+	configMapDataChanged, err = splctrl.ApplyConfigMap(client, SplunkAppListingConfigMap)
+	if err != nil {
+		return nil, configMapDataChanged, err
+	} else if configMapDataChanged {
+		// Create a token to check if the app list is on the pod is latest or not
+		mapAppListing[appsUpdateToken] = fmt.Sprintf(`%d`, time.Now().Unix())
+
+		// Apply the configMap with a fresh token, if there is a change
+		SplunkAppListingConfigMap = prepareConfigMap(configMapName, cr.GetNamespace(), mapAppListing)
```
kashok-splunk, line 289: Correct me please, so on data change in configmap we are preparing a new configmap? and then Applying it? 
We can just use ApplyConfigMap() to pick new changes right?

sgontla: Good catch. Ya, that should work. we don't need to call the prepare twice. Will remove it.

pkg/splunk/enterprise/configuration.go
sgontla: It started as fat logic, but now, its almost the same as GetConfgMap. I will get rid of it completely.

pkg/splunk/enterprise/names.go
sgontla: All it is doing is a printf, and it has enough coverage at higher level flows. Let me know if you still prefer a UT

kashok-splunk: I agree, just we have UT for all such APIs, to keep this consistent with them we can have or we can take it later also :)

pkg/splunk/enterprise/configuration.go
gaurav-splunk: Yeah that was the question, why not applicable to other pods? :)

sgontla: AS you see in the above conditional check If the pod type is Indexer/SH, the configuration is applied through the CM/Deployer, so, there is nothing that we need to propagate in those cases. Once, we have a separate CR for the Deployer, ideally, we don't even need this check. Anyhow, it is a good one-time check before finalizing the pod template.

gaurav-splunk (approved):

kashok-splunk (approved):

smohan-splunk (approved):

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -83,16 +83,13 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 		}
 	}()
 
-	//check if the apps need to be downloaded from remote storage
+	initAppFrameWorkContext(&cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
+
+	// check if the apps need to be downloaded from remote storage
+	// ToDo: sgontla: Once after we have the flow, move the following logic into a seperate function(Note:- all errors should not cause a return in that function)
 	if cr.Spec.CommonSplunkSpec.Mock != true && !reflect.DeepEqual(cr.Status.AppContext.AppFrameworkConfig, cr.Spec.AppFrameworkConfig) {
```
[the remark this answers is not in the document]
sgontla: I think we need this clean-up. I Will add a ToDo note for now.

jryb (approved):

PR #367, opened 2021-06-08
CSPL-796: Periodically poll S3 buckets for updates

gaurav-splunk:
1. Periodically poll S3 buckets for updates based on AppRepoPollInterval parameter.
2. Added `MaxRetries=3` to AWS client config.
3. Added some files and UTs to improve coverage.


pkg/splunk/enterprise/standalone.go
```diff
@@ -158,7 +167,13 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
 		if err != nil {
 			return result, err
 		}
-		result.Requeue = false
+
+		// requeue the reconcile after polling interval if we had set the flag before
+		if cr.Status.AppContext.AppInfoStatus.NeedToCheckAfterPollInterval == true {
```
jryb, line 172: Do we need this `NeedToCheckAfterPollInterval` in the `AppInfoStatus` or can this be a local variable?

pkg/splunk/enterprise/searchheadcluster_test.go
```diff
@@ -666,8 +666,8 @@ func TestAppFrameworkSearchHeadClusterShouldFail(t *testing.T) {
 	client.AddObject(&s3Secret)
 
 	_, err = ApplySearchHeadCluster(client, &cr)
-	if err == nil {
-		t.Errorf("ApplySearchHeadCluster should not be successful")
+	if err != nil {
```
jryb, line 669: So what made this test turn from a negative testcase to a positive?

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -477,12 +477,9 @@ func TestAppFrameworkApplyClusterMasterShouldFail(t *testing.T) {
 		t.Errorf(err.Error())
 	}
 
-	// ApplyClusterMaster below should return error as we are not using mock AWS client here
-	// and hence it will call the AWS SDK library to fetch the bucket content which will obviously
-	// return error.
 	_, err = ApplyClusterMaster(client, &cm)
-	if err == nil {
-		t.Errorf("ApplyClusterMaster should have returned error here.")
+	if err != nil {
```
jryb, line 481: Same question here.

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -272,6 +271,18 @@ type AppSrcDeployInfo struct {
 	AppDeploymentInfoList []AppDeploymentInfo `json:"appDeploymentInfo,omitempty"`
 }
 
+// AppInfoStatus indicates if bundle push required
+type AppInfoStatus struct {
+	// This is false for the first time when we enter the reconcile loop.
+	// For all other consequent times, this will be set to true.
```
jryb, line 277: Maybe: `For all subsequent times,`

pkg/apis/enterprise/v1/common_types.go
```diff
@@ -272,6 +271,18 @@ type AppSrcDeployInfo struct {
...
+	// This will help in checking for the apps status either right away(for the first time case)
+	// or after the polling interval.
```
jryb, line 279: Does this mean its used for determining whether the app list get is cause by the initial kubectl apply of the CR or the app interval timer popping?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2510,6 +2510,23 @@ spec:
...
+                        format: int64
+                        type: integer
+                      needToCheckAfterPollInterval:
+                        description: This is false for the first time when we enter
+                          the reconcile loop. For all other consequent times, this
+                          will be set to true. This will help in checking for the
+                          apps status either right away(for the first time case) or
+                          after the polling interval.
+                        type: boolean
+                    type: object
```
sgontla, line 2529: I think we don't need this, as we we are tracking in the above 'lastAppInfoCheckTime'

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -83,33 +83,44 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
-			return result, err
+		// Only handle the app repo changes if we were able to successfully get the apps list
+		if shouldHandleAppChanges == true {
+			err = handleAppRepoChanges(client, cr, &cr.Status.AppContext, sourceToAppsList, &cr.Spec.AppFrameworkConfig)
+			if err != nil {
+				scopedLog.Error(err, "Unable to use the App list retrieved from the remote storage")
+				return result, err
+			}
+
+			cr.Status.AppContext.AppFrameworkConfig = cr.Spec.AppFrameworkConfig
```
sgontla, line 119: We can move it to the above scope block when the expected and received count matches for the map entries.??

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -174,7 +185,12 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 		}
 
 		if cr.Status.BundlePushTracker.NeedToPushMasterApps == false {
-			result.Requeue = false
+			// requeue the reconcile after polling interval if we had set the flag before
+			if cr.Status.AppContext.AppInfoStatus.NeedToCheckAfterPollInterval == true {
+				result.RequeueAfter = time.Second * time.Duration(cr.Spec.AppFrameworkConfig.AppsRepoPollInterval)
+			} else {
+				result.Requeue = false
+			}
```
sgontla, line 193: What happens, if there is a reconcile event for some other reason, before the expiry? I think, in that case, we are pre-maturely resetting it? I think we need additional intelligence to pick the exact time for the next to reconcile events to accommodate any reconcile events for other non-app framewokr reasons.

pkg/splunk/enterprise/util.go
```diff
@@ -789,3 +790,38 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+// ShouldCheckAppStatus checks whether we need to check the app status in the current reconcile
+func ShouldCheckAppStatus(appFrameworkRef enterprisev1.AppFrameworkSpec, appInfoStatus enterprisev1.AppInfoStatus) bool {
+	scopedLog := log.WithName("ShouldCheckAppStatus")
+
+	scopedLog.Info("Checking to see if we need to probe the status of apps on remote storage")
+
+	// check if its the first time or we are coming after appsRepoPollInterval secs.
+	return appInfoStatus.NeedToCheckAfterPollInterval == false ||
+		(appInfoStatus.NeedToCheckAfterPollInterval == true && hasAppRepoCheckTimerExpired(appFrameworkRef, appInfoStatus))
+}
```
sgontla, line 818: As mentioned earlier, just returning hasAppRepoCheckTimerExpired() will be simple and clear, as we are not differentiating the error condition for an 'apply' command vs. timer expiry.(?)

pkg/splunk/enterprise/searchheadcluster_test.go
gaurav-splunk: The fact that we will not exit from the reconcile if `GetAppListFromS3Bucket` returns error and just continue with other reconcile work will make this test pass now.

pkg/apis/enterprise/v1/common_types.go
gaurav-splunk: Yeah you are right. But as I think about it and also as Subba has mentioned, we probably won't need this variable now.

pkg/splunk/enterprise/clustermaster.go
gaurav-splunk: Sorry did not understand this. The above if condition is for when the expected and received count does not match.

pkg/splunk/client/minioclient.go
```diff
@@ -47,8 +47,7 @@ func NewMinioClient(bucketName string, accessKeyID string, secretAccessKey strin
 	var s3SplunkClient SplunkMinioClient
 	var err error
 
-	region := GetRegion(endpoint)
-	cl := fn(region, accessKeyID, secretAccessKey)
+	cl := fn(endpoint, accessKeyID, secretAccessKey)
```
gaurav-splunk, line 50: @jryb this is something which I noticed. Was it working before? And I hope this change is right now?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -83,36 +83,14 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
-		for _, appSource := range cr.Spec.AppFrameworkConfig.AppSources {
-			scopedLog.Info("Apps List retrieved from remote storage", "App Source", appSource.Name, "Content", sourceToAppsList[appSource.Name].Objects)
-		}
-
-		err = handleAppRepoChanges(client, cr, &cr.Status.AppContext, sourceToAppsList, &cr.Spec.AppFrameworkConfig)
+	// If the app framework is configured then do following things -
+	// 1. Initialize the S3Clients based on providers
+	// 2. Check the status of apps on remote storage.
+	if len(cr.Spec.AppFrameworkConfig.AppSources) != 0 {
+		err := initAndCheckAppInfoStatus(client, cr, &cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
```
gaurav-splunk, line 90: This should only be called if the appFramework is configured.

pkg/splunk/client/minioclient.go
jryb: It was working before.  Maybe get region was failing and just returning `endpoint`.  I'll verify minio once this is merged as part of the minio ut work.

jryb (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,85 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+	appInfoStatus.LastAppInfoCheckTime = currentEpoch
+}
+
+// HasAppRepoCheckTimerExpired checks if the polling interval has expired
+func HasAppRepoCheckTimerExpired(appFrameworkRef *enterprisev1.AppFrameworkSpec, appInfoContext *enterprisev1.AppDeploymentContext) bool {
+	scopedLog := log.WithName("hasAppRepoCheckTimerExpired")
+	currentEpoch := time.Now().Unix()
+	scopedLog.Info("Checking if the app repo polling interval has expired", "LastAppInfoCheckTime", strconv.FormatInt(appInfoContext.LastAppInfoCheckTime, 10), "current epoch time", strconv.FormatInt(currentEpoch, 10))
+	return appInfoContext.LastAppInfoCheckTime+int64(appFrameworkRef.AppsRepoPollInterval) <= currentEpoch
+}
```
sgontla, line 886: Nit: To avoid excessive logging across the reconciles, we can just log when it is true

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,85 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+	currentEpoch := time.Now().Unix()
+
+	scopedLog.Info("Setting the LastAppInfoCheckTime to current time", "current epoch time", currentEpoch)
+
+	appInfoStatus.LastAppInfoCheckTime = currentEpoch
+}
+
+// HasAppRepoCheckTimerExpired checks if the polling interval has expired
+func HasAppRepoCheckTimerExpired(appFrameworkRef *enterprisev1.AppFrameworkSpec, appInfoContext *enterprisev1.AppDeploymentContext) bool {
+	scopedLog := log.WithName("hasAppRepoCheckTimerExpired")
```
sgontla, line 882: Nit: "HasAppRepoCheckTimerExpired"

jryb (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,85 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+	//check if the apps need to be downloaded from remote storage
+	if HasAppRepoCheckTimerExpired(appFrameworkConf, appStatusContext) || !reflect.DeepEqual(appStatusContext.AppFrameworkConfig, *appFrameworkConf) {
+		var sourceToAppsList map[string]splclient.S3Response
+
+		scopedLog.Info("Checking status of apps on remote storage...")
+
+		sourceToAppsList, err = GetAppListFromS3Bucket(client, cr, appFrameworkConf)
+		if len(sourceToAppsList) != len(appFrameworkConf.AppSources) {
+			scopedLog.Error(err, "Unable to get apps list, will retry in next reconcile...")
+		} else {
```
sgontla, line 924: I think this is something that we need to strengthen how we want to handle, particularly for phase 3. With the current state machine logic, we mark any missing appSrc as deleted in the operator state. The only downsize for phase 2 is, once that appSrc is up next time,  we reset the Pod, and try to install the app again. If this behavior is fine, we can go ahead with a ToDo item for phase-3.

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,85 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+
+	appInfoStatus.LastAppInfoCheckTime = currentEpoch
+}
+
+// HasAppRepoCheckTimerExpired checks if the polling interval has expired
+func HasAppRepoCheckTimerExpired(appFrameworkRef *enterprisev1.AppFrameworkSpec, appInfoContext *enterprisev1.AppDeploymentContext) bool {
+	scopedLog := log.WithName("hasAppRepoCheckTimerExpired")
+	currentEpoch := time.Now().Unix()
+	scopedLog.Info("Checking if the app repo polling interval has expired", "LastAppInfoCheckTime", strconv.FormatInt(appInfoContext.LastAppInfoCheckTime, 10), "current epoch time", strconv.FormatInt(currentEpoch, 10))
+	return appInfoContext.LastAppInfoCheckTime+int64(appFrameworkRef.AppsRepoPollInterval) <= currentEpoch
```
sgontla, line 885: nit: AppsRepoPollInterval  is already int64

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,85 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+
+	var err error
+	// Register the S3 clients specific to providers if not done already
+	// This is done to prevent the null pointer dereference in case when
+	// operator crashes and comes back up and the status of app context was updated
+	// to match the spec in the previous run.
+	initAppFrameWorkContext(appFrameworkConf, appStatusContext)
+
+	//check if the apps need to be downloaded from remote storage
+	if HasAppRepoCheckTimerExpired(appFrameworkConf, appStatusContext) || !reflect.DeepEqual(appStatusContext.AppFrameworkConfig, *appFrameworkConf) {
```
sgontla, line 916: I think the timer should cover the boundary cases. Ideally, the polling time should be covering from the status context?

gaurav-splunk: The check for reflect.DeepEqual takes care of the case where if someone modifies the CR and applies it before the timer expires, then we should check the status immediately. And about the polling time, I think it doesn't matter where you pick it up from. On a personal front, I think its better to check it from the CR spec than from the context since there might be some delay in writing the status to the etcd.

pkg/splunk/enterprise/util.go
gaurav-splunk: Yeah probably fine for now.

pkg/splunk/enterprise/util.go
sgontla: My comment was about the boundary conditions, say, somebody configures like 1 second, OR 100 days, in which, we are using default min/max :-)

pkg/splunk/enterprise/util.go
sgontla: Ok, probably, we can have a ToDo comment for now? Alternatively, a separate task is a better idea? as we have to do some amount of code changes for handling this case?

gaurav-splunk: Ok I will add a TODO comment for now, we can take it up later in phase 3

pkg/splunk/enterprise/util.go
```diff
@@ -240,22 +244,31 @@ func ApplyAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.Me
...
+	sort.Strings(mapKeys)
+
+	for _, appSrc := range mapKeys {
+		appDeployList := appsSrcDeployStatus[appSrc].AppDeploymentInfoList
 
 		switch scope := getAppSrcScope(appConf, appSrc); scope {
 		case "local":
 			for idx := range appDeployList {
-				if appDeployList[idx].DeployStatus == enterprisev1.DeployStatusPending {
+				if appDeployList[idx].DeployStatus == enterprisev1.DeployStatusPending &&
```
smohan-splunk, line 261: When does DeployStatus be set to Pending. What happens to configMap (& subsequent steps) contents if DeployStatus is not Pending?

smohan-splunk (approved):

pkg/splunk/enterprise/util.go
gaurav-splunk: I'll let @sgontla answer that :)

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2615,10 +2617,22 @@ spec:
                       type: object
                     description: Represents the Apps deployment status
                     type: object
+                  appsRepoStatusPollIntervalSeconds:
+                    description: Interval in seconds to check the Remote Storage for
```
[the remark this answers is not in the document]
gaurav-splunk: We can, good to add actually. I will add it

pkg/splunk/enterprise/util.go
```diff
@@ -882,3 +866,93 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+		var sourceToAppsList map[string]splclient.S3Response
+
+		scopedLog.Info("Checking status of apps on remote storage...")
+
+		sourceToAppsList, err = GetAppListFromS3Bucket(client, cr, appFrameworkConf)
+		// TODO: gaurav, we need to handle this case better in Phase-3. There can be a possibility
+		// where if an appSource is missing in remote store, we mark it for deletion. But if it comes up
+		// next time, we will recycle the pod to install the app. We need to find a way to reduce the pod recycles.
+		if len(sourceToAppsList) != len(appFrameworkConf.AppSources) {
+			scopedLog.Error(err, "Unable to get apps list, will retry in next reconcile...")
```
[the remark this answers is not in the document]
gaurav-splunk: Yes right.

sgontla (approved):

pkg/splunk/enterprise/util.go
smohan-splunk: Had a chat with @sgontla 

 - `DeployStatus` will be set to Pending when there are App updates in `handleAppRepoChanges` (prior to call to `ApplyAppListingConfigMap`
 - If there are no updates to apps, configMap will have `app-list-local.yaml` & `app-list-cluster.yaml` maps with only headers (& empty data) which will avoid any unnecessary resets

PR #388, opened 2021-06-22
CSPL-1053 Removal of Automated MC on splunk operator upgrade

kashok-splunk:
After upgrade from Old Operator to new, no MC pod should be running and no references of old MC pod (such as StatefulSet) should exist
1. Check if the MC statefulset exist
2. Retrive all the owners of the MC statefulset
3. if Multiple owners of the MC statefulset then delete the MC statefulset
4. else if "this" CR is the owner of the MC statefulset then delete the MC sts
5. Delete the corresponding MC configmap


sgontla:
@kashok-splunk , I guess, with this approach, we lose all the existing knowledge objects, as we lose the volume claims? For the existing customers, I feel the better approach may be, on every existing CR(other than the new CR for the MC)
1. Make sure that there is a CR with the name that can map to the existing statefulset for the MC
2. Just remove our owner-reference. (Doing #2 without doing #1 can cause the garbage collector to delete the stateful set, so, we need to make sure #1 is satisfying)

pkg/splunk/controller/statefulset.go
```diff
@@ -282,3 +282,37 @@ func GetStatefulSetByName(c splcommon.ControllerClient, namespacedName types.Nam
 
 	return &statefulset, nil
 }
+
+// VerfiyMCStatefulSetOwnerRef deletes the automated MC sts. This is when customer migrates from automated MC to MC CRD
+// Check if MC CR is not the owner of the MC statefulset then delete that Statefulset
+func VerfiyMCStatefulSetOwnerRef(client splcommon.ControllerClient, cr splcommon.MetaObject, namespacedName types.NamespacedName) error {
```
sgontla, line 288: Nit: Can we rename it to something that reflects we are deleting the old-style automatic MC pod
ex. deleteReferencesToAutomaticMcIfExists() ? or something like that.

sgontla (approved):

pkg/splunk/controller/statefulset.go
smohan-splunk: Nit: Typo

pkg/splunk/controller/statefulset.go
```diff
@@ -282,3 +282,37 @@ func GetStatefulSetByName(c splcommon.ControllerClient, namespacedName types.Nam
...
+		}
+	} else if len(currentOwnerRef) == 1 {
+		//check if current cr is the owner of the mc sts
+		if reflect.DeepEqual(currentOwnerRef[0], splcommon.AsOwner(cr, false)) {
+			err := splutil.DeleteResource(client, statefulset)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
gaurav-splunk, line 310: Probably another way of doing this might be like this - 
```
.....
currentOwners := len(currentOwnerRef)
if currentOwners > 1 ||  isCurrentCROwner(cr, currentOwners) {
		err := splutil.DeleteResource(client, statefulset)
		if err != nil {
			return err
		}
}
....
}

func isCurrentCROwner(cr splcommon.MetaObject, currentOwners int) bool {
 return currentOwners == 1 && reflect.DeepEqual(currentOwnerRef[0], splcommon.AsOwner(cr, false))
}

File: pkg/splunk/controller/statefulset.go [hunk 3]
```diff
@@ -282,3 +282,37 @@ func GetStatefulSetByName(c splcommon.ControllerClient, namespacedName types.Nam
...
+
+// VerfiyMCStatefulSetOwnerRef deletes the automated MC sts. This is when customer migrates from automated MC to MC CRD
+// Check if MC CR is not the owner of the MC statefulset then delete that Statefulset
+func VerfiyMCStatefulSetOwnerRef(client splcommon.ControllerClient, cr splcommon.MetaObject, namespacedName types.NamespacedName) error {
+	statefulset, err := GetStatefulSetByName(client, namespacedName)
+	if err != nil {
+		// if MC Sts doesn't exist return nil, may have been deleted by other CR
+		return nil
+	}
+	//2. Retrive all the owners of the MC statefulset
```
gaurav-splunk, line 294: nit: Retrieve

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -127,6 +128,12 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
 
 	// no need to requeue if everything is ready
 	if cr.Status.Phase == splcommon.PhaseReady {
+		//upgrade fron automated MC to MC CRD
+		namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: GetSplunkStatefulsetName(SplunkMonitoringConsole, cr.GetNamespace())}
```
smohan-splunk, line 132: Can the namespace be extracted within `splctrl.VerfiyMCStatefulSetOwnerRef` itself? Since. are anyway passing the cr as an argument..

Same comment for other cases too

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -127,6 +128,12 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterprisev1.Clus
...
+		err = splctrl.VerfiyMCStatefulSetOwnerRef(client, cr, namespacedName)
+		if err != nil {
+			return result, err
```
smohan-splunk, line 135: Should we really return an error here? May be just throw a warning & move on with other reconcile activities?

Same comment for other cases too

pkg/splunk/controller/statefulset.go
```diff
@@ -282,3 +283,34 @@ func GetStatefulSetByName(c splcommon.ControllerClient, namespacedName types.Nam
...
+// Check if MC CR is not the owner of the MC statefulset then delete that Statefulset
+func DeleteReferencesToAutomatedMCIfExists(client splcommon.ControllerClient, cr splcommon.MetaObject, namespacedName types.NamespacedName) error {
+	statefulset, err := GetStatefulSetByName(client, namespacedName)
+	if err != nil {
+		// if MC Sts doesn't exist return nil, may have been deleted by other CR
+		return nil
+	}
+	//2. Retrieve all the owners of the MC statefulset
+	currentOwnersRef := statefulset.GetOwnerReferences()
+	//3. if Multiple owners OR id current CR is the owner of the MC statefulset then delete the MC statefulset
```
gaurav-splunk, line 297: nit: if current CR is the...

gaurav-splunk (approved):

pkg/splunk/enterprise/clustermaster.go
smohan-splunk: Discussed offline with @kashok-splunk . `GetSplunkStatefulsetName` can't be called in the controller package (where `DeleteReferencesToAutomatedMCIfExists` resides) due to a cyclic dependency, hence the namespace calculation needs to happen outside.

smohan-splunk (approved):

PR #391, opened 2021-06-25
App Framework POC for phase 3.5

jryb:
This code adds download and install capabilities when an app package
is determined to have changed, without the need for any restart.

App packages are downloaded by the operator then pushed to the
cooresponding pod.  Once successfuly they are either a) installed
for local scope, or b) extracted to the bundle directory for cluster
scope.  Once all apps are finished if any were extracted to a
bundle dir, then a bundle push is performed.  All splunk cmds are
handled via the operator and not ansible.

Ansible still handles the initial install of all apps.

### DIFF
--- a/pkg/splunk/client/awss3client.go
+++ b/pkg/splunk/client/awss3client.go
@@ -17,6 +17,7 @@ package client
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 
@@ -24,10 +25,12 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
 // blank assignment to verify that AWSS3Client implements S3Client
 var _ S3Client = &AWSS3Client{}
+var gSess *session.Session
 
 // SplunkAWSS3Client is an interface to AWS S3 client
 type SplunkAWSS3Client interface {
@@ -76,6 +79,9 @@ func InitAWSClientSession(region, accessKeyID, secretAccessKey string) SplunkAWS
 		scopedLog.Error(err, "Failed to initialize an AWS S3 session.")
 		return nil
 	}
+
+	// TODO: Fix this and store it in the SplunkAWSS3Client
+	gSess = sess
 
 	return s3.New(sess)
 }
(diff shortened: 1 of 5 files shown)


pkg/splunk/client/awss3client.go
```diff
@@ -77,6 +80,9 @@ func InitAWSClientSession(region, accessKeyID, secretAccessKey string) SplunkAWS
 		return nil
 	}
 
+	// TODO: Fix this and store it in the SplunkAWSS3Client
+	gSess = sess
```
jryb, line 84: @gaurav-splunk I may need you help with merging this into the correct interface since its different (`s3manager`) than the `s3` object used by the list APIs

pkg/splunk/client/minioclient.go
```diff
@@ -146,3 +149,37 @@ func (client *MinioClient) GetInitContainerImage() string {
 func (client *MinioClient) GetInitContainerCmd(endpoint string, bucket string, path string, appSrcName string, appMnt string) []string {
 	return ([]string{fmt.Sprintf("--endpoint-url=%s", endpoint), "s3", "sync", fmt.Sprintf("s3://%s/%s", bucket, path), fmt.Sprintf("%s/%s", appMnt, appSrcName)})
 }
+
+// DownloadFile downloads an app package from remote storage
+func (client *MinioClient) DownloadFile(remoteFile string, localFile string) error {
```
jryb, line 154: Still need to verify minio.

pkg/splunk/enterprise/configuration.go
```diff
@@ -581,7 +581,7 @@ func updateSplunkPodTemplateWithConfig(client splcommon.ControllerClient, podTem
 		// ToDo: sgontla: for Phase-2, to install the new apps, always reset the pod.(need to change the behavior for phase-3)
 		// Once the apps are installed, and on a reconcile entry triggered by polling interval expirty, if there is no new
 		// App changes on remote store, then the config map data is erased. In such case, no need to reset the Pod
-		if len(appListingConfigMap.Data) > 0 {
+		if len(appListingConfigMap.Data) > 0 && 1 == 0 /* TODO jryb: turn this off for now*/ {
```
jryb, line 584: Basically turns off the pod restart of app updates.

pkg/splunk/enterprise/util.go
```diff
@@ -681,14 +728,15 @@ func isAppExtentionValid(receivedKey string) bool {
 }
 
 // AddOrUpdateAppSrcDeploymentInfoList  modifies the App deployment status as perceived from the remote object listing
-func AddOrUpdateAppSrcDeploymentInfoList(appSrcDeploymentInfo *enterprisev1.AppSrcDeployInfo, remoteS3ObjList []*splclient.RemoteObject) bool {
+func AddOrUpdateAppSrcDeploymentInfoList(appSrcDeploymentInfo *enterprisev1.AppSrcDeployInfo, remoteS3ObjList []*splclient.RemoteObject, appFrameworkConfig *enterprisev1.AppFrameworkSpec, appSrc string, cr splcommon.MetaObject, client splcommon.ControllerClient) bool {
```
jryb, line 731: @sgontla I added these so that I could handle the download/copy/install inline in this function.  I know this isn't the right place for it so I many need so help in finding it a proper home.

pkg/splunk/enterprise/util.go
```diff
@@ -736,6 +786,119 @@ func AddOrUpdateAppSrcDeploymentInfoList(appSrcDeploymentInfo *enterprisev1.AppS
 			newAppInfoList = append(newAppInfoList, appDeployInfo)
 			appChangesDetected = true
 		}
+
+		// Did an app change?  Let's do something about it... vvvv-Maybe remove this check-vvvv or make it "if pod is ready"
+		if appChangesDetected == true && prevDeployStatus == enterprisev1.DeployStatusComplete {
```
jryb, line 791: This check isn't right either.  It will start trying to download/install apps prior to the pod/container being ready and simply fail.  Ideally we should wait until after the first run/initialization of splunk before we start running this code on appSrc changes.

pkg/splunk/enterprise/util.go
```diff
@@ -964,3 +1127,170 @@ func initAndCheckAppInfoStatus(client splcommon.ControllerClient, cr splcommon.M
 
 	return nil
 }
+
+//////////////////////////
+//
+// App Framework Action Functions
```
jryb, line 1133: This need a new home too to isolate these actions.  With this method too its fairly easy to see how we could implement a disable/delete as well.

PR #403, opened 2021-07-08
CSPL-1044 Full MC configuration/reconfiguration (All deployments other than IDC)

kashok-splunk:
Full MC configuration/reconfiguration (All deployments other than IDC)
Example to create Standalone with mcRef 
```
apiVersion: enterprise.splunk.com/v1
kind: Standalone
metadata:
  name: s2
  finalizers:
spec:
  monitoringConsoleRef:
    name: test
```

### DIFF
--- a/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
@@ -875,6 +875,43 @@ spec:
               licenseUrl:
                 description: Full path or URL for a Splunk Enterprise license file
                 type: string
+              monitoringConsoleRef:
+                description: MonitoringConsoleRef refers to a Splunk Enterprise monitoring
+                  console managed by the operator within Kubernetes
+                properties:
+                  apiVersion:
+                    description: API version of the referent.
+                    type: string
+                  fieldPath:
+                    description: 'If referring to a piece of an object instead of
+                      an entire object, this string should contain a valid JSON/Go
+                      field access statement, such as desiredState.manifest.containers[2].
+                      For example, if the object reference is to a container within
+                      a pod, this would take on a value like: "spec.containers{name}"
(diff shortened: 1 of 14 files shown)


smohan-splunk (approved):

deploy/crds/enterprise.splunk.com_indexerclusters_crd.yaml
```diff
@@ -875,6 +875,43 @@ spec:
...
+                    type: string
+                  resourceVersion:
+                    description: 'Specific resourceVersion to which this reference
+                      is made, if any. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency'
+                    type: string
+                  uid:
+                    description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
+                    type: string
+                type: object
               replicas:
```
sgontla, line 915: Do we need this for IDXC? What will be the net affect, if somebody configures this?

sgontla: Had a discussion with Kriti:
1. Good to have the MC ref as part of the MC CR, in case if we introduce the Global MC in the future. 
2. SHC (even if we have a separate CR in the future) needs the MC reference, as the logic expects both for the Deployer and SHs.
3. IDXC case, each site-specific CR having MC reference can lead to confusion, as different sites can't go to different MCs. The best way is if we can pull this info from the CM CR, by reading it from the API server(?). If that is not feasible, an alternative can be using the configMap(CM writes it, and the site-specific CRs read and use it).

With the above understanding, we can continue using it in the current form(having the MC reference), but the implementation can ignore when dealing with the IDXC, and use it from the CM CR or through a configMap.

sgontla (approved):

PR #410, opened 2021-07-12
CSPL-1034: Handle scaling up of standalone with app framework

gaurav-splunk:


pkg/splunk/enterprise/standalone.go
```diff
@@ -123,6 +123,35 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
 		return result, err
 	}
 
+	// If we are using appFramework and are scaling up, we should re-populate the
```
jryb, line 126: More of a comment on the code rasther than this PR itself: I think something similar needs to be done for CRs with app framework and ephemeral storage for etc.

jryb (approved):

pkg/splunk/enterprise/standalone.go
sgontla: ephemeral is an interesting use case. I believe we need to :
1. Error out if the App framework is configured with ephemeral storage
2. Also, need to highlight this in the docs.

pkg/splunk/enterprise/standalone.go
```diff
@@ -123,6 +123,35 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterprisev1.Standal
...
+	// configMap with all the appSource entries. This is done so that the new pods
+	// that come up now will have the complete list of all the apps and then can
+	// download and install all the apps.
+	// TODO: Improve this logic so that we only recycle the new pod/replica
+	// and not all the existing pods.
+	if len(cr.Spec.AppFrameworkConfig.AppSources) != 0 && cr.Spec.Replicas > 1 {
+
+		statefulsetName := GetSplunkStatefulsetName(SplunkStandalone, cr.GetName())
+
+		ok, err := splctrl.IsStatefulSetScalingUp(client, cr, statefulsetName, cr.Spec.Replicas)
```
sgontla, line 136: nit: instead of ok, you may want to use a more logical flag name for better clarity :-)

sgontla (approved):

pkg/splunk/enterprise/standalone.go
gaurav-splunk: Yeah probably we can fix the ephemeral part as a separate JIRA. Most probably it will have similar fix but I just want to keep things separate.

smohan-splunk (approved):

PR #416, opened 2021-07-13
cspl-1120: New CRDs with App framework

sgontla:
- Changed the CRD Api Version to v2
- added the release label 1.0.2
- Added a soft link to the latest API, so that the Enterprise code refers to the softlink, that way, no need to modify the Enterprise code for any future CRD version changes.
- Modified the IndexerCluster test case, to fetch the latest CRD API version, so that we don't need to modify every time.


deploy/olm-catalog/splunk/1.0.2/splunk.v1.0.2.clusterserviceversion.yaml
```diff
@@ -0,0 +1,349 @@
...
+  - ClusterMaster
+  - Deployer
+  links:
+  - name: Splunk
+    url: https://splunk.domain
+  maintainers:
+  - email: smohan@splunk.com
+    name: '''Sirish Mohan'
+  - email: sgontla@splunk.com'
+    name: Subba Gontla
```
sgontla, line 344: Didn't think that our names come here :-)
@smohan-splunk , just noticed, following is what is on develop branch. hope that is the same info we need to use here.

```  keywords:
  - Splunk
  - logging
  links:
  - name: Splunk
    url: https://splunk.domain
  maintainers:
  - email: docker-maint@splunk.com
    name: Splunk Docker Maintainers

File: build/make_bundle.sh [hunk 2]
```diff
@@ -5,13 +5,17 @@
 set -e
 
 VERSION=`grep "Version.*=.*\".*\"" version/version.go | sed "s,.*Version.*=.*\"\(.*\)\".*,\1,"`
-OLD_VERSIONS="v1beta1 v1alpha3 v1alpha2"
+OLD_VERSIONS="v1 v1beta1 v1alpha3 v1alpha2"
```
gaurav-splunk, line 8: how many old versions do we usually support?

build/make_bundle.sh
```diff
@@ -5,13 +5,17 @@
...
 DOCKER_IO_PATH="docker.io/splunk"
 REDHAT_REGISTRY_PATH="registry.connect.redhat.com/splunk"
 OPERATOR_IMAGE="$DOCKER_IO_PATH/splunk-operator:${VERSION}"
 OLM_CATALOG=deploy/olm-catalog
 OLM_CERTIFIED=deploy/olm-certified
 YAML_SCRIPT_FILE=.yq_script.yaml
+ENTERPRISE_API_LOC=pkg/apis/enterprise
+
+# create a soft link that point to the latest Splunk Enterprise CRDs
+ln -snf `ls -lr $ENTERPRISE_API_LOC| awk 'NF > 2 {print $NF}' | head -1` $ENTERPRISE_API_LOC/latest
```
gaurav-splunk, line 18: Are you sure this is the right commmand? Because when I run it in my local env. this is what I get - 
```$ ls -lr pkg/apis/enterprise/ | awk 'NF >2 {print$NF}'
v1beta1
v1alpha3
v1alpha2
v1
group.go
```

But if I run `ls -lrt` then I get this - 
```$ ls -lrt pkg/apis/enterprise/ | awk 'NF >2 {print$NF}'
group.go
v1alpha2
v1alpha3
v1beta1
v1
```
So probably better command could be 
`ls -lrt $ENTERPRISE_API_LOC| awk 'NF > 2 {print $NF}' | tail -1`  ?

gaurav-splunk:
> * Changed the CRD Api Version to v2
> * added the release label 1.0.2
> * Added a soft link to the latest API, so that the Enterprise code refers to the softlink, that way, no need to modify the Enterprise code for any future CRD version changes.
> * Modified the IndexerCluster test case, to fetch the latest CRD API version, so that we don't need to modify every time.

@sgontla I think we should also change versions in docs too

build/make_bundle.sh
sgontla: Assuming the versions in chronological order, say, v4, v5, v5.1, it should work. 

Anyhow, I am stashing it for now, as the Golang doesn't like referring the modules through the soft links. Will see how we can achieve it.

PR #422, opened 2021-07-16
Enabled EKS cluster for app fw test

pdhanoyasplunk:
Enabled EKS cluster for app fw test


smohan-splunk (approved):
This is an interim PR to confirm int-tests for app-framework execute successfully against the feature branch. A new PR will follow this one, which will effectively revert this PR and instead split the app-framework test cases into smoke & integration(nightly)

PR #428, opened 2021-07-20
Int test changes

pdhanoyasplunk:
Int test changes to disable failing validations and increase probe time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,7 @@ executors:
       # Set cluster workers to 5 (5 cluster nodes)
       # NUM_NODES represent number of parallel test executions
       # NUM_WORKERS represent number of nodes in a k8 cluster
-      NUM_NODES: 5 
+      NUM_NODES: 3 
       NUM_WORKERS: 8
       CLUSTER_NAME: eks-integration-test-cluster
       ENTERPRISE_IMAGE_NAME: splunk/splunk:edge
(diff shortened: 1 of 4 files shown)


smohan-splunk (approved):

tpereirasplunk (approved):

jambrosiano (approved):

test/testenv/deployment.go
```diff
@@ -669,7 +673,9 @@ func (d *Deployment) DeployMultisiteClusterWithSearchHeadAndAppFramework(name st
 			LicenseMasterRef: corev1.ObjectReference{
 				Name: licenseMaster,
 			},
-			Defaults: defaults,
+			Defaults:                     defaults,
+			LivenessInitialDelaySeconds:  int32(delaySeconds),
+			ReadinessInitialDelaySeconds: int32(delaySeconds),
```
sgontla, line 678: Nit: By default, the liveness probe is always 290 seconds higher than the readiness probe time. But, if this settings work fine for the test, that is fine as well :-)

.circleci/config.yml
```diff
@@ -106,7 +106,7 @@ executors:
       # Set cluster workers to 5 (5 cluster nodes)
       # NUM_NODES represent number of parallel test executions
       # NUM_WORKERS represent number of nodes in a k8 cluster
-      NUM_NODES: 5 
+      NUM_NODES: 3
```
sgontla, line 109: Nit: Is this done for a specific purpose?

sgontla (approved):

.circleci/config.yml
tpereirasplunk: We want to reduce the number of test cases that run in parallel.

PR #443, opened 2021-07-27
Commenting ES app test on C3 failing due to change introduced in CSPL-1167

jambrosiano:


pdhanoyasplunk (approved):

smohan-splunk (approved):
Test case will be enabled back as part of CSPL-1206 / CSPL-1207

tpereirasplunk (approved):

PR #451, opened 2021-08-03
[CSPL-1089] Update MC for idxc

kashok-splunk:
Update MC for idxc

### DIFF
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -183,7 +183,7 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	return result, nil
 }
 
-// clusterMasterPodManager is used to manage the cluster master pof
+// clusterMasterPodManager is used to manage the cluster master pod
 type clusterMasterPodManager struct {
 	c               splcommon.ControllerClient
 	log             logr.Logger
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/go-logr/logr"
 	enterpriseApi "github.com/splunk/splunk-operator/pkg/apis/enterprise/v2"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/splunk/enterprise/configuration.go
```diff
@@ -123,10 +123,6 @@ func getSplunkService(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunkS
 	service.ObjectMeta.Namespace = cr.GetNamespace()
 	instanceIdentifier := cr.GetName()
 	var partOfIdentifier string
-	if instanceType == SplunkMonitoringConsole {
```
smohan-splunk, line 126: Could you please clarify why this isn't required anymore?

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -136,6 +136,22 @@ func ApplyIndexerCluster(client splcommon.ControllerClient, cr *enterpriseApi.In
...
+		if cmMonitoringConsoleConfigRef != "" {
+			namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: GetSplunkStatefulsetName(SplunkMonitoringConsole, cmMonitoringConsoleConfigRef)}
+			_, err := splctrl.GetStatefulSetByName(client, namespacedName)
+			//if MC pod already exists
+			if err == nil {
+				c := mgr.getMonitoringConsoleClient(cr, cmMonitoringConsoleConfigRef)
+				err := c.AutomateMCApplyChanges(false)
+				return result, err
+			}
+			if cr.Spec.MonitoringConsoleRef.Name != "" || cr.Spec.MonitoringConsoleRef.Name != cmMonitoringConsoleConfigRef {
```
smohan-splunk, line 151: Do we really allow mcRef to be configured on Indexer cluster CR? Even if it matches one in CM, I am just worried it will cause confusion.. any thoughts?

smohan-splunk (approved):

pkg/splunk/enterprise/configuration.go
kashok-splunk: This is for creating the service name for the MC pod. I realized this change should have been done before only. This creates the service name of MC pod as splunk-<namespace>-monitoring-console but what we need is splunk-<crname>-monitoring-console so removed this

pkg/splunk/enterprise/indexercluster.go
kashok-splunk: This "monitoringConsoleRef" is a part of commonSplunkSpec it is available for indexer CR also but what we are doing here is we always check to which MC is CM connected to and attach Indexer with it but even if the customer provides monitoringConsoleRef in indexer CR that will be ignored. I believe to avoid this I will update the MC documentation for indexer CR. Same conclusion we had last time https://github.com/splunk/splunk-operator/pull/403#discussion_r667141568

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -144,6 +152,15 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 		if err != nil {
 			scopedLog.Error(err, "Error in deleting automated monitoring console resource")
 		}
+		//Update MC configmap
+		if cr.Spec.MonitoringConsoleRef.Name != "" {
```
jryb, line 156: Possibly a dumb question but do we need to do anything is an IDC CR removes a MC ref?  Are we handling that case?

kashok-splunk: if the mcref is changed/removed we have a story to cover that scenario CSPL-1087

jryb (approved):

PR #463, opened 2021-08-12
CSPL-1240: Fix the bug where during upgrade to latest CRD version, K8s server prunes the CRD fields

gaurav-splunk:
**Problem**
If we upgrade the operator from 1.0.1 to 1.0.2, the CRD version is also upgraded from v1 to v2. Now after the upgrade the `spec` section of the CRs(which were still on version v1) was pruned and pods either started crashing or they came up with no spec section at all.

This happened because of the default kubernetes behavior of pruning the unknown fields(fields that are not native to K8s server, such as the spec section of a CRD) of a CRD.

**Solution**
To fix this we had to add the field **x-kubernetes-preserve-unknown-fields: true** to older versions of the CRDs so that when  we upgrade to newer version/s, we still retain all the fields of the older versions.

### DIFF
--- a/deploy/olm-catalog/splunk/1.0.2/splunk.v1.0.2.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/splunk/1.0.2/splunk.v1.0.2.clusterserviceversion.yaml
@@ -258,7 +258,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: splunk-operator
                 - name: RELATED_IMAGE_SPLUNK_ENTERPRISE
-                  value: docker.io/splunk/splunk:8.2.1-a1
+                  value: docker.io/splunk/splunk:8.2.1-a2
                 image: docker.io/splunk/splunk-operator:1.0.2
                 imagePullPolicy: IfNotPresent
                 name: splunk-operator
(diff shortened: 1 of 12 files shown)


sgontla (approved):

deploy/olm-catalog/splunk/1.0.2/splunk.v1.0.2.clusterserviceversion.yaml
```diff
@@ -258,7 +258,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: splunk-operator
                 - name: RELATED_IMAGE_SPLUNK_ENTERPRISE
-                  value: docker.io/splunk/splunk:8.2.1-a1
+                  value: docker.io/splunk/splunk:8.2.1-a2
```
akondur, line 261: Nit: Was this change deliberate?

akondur (approved):
Looks good, great find @gaurav-splunk

jryb (approved):

deploy/olm-catalog/splunk/1.0.2/splunk.v1.0.2.clusterserviceversion.yaml
smohan-splunk: Yes, was missed last time around

smohan-splunk (approved):

PR #469, opened 2021-08-18
[CSPL-1087] Full re-configuration when mcRef is changed/removed

kashok-splunk:
- Once the MC CR is applied and has been configured with all the other instances whose CR has reference to the new MC, deleting the mcRef in the deployment's CR should make MC remove reference to that deployment
- Once the MC CR is applied and has been configured with all the other instances whose CR has reference to the new MC, changing the mcRef (from current MC to new) in the deployment's CR should make the current MC remove reference to that deployment, and new MC to configure the deployment

### DIFF
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -140,7 +140,7 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 
 	//make changes to respective mc configmap when changing/removing mcRef from spec
 	extraEnv, err := VerifyCMisMultisite(cr, namespaceScopedSecret)
-	err = validateMonitoringConosoleRef(client, statefulSet, extraEnv)
+	err = validateMonitoringConsoleRef(client, statefulSet, extraEnv)
 	if err != nil {
 		return result, err
 	}
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -141,6 +141,9 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	//make changes to respective mc configmap when changing/removing mcRef from spec
 	extraEnv, err := VerifyCMisMultisite(cr, namespaceScopedSecret)
 	err = validateMonitoringConosoleRef(client, statefulSet, extraEnv)
+	if err != nil {
+		return result, err
+	}
 
 	clusterMasterManager := splctrl.DefaultStatefulSetPodManager{}
 	phase, err := clusterMasterManager.Update(client, statefulSet, 1)
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -137,6 +137,11 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	if err != nil {
 		return result, err
 	}
+
+	//make changes to respective mc configmap when changing/removing mcRef from spec
+	extraEnv, err := VerifyCMisMultisite(cr, namespaceScopedSecret)
+	err = validateMonitoringConosoleRef(client, statefulSet, extraEnv)
+
 	clusterMasterManager := splctrl.DefaultStatefulSetPodManager{}
 	phase, err := clusterMasterManager.Update(client, statefulSet, 1)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/splunk/enterprise/licensemaster_test.go
```diff
@@ -38,6 +38,7 @@ func TestApplyLicenseMaster(t *testing.T) {
 		{MetaName: "*v1.Secret-test-splunk-stack1-license-master-secret-v1"},
 		{MetaName: "*v1.ConfigMap-test-splunk-stack1-licensemaster-app-list"},
 		{MetaName: "*v1.StatefulSet-test-splunk-stack1-license-master"},
+		{MetaName: "*v1.StatefulSet-test-splunk-stack1-license-master"},
```
smohan-splunk, line 41: Why is this func call added again? Same question for `pkg/splunk/enterprise/searchheadcluster_test.go` and `pkg/splunk/enterprise/standalone_test.go`

pkg/splunk/enterprise/util.go
```diff
@@ -999,3 +1002,54 @@ func initAndCheckAppInfoStatus(client splcommon.ControllerClient, cr splcommon.M
...
+			//1. if revised Spec has different mcRef defined
+			_, err = ApplyMonitoringConsoleEnvConfigMap(c, current.ObjectMeta.GetNamespace(), current.ObjectMeta.GetName(), cEnv.Value, serviceURLs, false)
+			if err != nil {
+				return err
+			}
+			_, err = ApplyMonitoringConsoleEnvConfigMap(c, current.ObjectMeta.GetNamespace(), current.ObjectMeta.GetName(), rEnv.Value, serviceURLs, true)
+			if err != nil {
+				return err
+			}
+		} else if cEnv.Value != "" && monitoringConsoleRevEmpty {
```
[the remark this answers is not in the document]
kashok-splunk: If we remove the mcRef from the spec the SPLUNK_MONITORING_CONSOLE_REF env variable itself is deleted from the Env variables list, so there is no way to compare their values. So added this extra boolean to verify that if the env variable itself exists or not

pkg/splunk/enterprise/util.go
```diff
@@ -999,3 +1002,54 @@ func initAndCheckAppInfoStatus(client splcommon.ControllerClient, cr splcommon.M
...
+
+		for _, rEnvTemp := range revEnv {
+			if rEnvTemp.Name == "SPLUNK_MONITORING_CONSOLE_REF" {
+				rEnv.Name = "SPLUNK_MONITORING_CONSOLE_REF"
+				rEnv.Value = rEnvTemp.Value
+				monitoringConsoleRevEmpty = false
+			}
+		}
+
+		if cEnv.Name == "SPLUNK_MONITORING_CONSOLE_REF" && rEnv.Name == "SPLUNK_MONITORING_CONSOLE_REF" && cEnv.Value != rEnv.Value {
```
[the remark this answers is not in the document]
kashok-splunk: Let me re-check I believe yes this can be handled in one more way, maybe we can avoid using boolean. Taking care of this in new commit

pkg/splunk/enterprise/licensemaster_test.go
kashok-splunk: validateMonitoringConsoleRef function does a GET call to pull the statefulset like below, so one more call added
```
var current appsv1.StatefulSet
err = c.Get(context.TODO(), namespacedName, &current)
```

smohan-splunk (approved):

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -137,6 +137,14 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
...
 		return result, err
 	}
+
+	//make changes to respective mc configmap when changing/removing mcRef from spec
+	extraEnv, err := VerifyCMisMultisite(cr, namespaceScopedSecret)
+	err = validateMonitoringConsoleRef(client, statefulSet, extraEnv)
+	if err != nil {
+		return result, err
+	}
+
```
sgontla, line 147: I think, moving the `validateMonitoringConsoleRef`(and also renaming it as `validateAndMonitoringConsoleRef`, then, also covering the case where the MC ref is added for the very first time) to the `ready phase checking` will reduce the redundant calling across the reconciliations.

sgontla (approved):

pkg/splunk/enterprise/configuration.go
```diff
@@ -741,17 +745,12 @@ func getLivenessProbe(cr splcommon.MetaObject, instanceType InstanceType, spec *
 
 	livenessDelay := int32(livenessProbeDefaultDelaySec)
 
-	// Exclude MC, as it derives the spec from Multiple CRs.
-	// ToDo: Remove the Check once the MC CRD is in place
-	if instanceType != SplunkMonitoringConsole {
```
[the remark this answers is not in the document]
kashok-splunk: Left that change intentionally, to include it when bringing in changes for app framework

pkg/splunk/enterprise/clustermaster.go
kashok-splunk: Discussed with @sgontla offline. When adding mcRef for the first time we need some information from the CM, and that should be in the ready phase, so we still need to perform that in that ready phase. For re-config, we need to update MC configmap before update() kicks in. So separating both the cases

PR #485, opened 2021-08-25
Removed integration keyword from appframework test to remove test fro…

pdhanoyasplunk:
Removed integration keyword from appframework test to remove test from running as part of integration test pipeline.


pdhanoyasplunk:
Not needed after Github Actions Migration.

PR #486, opened 2021-08-26
CSPL-1267: Merge develop to app framework

sgontla:
Merge develop to app framework


.github/workflows/int-test-workflow.yml
```diff
@@ -0,0 +1,122 @@
+name: Integration Test WorkFlow
+on:
+  push:
+    branches:
+      - develop
+      - master
+      - feature-appframework
```
smohan-splunk, line 7: Just a note: Added for app framework .. preferable to remove while merging back to develop

sgontla: Yes Sirish. I think @pdhanoyasplunk  was exploring if we can add a regex pattern like `feature*`, which covers all the feature branches.

pdhanoyasplunk: @smohan-splunk  We can add a pattern `'feature**` to run against any branch that runs with `feature` keyword.

smohan-splunk (approved):

PR #489, opened 2021-08-27
CSPL-1271: Fix a bug where standalone with replicas>1 won't come up

gaurav-splunk:


smohan-splunk (approved):

jambrosiano:
Got the pods replicas correctly deployed when using code from this PR, test details added in CSPL-1271

PR #490, opened 2021-08-27
Feature circleci migration

pdhanoyasplunk:
Introduced Github Action Workflows to replace CircleCI


.github/workflows/int-test-workflow.yml
```diff
@@ -0,0 +1,150 @@
+name: Integration Test WorkFlow
+on:
+  schedule:
+    - cron: "0 12 * * *"
```
sgontla, line 4: nit: Does it make sense to run at the end of the day, instead of at the middle of the day, so that it covers all the merges for that day?

sgontla (approved):

smohan-splunk (approved):

PR #494, opened 2021-09-01
Modify monitoring console selection name

pdhanoyasplunk:
Modified Monitoring Console test selection keyword to avoid cluster creation error on eks.


smohan-splunk:
As explained by Param,

`There is a regex on Amazon which dictates cluster name, have _ in monitoring_console caused the failure`

smohan-splunk (approved):

PR #503, opened 2021-09-08
CSPL-1301: Trigger app install for modified app pkgs

jryb:
### Problem
App packages that are modified in a tracked `appRepo` are identified by the Splunk Operator, however this does not always trigger a change in the configmap.  If the configmap data section does not change, then this modified app is never installed on the Splunk instance.

### Solution
While an app package modification does trigger rewriting the configmap.  There are cases where this rewrite will result in the same data section as the previous version of the configmap (for example, when a single app is installed initially then modified.)  When this is applied no change is detected and the Pod does not reset or install the new app.

By resetting the data in the app listing configmap when an app package change is detected, we can force a new resourceVersion of the configmap which will trigger the app install.  The new ResourceVersion of the configmap will restart the pod, triggering the modified app install.

Also fix the logging in `splunk.enterprise.AddOrUpdateAppSrcDeploymentInfoList` so it no longer throws errors:

```
{"level":"info","ts":1631123552.1276112,"logger":"splunk.enterprise.AddOrUpdateAppSrcDeploymentInfoList","msg":"App change detected.  Marking for an update","Called with length: ":1,"appName":"demo_my_dashboards.spl"}
```

### Testing

Start a standalone with one app to install:
```
apiVersion: enterprise.splunk.com/v2
kind: Standalone
metadata:
  name: demo-s1
  finalizers:
spec:
  image: splunk/splunk:edge
  imagePullPolicy: IfNotPresent
  replicas: 1
  appRepo:
    appsRepoPollIntervalSeconds: 60
    volumes:
      - name: volume_app_repo1
        path: appframework-demo-apps/apps
        secretRef: demo-af-secret
        storageType: s3
        provider: aws
    appSources:
      - name: admin-apps
        location: standalones/
        volumeName: volume_app_repo1
        scope: local
```

```
[splunk@splunk-demo-s1-standalone-0 splunk]$ ls -Rl /init-apps/
/init-apps/:
total 0
drwxr-sr-x 2 splunk splunk 36 Sep  8 17:30 admin-apps

/init-apps/admin-apps:
total 4
-rw-r--r-- 1 splunk splunk 1804 Sep  8 17:29 demo_my_dashboards.spl
```

Validate that the initial confimap is correct:
```
jryb:AppFramework$ ka ../Demo/yamls/demo-standalone.yaml 
standalone.enterprise.splunk.com/demo-s1 created
jryb:AppFramework$ k get configmap/splunk-demo-s1-standalone-app-list -o=yaml
apiVersion: v1
data:
  app-list-local.yaml: |-
    splunk:
      app_paths_install:
        default:
          - "/init-apps/admin-apps/demo_my_dashboards.spl"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-09T18:20:32Z"
  name: splunk-demo-s1-standalone-app-list
  namespace: default
  ownerReferences:
  - apiVersion: enterprise.splunk.com/v2
    controller: true
    kind: Standalone
    name: demo-s1
    uid: 7432c5bf-e906-4093-97cf-1266fe780eee
  resourceVersion: "68594987"
  selfLink: /api/v1/namespaces/default/configmaps/splunk-demo-s1-standalone-app-list
  uid: 83e09653-359e-4d28-a318-a6b1d797486e
```

Update that single app in the `appRepo`.  Validate that the pod gets restarted and changed app installed even though the data in the configmap remains the same:
```
{"level":"info","ts":1631211632.5365407,"logger":"splunk.enterprise.ValidateAppFrameworkSpec","msg":"App framework configuration is valid"}
{"level":"info","ts":1631211632.5365674,"logger":"splunk.enterprise.HasAppRepoCheckTimerExpired","msg":"App repo polling interval timer has expired","LastAppInfoCheckTime":"0","current epoch time":"1631211632"}
{"level":"info","ts":1631211632.536581,"logger":"splunk.enterprise.initAndCheckAppInfoStatus","msg":"Checking status of apps on remote storage...","name":"demo-s1","namespace":"default"}
{"level":"info","ts":1631211632.5365899,"logger":"splunk.enterprise.GetAppListFromS3Bucket","msg":"Getting the list of apps from remote storage...","name":"demo-s1","namespace":"default"}
{"level":"info","ts":1631211632.541486,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"demo-s1","namespace":"default","volume":"volume_app_repo1","bucket":"appframework-demo-apps","bucket path":"apps/standalones/"}
{"level":"info","ts":1631211632.5416343,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1631211632.541649,"logger":"splunk.client.GetAppsList","msg":"Getting Apps list","AWS S3 Bucket":"appframework-demo-apps"}
{"level":"info","ts":1631211632.6500275,"logger":"splunk.enterprise.initAndCheckAppInfoStatus","msg":"Apps List retrieved from remote storage","name":"demo-s1","namespace":"default","App Source":"admin-apps","Content":[{"Etag":"\"3091594731e8b9c2d550067556f9bf63\"","Key":"apps/standalones/demo_my_dashboards.spl","LastModified":"2021-09-09T00:52:13Z","Size":1804,"StorageClass":"STANDARD"}]}
{"level":"info","ts":1631211632.6501029,"logger":"splunk.enterprise.handleAppRepoChanges","msg":"received App listing","kind":"Standalone","name":"demo-s1","namespace":"default","for App sources":1}
{"level":"info","ts":1631211632.650114,"logger":"splunk.enterprise.AddOrUpdateAppSrcDeploymentInfoList","msg":"New App found","Called with length: ":1,"appName":"demo_my_dashboards.spl"}
{"level":"info","ts":1631211632.661485,"logger":"splunk.reconcile.CreateResource","msg":"Created resource","name":"splunk-demo-s1-standalone-app-list","namespace":"default"}
{"level":"info","ts":1631211632.6709652,"logger":"splunk.reconcile.ApplyConfigMap","msg":"Updating existing ConfigMap","name":"splunk-demo-s1-standalone-app-list","namespace":"default","ResourceVerison":"68594986"}
{"level":"info","ts":1631211632.678169,"logger":"splunk.reconcile.UpdateResource","msg":"Updated resource","name":"splunk-demo-s1-standalone-app-list","namespace":"default"}
{"level":"info","ts":1631211632.6781857,"logger":"splunk.enterprise.SetLastAppInfoCheckTime","msg":"Setting the LastAppInfoCheckTime to current time","current epoch time":1631211632}
```
Show configmap with revision label to force configmap update and pod restart:
```
jryb:AppFramework$   k get configmap/splunk-demo-s1-standalone-app-list -o=yaml
apiVersion: v1
data:
  app-list-local.yaml: |-
    splunk:
      app_paths_install:
        default:
          - "/init-apps/admin-apps/demo_my_dashboards.spl"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-09T18:20:32Z"
  name: splunk-demo-s1-standalone-app-list
  namespace: default
  ownerReferences:
  - apiVersion: enterprise.splunk.com/v2
    controller: true
    kind: Standalone
    name: demo-s1
    uid: 7432c5bf-e906-4093-97cf-1266fe780eee
  resourceVersion: "68595791"
  selfLink: /api/v1/namespaces/default/configmaps/splunk-demo-s1-standalone-app-list
  uid: 83e09653-359e-4d28-a318-a6b1d797486e
```
Check app on Splunk instance:
```
[splunk@splunk-demo-s1-standalone-0 splunk]$ cat etc/apps/demo_my_dashboards/default/app.conf 
#
# Splunk app configuration file
#

[install]
is_configured = 1
install_source_checksum = cd92d38f183c0a27fcd4d403f84088d693f7af8a

[ui]
is_visible = 1
label = My Dashboard

[launcher]
author = Maude Pie
description = App that has some pretty cool dashboards.
version = 2.2.3

```

Switch app in `appRepo` again and make sure this keeps working:
```
{"level":"info","ts":1631211872.041086,"logger":"splunk.reconcile.Reconcile","msg":"Reconciling custom resource","Group":"enterprise.splunk.com","Version":"v2","Kind":"Standalone","Namespace":"default","Name":"demo-s1"}
{"level":"info","ts":1631211872.0457063,"logger":"splunk.enterprise.HasAppRepoCheckTimerExpired","msg":"App repo polling interval timer has expired","LastAppInfoCheckTime":"1631211812","current epoch time":"1631211872"}
{"level":"info","ts":1631211872.045723,"logger":"splunk.enterprise.initAndCheckAppInfoStatus","msg":"Checking status of apps on remote storage...","name":"demo-s1","namespace":"default"}
{"level":"info","ts":1631211872.0457299,"logger":"splunk.enterprise.GetAppListFromS3Bucket","msg":"Getting the list of apps from remote storage...","name":"demo-s1","namespace":"default"}
{"level":"info","ts":1631211872.0522978,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"demo-s1","namespace":"default","volume":"volume_app_repo1","bucket":"appframework-demo-apps","bucket path":"apps/standalones/"}
{"level":"info","ts":1631211872.0524197,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1631211872.0524354,"logger":"splunk.client.GetAppsList","msg":"Getting Apps list","AWS S3 Bucket":"appframework-demo-apps"}
{"level":"info","ts":1631211872.105068,"logger":"splunk.enterprise.initAndCheckAppInfoStatus","msg":"Apps List retrieved from remote storage","name":"demo-s1","namespace":"default","App Source":"admin-apps","Content":[{"Etag":"\"b0cb4c1f540e5488de8f3794afc6034b\"","Key":"apps/standalones/demo_my_dashboards.spl","LastModified":"2021-09-09T18:24:10Z","Size":1804,"StorageClass":"STANDARD"}]}
{"level":"info","ts":1631211872.10511,"logger":"splunk.enterprise.handleAppRepoChanges","msg":"received App listing","kind":"Standalone","name":"demo-s1","namespace":"default","for App sources":1}
{"level":"info","ts":1631211872.1051219,"logger":"splunk.enterprise.AddOrUpdateAppSrcDeploymentInfoList","msg":"App change detected.  Marking for an update.","Called with length: ":1,"appName":"demo_my_dashboards.spl"}
{"level":"info","ts":1631211872.1110475,"logger":"splunk.reconcile.ApplyConfigMap","msg":"Updating existing ConfigMap","name":"splunk-demo-s1-standalone-app-list","namespace":"default","ResourceVerison":"68594987"}
{"level":"info","ts":1631211872.1179957,"logger":"splunk.reconcile.UpdateResource","msg":"Updated resource","name":"splunk-demo-s1-standalone-app-list","namespace":"default"}
{"level":"info","ts":1631211872.121728,"logger":"splunk.reconcile.ApplyConfigMap","msg":"Updating existing ConfigMap","name":"splunk-demo-s1-standalone-app-list","namespace":"default","ResourceVerison":"68595790"}
{"level":"info","ts":1631211872.1286206,"logger":"splunk.reconcile.UpdateResource","msg":"Updated resource","name":"splunk-demo-s1-standalone-app-list","namespace":"default"}
```
Check configmap:
```jryb:AppFramework$ k get configmap/splunk-demo-s1-standalone-app-list -o=yaml
apiVersion: v1
data:
  app-list-local.yaml: |-
    splunk:
      app_paths_install:
        default:
          - "/init-apps/admin-apps/demo_my_dashboards.spl"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-09T18:20:32Z"
  name: splunk-demo-s1-standalone-app-list
  namespace: default
  ownerReferences:
  - apiVersion: enterprise.splunk.com/v2
    controller: true
    kind: Standalone
    name: demo-s1
    uid: 7432c5bf-e906-4093-97cf-1266fe780eee
  resourceVersion: "68596897"
  selfLink: /api/v1/namespaces/default/configmaps/splunk-demo-s1-standalone-app-list
  uid: 83e09653-359e-4d28-a318-a6b1d797486e
```
Check app installed on Splunk instance:
```
jryb:AppFramework$ kex-sc splunk-demo-s1-standalone-0
[splunk@splunk-demo-s1-standalone-0 splunk]$ cat etc/apps/demo_my_dashboards/default/app.conf
#
# Splunk app configuration file
#

[install]
is_configured = 1
install_source_checksum = 685cfb614635dc017bf8b54d345d31b2e0fb2df0

[ui]
is_visible = 1
label = My Dashboard

[launcher]
author = Maude Pie
description = App that has some pretty cool dashboards.
version = 2.2.2

```

### DIFF
--- a/pkg/splunk/controller/configmap.go
+++ b/pkg/splunk/controller/configmap.go
@@ -17,6 +17,7 @@ package controller
 import (
 	"context"
 	"reflect"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -27,7 +28,7 @@ import (
 )
 
 // ApplyConfigMap creates or updates a Kubernetes ConfigMap
-func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap) (bool, error) {
+func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap, forceUpdate bool) (bool, error) {
 	scopedLog := log.WithName("ApplyConfigMap").WithValues(
 		"name", configMap.GetObjectMeta().GetName(),
 		"namespace", configMap.GetObjectMeta().GetNamespace())
@@ -38,8 +39,24 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
 	err := client.Get(context.TODO(), namespacedName, &current)
 	var dataUpdated bool
 	if err == nil {
-		if !reflect.DeepEqual(configMap.Data, current.Data) {
-			scopedLog.Info("Updating existing ConfigMap")
+		if !reflect.DeepEqual(configMap.Data, current.Data) || forceUpdate {
+			scopedLog.Info("Updating existing ConfigMap", "forceUpdate", forceUpdate, "ResourceVerison", current.GetResourceVersion())
+			if forceUpdate {
+				// Add or Increment the revision label to force a new ResourceVersion for the ConfigMap
+				labels := current.GetLabels()
+				if labels == nil {
+					newLabels := make(map[string]string)
+					newLabels["revision"] = "1"
+					current.SetLabels(newLabels)
+				} else if val, ok := labels["revision"]; ok {
+					revision, _ := strconv.Atoi(val)
+					labels["revision"] = strconv.Itoa(revision + 1)
+					current.SetLabels(labels)
+				} else {
+					labels["revision"] = "1"
+					current.SetLabels(labels)
+				}
+			}
 			current.Data = configMap.Data
 			err = splutil.UpdateResource(client, &current)
 			if err == nil {
--- a/pkg/splunk/controller/configmap.go
+++ b/pkg/splunk/controller/configmap.go
@@ -42,20 +42,19 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
 		if !reflect.DeepEqual(configMap.Data, current.Data) || forceUpdate {
 			scopedLog.Info("Updating existing ConfigMap", "forceUpdate", forceUpdate, "ResourceVerison", current.GetResourceVersion())
 			if forceUpdate {
-				// Add or Increment the revision label to force a new ResourceVersion for the ConfigMap
+				// Add or increment the revision label to force a new resourceVersion for the ConfigMap
 				labels := current.GetLabels()
 				if labels == nil {
 					newLabels := make(map[string]string)
 					newLabels["revision"] = "1"
-					current.SetLabels(newLabels)
+					labels = newLabels
 				} else if val, ok := labels["revision"]; ok {
 					revision, _ := strconv.Atoi(val)
 					labels["revision"] = strconv.Itoa(revision + 1)
-					current.SetLabels(labels)
 				} else {
 					labels["revision"] = "1"
-					current.SetLabels(labels)
 				}
+				current.SetLabels(labels)
 			}
 			current.Data = configMap.Data
 			err = splutil.UpdateResource(client, &current)
--- a/pkg/splunk/controller/configmap.go
+++ b/pkg/splunk/controller/configmap.go
@@ -17,7 +17,6 @@ package controller
 import (
 	"context"
 	"reflect"
-	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,7 +27,7 @@ import (
 )
 
 // ApplyConfigMap creates or updates a Kubernetes ConfigMap
-func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap, forceUpdate bool) (bool, error) {
+func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap) (bool, error) {
 	scopedLog := log.WithName("ApplyConfigMap").WithValues(
 		"name", configMap.GetObjectMeta().GetName(),
 		"namespace", configMap.GetObjectMeta().GetNamespace())
@@ -39,23 +38,8 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
 	err := client.Get(context.TODO(), namespacedName, &current)
 	var dataUpdated bool
 	if err == nil {
-		if !reflect.DeepEqual(configMap.Data, current.Data) || forceUpdate {
-			scopedLog.Info("Updating existing ConfigMap", "forceUpdate", forceUpdate, "ResourceVerison", current.GetResourceVersion())
-			if forceUpdate {
-				// Add or increment the revision label to force a new resourceVersion for the ConfigMap
-				labels := current.GetLabels()
-				if labels == nil {
-					newLabels := make(map[string]string)
-					newLabels["revision"] = "1"
-					labels = newLabels
-				} else if val, ok := labels["revision"]; ok {
-					revision, _ := strconv.Atoi(val)
-					labels["revision"] = strconv.Itoa(revision + 1)
-				} else {
-					labels["revision"] = "1"
-				}
-				current.SetLabels(labels)
-			}
+		if !reflect.DeepEqual(configMap.Data, current.Data) {
+			scopedLog.Info("Updating existing ConfigMap", "ResourceVerison", current.GetResourceVersion())
 			current.Data = configMap.Data
 			err = splutil.UpdateResource(client, &current)
 			if err == nil {
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -112,7 +112,7 @@ func ApplySplunkConfig(client splcommon.ControllerClient, cr splcommon.MetaObjec
 	if spec.Defaults != "" {
 		defaultsMap := getSplunkDefaults(cr.GetName(), cr.GetNamespace(), instanceType, spec.Defaults)
 		defaultsMap.SetOwnerReferences(append(defaultsMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
-		_, err = splctrl.ApplyConfigMap(client, defaultsMap, false)
+		_, err = splctrl.ApplyConfigMap(client, defaultsMap)
 		if err != nil {
 			return nil, err
 		}
@@ -331,8 +331,20 @@ func ApplyAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 
 	appListingConfigMap.SetOwnerReferences(append(appListingConfigMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
 
-	if len(appListingConfigMap.Data) > 0 || appsModified {
-		configMapDataChanged, err = splctrl.ApplyConfigMap(client, appListingConfigMap, appsModified)
+	if len(appListingConfigMap.Data) > 0 {
+		if appsModified {
+			// App packages are modified, reset configmap to insure a new resourceVersion
+			scopedLog.Info("Resetting App ConfigMap to force new resourceVersion", "configMapName", configMapName)
+			savedData := appListingConfigMap.Data
+			appListingConfigMap.Data = make(map[string]string)
+			_, err = splctrl.ApplyConfigMap(client, appListingConfigMap)
+			if err != nil {
+				scopedLog.Error(err, "failed reset of configmap", "configMapName", configMapName)
+			}
+			appListingConfigMap.Data = savedData
+		}
+
+		configMapDataChanged, err = splctrl.ApplyConfigMap(client, appListingConfigMap)
 
 		if err != nil {
 			return nil, configMapDataChanged, err
@@ -388,7 +400,7 @@ func ApplySmartstoreConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 	SplunkOperatorAppConfigMap := splctrl.PrepareConfigMap(configMapName, cr.GetNamespace(), mapSplunkConfDetails)
 
 	SplunkOperatorAppConfigMap.SetOwnerReferences(append(SplunkOperatorAppConfigMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
-	configMapDataChanged, err = splctrl.ApplyConfigMap(client, SplunkOperatorAppConfigMap, false)
+	configMapDataChanged, err = splctrl.ApplyConfigMap(client, SplunkOperatorAppConfigMap)
 	if err != nil {
 		return nil, configMapDataChanged, err
 	} else if configMapDataChanged {
@@ -396,7 +408,7 @@ func ApplySmartstoreConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 		mapSplunkConfDetails[configToken] = fmt.Sprintf(`%d`, time.Now().Unix())
 
 		// Apply the configMap with a fresh token
-		configMapDataChanged, err = splctrl.ApplyConfigMap(client, SplunkOperatorAppConfigMap, false)
+		configMapDataChanged, err = splctrl.ApplyConfigMap(client, SplunkOperatorAppConfigMap)
 		if err != nil {
 			return nil, configMapDataChanged, err
 		}
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -333,7 +333,7 @@ func ApplyAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 
 	if len(appListingConfigMap.Data) > 0 {
 		if appsModified {
-			// App packages are modified, reset configmap to insure a new resourceVersion
+			// App packages are modified, reset configmap to ensure a new resourceVersion
 			scopedLog.Info("Resetting App ConfigMap to force new resourceVersion", "configMapName", configMapName)
 			savedData := appListingConfigMap.Data
 			appListingConfigMap.Data = make(map[string]string)
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -112,7 +112,7 @@ func ApplySplunkConfig(client splcommon.ControllerClient, cr splcommon.MetaObjec
 	if spec.Defaults != "" {
 		defaultsMap := getSplunkDefaults(cr.GetName(), cr.GetNamespace(), instanceType, spec.Defaults)
 		defaultsMap.SetOwnerReferences(append(defaultsMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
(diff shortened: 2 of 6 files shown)


pkg/splunk/controller/configmap.go
```diff
@@ -38,8 +39,24 @@ func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigM
...
+					current.SetLabels(newLabels)
+				} else if val, ok := labels["revision"]; ok {
+					revision, _ := strconv.Atoi(val)
+					labels["revision"] = strconv.Itoa(revision + 1)
+					current.SetLabels(labels)
+				} else {
+					labels["revision"] = "1"
+					current.SetLabels(labels)
+				}
+			}
```
sgontla, line 59: I may be wrong, but AFAIK, doing a `UpdateResource()` always gives us a new revision. In that case, we don't need the extra labels?

pkg/splunk/enterprise/util.go
```diff
@@ -331,8 +331,8 @@ func ApplyAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 
 	appListingConfigMap.SetOwnerReferences(append(appListingConfigMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
 
-	if len(appListingConfigMap.Data) > 0 {
-		configMapDataChanged, err = splctrl.ApplyConfigMap(client, appListingConfigMap)
+	if len(appListingConfigMap.Data) > 0 || appsModified {
+		configMapDataChanged, err = splctrl.ApplyConfigMap(client, appListingConfigMap, appsModified)
```
sgontla, line 335: Since it is a unique situation for the Phase-23 app framework, i feel it m

pkg/splunk/controller/configmap.go
jryb: I thought so too but thats not the case.  If you call `UpdateResource()` with the same info, it doesn't update the revision.
```
jryb:AppFramework$ k get configmap/splunk-demo-s1-standalone-app-list -o=yaml
apiVersion: v1
data:
  app-list-local.yaml: |-
    splunk:
      app_paths_install:
        default:
          - "/init-apps/admin-apps/demo_my_dashboards.spl"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-08T22:02:47Z"
  name: splunk-demo-s1-standalone-app-list
  namespace: default
  ownerReferences:
  - apiVersion: enterprise.splunk.com/v2
    controller: true
    kind: Standalone
    name: demo-s1
    uid: fa455462-2046-46a1-aeb3-70db12fa03a7
  resourceVersion: "68382724"
  selfLink: /api/v1/namespaces/default/configmaps/splunk-demo-s1-standalone-app-list
  uid: 21261b76-6e96-4e19-aaa9-8adbab9a9b6b
jryb:AppFramework$ ##### UPDATE S3
```

```
{"level":"info","ts":1631139348.5761461,"logger":"splunk.enterprise.handleAppRepoChanges","msg":"received App listing","kind":"Standalone","name":"demo-s1","namespace":"default","for App sources":1}
{"level":"info","ts":1631139348.5761857,"logger":"splunk.enterprise.AddOrUpdateAppSrcDeploymentInfoList","msg":"App change detected.  Marking for an update.","Called with length: ":1,"appName":"demo_my_dashboards.spl"}
{"level":"info","ts":1631139348.5796883,"logger":"splunk.reconcile.ApplyConfigMap","msg":"Updating existing ConfigMap","name":"splunk-demo-s1-standalone-app-list","namespace":"default","forceUpdate":true,"ResourceVerison":"68382724"}
{"level":"info","ts":1631139348.584546,"logger":"splunk.reconcile.UpdateResource","msg":"Updated resource","name":"splunk-demo-s1-standalone-app-list","namespace":"default"}
```

```
jryb:AppFramework$ k get configmap/splunk-demo-s1-standalone-app-list -o=yaml
apiVersion: v1
data:
  app-list-local.yaml: |-
    splunk:
      app_paths_install:
        default:
          - "/init-apps/admin-apps/demo_my_dashboards.spl"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-08T22:02:47Z"
  name: splunk-demo-s1-standalone-app-list
  namespace: default
  ownerReferences:
  - apiVersion: enterprise.splunk.com/v2
    controller: true
    kind: Standalone
    name: demo-s1
    uid: fa455462-2046-46a1-aeb3-70db12fa03a7
  resourceVersion: "68382724"
  selfLink: /api/v1/namespaces/default/configmaps/splunk-demo-s1-standalone-app-list
  uid: 21261b76-6e96-4e19-aaa9-8adbab9a9b6b

File: pkg/splunk/enterprise/util.go (hunk 2 above)
Comment (jryb) [reply pkg/splunk/enterprise/util.go:335]: Ya this is only applicable for as long as we use the configmap method.  But its needed for the case were the set of applications that has been modified is the same as the previous change (i.e. no change in the configmap data section).

Comment (sgontla) [reply pkg/splunk/enterprise/util.go:335]: My bad, seems like I messed up my comment. What I wanted to mention was, since it is a unique case just for phase-2, we can simply delete the existing configMap and then write a new one, so we don't need to change the ApplyConfigMap API. Probably, we need to have a new API for deleting the configMap.

Changing the conf rev. was a kind of workaround we are using to reset the Pod, otherwise, we never see a use case for the force update to happen the configMap for the same content.  If somebody is using the force option, they should make sure that the code path that gets there is not re-entrant across the reconciles, otherwise it may keep updating unnecessarily.

File: pkg/splunk/controller/configmap.go (hunk 1 above)
Comment (gaurav-splunk) [inline pkg/splunk/controller/configmap.go:59]: Probably change it something like this -
```
                                labels := current.GetLabels()
				if labels == nil {
					newLabels := make(map[string]string)
					newLabels["revision"] = "1"
                                        labels = newLabels
				} else if val, ok := labels["revision"]; ok {
					revision, _ := strconv.Atoi(val)
					labels["revision"] = strconv.Itoa(revision + 1)
				} else {
					labels["revision"] = "1"
				}
                                 current.SetLabels(labels)
```

File: pkg/splunk/enterprise/util.go (hunk 2 above)
Comment (jryb) [reply pkg/splunk/enterprise/util.go:335]: Honestly, either way is fine.  They both require an additional API or API parameter that we can eventually remove once we no longer need the resets.  Only app framework will use the `force` parameter, which does not rerun this code every reconcile due to the appContext status.  I would suspect that even if we created a new ConfigMap it would also have to be non-reentrant or the same issue would occur.

File: pkg/splunk/controller/configmap.go [hunk 3]
```diff
@@ -27,7 +28,7 @@ import (
 )
 
 // ApplyConfigMap creates or updates a Kubernetes ConfigMap
-func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap) (bool, error) {
+func ApplyConfigMap(client splcommon.ControllerClient, configMap *corev1.ConfigMap, forceUpdate bool) (bool, error) {
```
gaurav-splunk, line 31: I wish there was a way to pass optional params in go :)

jryb: Would be nice.  It would make this PR much smaller....

pkg/splunk/enterprise/util.go
sgontla: IMO, having the delete API is a better option for multiple reasons: 
- We are getting a new API :-).  
- We don't need to change the API behavior that is not something for a standard use case. 
- We don't need to address it in several places, which we need to revert again after some time.

Otherwise, I am fine with either option :-).

sgontla: Actually, looking back at it, even the DELETE and CREATE calls are expensive, and we can avoid it.
All that we need is, just one more UPDATE call. i.e calling `splctrl.ApplyConfigMap` once with an empty data map. That should solve our unique situation.

jryb: Ya I agree that a delete and readd is likely much more expensive than a modify.  So I can basically add a reset of the config prior to the set.  This means if we have any app updates and the appListing changes, we would now update the configmap twice: once setting the `Data` section to "" and a second time to set it to the actual `appListing` as opposed to the original fix that only updated the configmap once.  This way there are no changes to the `ApplyConfigMap()` API and this is completely isolated to the AppFramework code.

Since the two updates are in quick secession, there shouldn't be any adverse side effects.  I'm testing it out to verify this.

sgontla: One way we can avoid extra update calls is by doing a deep equal comparison on the existing configMap data. However, that needs a READ CRUD call. So, I guess, we are fine with this approach, considering we are patching a unique use case. :-)

pkg/splunk/enterprise/util.go
```diff
@@ -332,6 +332,18 @@ func ApplyAppListingConfigMap(client splcommon.ControllerClient, cr splcommon.Me
 	appListingConfigMap.SetOwnerReferences(append(appListingConfigMap.GetOwnerReferences(), splcommon.AsOwner(cr, true)))
 
 	if len(appListingConfigMap.Data) > 0 {
+		if appsModified {
```
sgontla, line 335: I think the check @ line 334 does the same thing, and you may not need this extra flag.

jryb: I thought so too and initially tried without the flag.  It doesn't work because if there are apps pending to be installed, the `len(Data)` will be > 0 every reconcile.  So we will reset the configMap each time causing multiple restarts.  The `appModified` flag will only be set to true when the `GetList()` of the app sources differs from what the operator has, regardless of the deployment state of the apps.  So this check is needed.

sgontla: I think, from day one, we didn't want the App framework code to be re-entrant, otherwise, we will be messing up the state/status info.  If we are entering this code across the reconciliations, wondering if we have any other bug that popping up now :-)

sgontla (approved):

gaurav-splunk (approved):

smohan-splunk (approved):

pkg/splunk/enterprise/util.go
jryb: I don't think so.  The `appListing` is build from the app status for any app that has yet to be fully installed.  This makes since because if the Pod gets reset for some reason prior to ansible finishing up the install, we need that list to persist so they can be still be installed.  
The `appModified` flag is a trigger to tell us to reset the configmap for the `appListing`, particularly for the case where the app set to be installed is the same as before, just the app packages have been modified.  This would only get triggered once when our S3 client code identifies a change in the app packages and updates its state.  We only need this once since if a restart/failure happens and the apps are not fully installed, the ConfigMap will still be correct.

sgontla: If we are entering multiple times, then, it may cause some issues particularly from the state machine point of view. Pod template doesn't look at `appListing`, and just uses the configMap If exists. 

As that is a separate issue, let us discuss it tomorrow, and see if we really need a fix for the same.

smohan-splunk: CSPL-1311 created to discuss this further

jambrosiano:
Verification details in CSPL-1312, 1.0.3 operator build with fix behaviour is as expected.

PR #506, opened 2021-09-13
CSPL-1316: Avoid app framework flow from re-entrancy

sgontla:
When the App framework flow is in execution, re-entring the flow can cause the following issues:
1. For SHC, CR phase `Ready` doesn’t include the deployer, so, we should not mark the app deployment as complete. Right now, once the SHC becomes ready, we are marking the apps are deployed. So, if the AppSources are changing in between while the app installation is in progress, as the config Map is changing, we are marking all the apps as deployed, when some of them are not even installed: which needs to be fixed.

2. Even after the above case is fixed, still, the re-entrance is a problem, as the deployer Pod is ready, we are going to mark all the apps as deployed. On follow-up Pod reset if any app installation fails, our CR says all the apps are deployed, but that is not fully correct. This is the same problem for all the deployments. 

Code changes mentioned here addresses: 
1. Avoid the app framework flow re-entering during the reconciles
2. For SHC, use the deployer phase for marking the app installation status to complete.

### DIFF
--- a/pkg/splunk/enterprise/clustermaster.go
+++ b/pkg/splunk/enterprise/clustermaster.go
@@ -89,6 +89,7 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	if len(cr.Spec.AppFrameworkConfig.AppSources) != 0 {
 		err := initAndCheckAppInfoStatus(client, cr, &cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
 		if err != nil {
+			cr.Status.AppContext.IsDeploymentInProgress = false
 			return result, err
 		}
 	}
@@ -144,6 +145,11 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	if cr.Status.Phase == splcommon.PhaseReady {
 		if cr.Status.AppContext.AppsSrcDeployStatus != nil {
 			markAppsStatusToComplete(client, cr, &cr.Spec.AppFrameworkConfig, cr.Status.AppContext.AppsSrcDeployStatus)
+			// Schedule one more reconcile in next 5 seconds, just to cover any latest app framework config changes
+			if cr.Status.AppContext.IsDeploymentInProgress {
+				cr.Status.AppContext.IsDeploymentInProgress = false
+				return result, nil
+			}
 		}
 
 		err = ApplyMonitoringConsole(client, cr, cr.Spec.CommonSplunkSpec, getClusterMasterExtraEnv(cr, &cr.Spec.CommonSplunkSpec))
--- a/pkg/splunk/enterprise/searchheadcluster.go
+++ b/pkg/splunk/enterprise/searchheadcluster.go
@@ -53,6 +53,7 @@ func ApplySearchHeadCluster(client splcommon.ControllerClient, cr *enterpriseApi
 	if len(cr.Spec.AppFrameworkConfig.AppSources) != 0 {
 		err := initAndCheckAppInfoStatus(client, cr, &cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
 		if err != nil {
+			cr.Status.AppContext.IsDeploymentInProgress = false
 			return result, err
 		}
 	}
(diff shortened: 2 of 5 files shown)


pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -147,12 +148,17 @@ func ApplySearchHeadCluster(client splcommon.ControllerClient, cr *enterpriseApi
 	}
 	cr.Status.Phase = phase
 
-	// no need to requeue if everything is ready
-	if cr.Status.Phase == splcommon.PhaseReady {
-		if cr.Status.AppContext.AppsSrcDeployStatus != nil {
-			markAppsStatusToComplete(client, cr, &cr.Spec.AppFrameworkConfig, cr.Status.AppContext.AppsSrcDeployStatus)
+	if cr.Status.AppContext.AppsSrcDeployStatus != nil && cr.Status.DeployerPhase == splcommon.PhaseReady {
```
jryb, line 151: Does this mean for SHC members ready state is not based on the app install status?

pkg/splunk/enterprise/util.go
```diff
@@ -950,13 +947,16 @@ func HasAppRepoCheckTimerExpired(appInfoContext *enterpriseApi.AppDeploymentCont
...
-// so that the next reconile would happen at the right time.
+// so that the next reconcile would happen at the right time.
 func GetNextRequeueTime(appRepoPollInterval, lastCheckTime int64) time.Duration {
 	scopedLog := log.WithName("GetNextRequeueTime")
 	currentEpoch := time.Now().Unix()
 
 	var nextRequeueTimeInSec int64
 	nextRequeueTimeInSec = appRepoPollInterval - (currentEpoch - lastCheckTime)
+	if nextRequeueTimeInSec < 0 {
+		nextRequeueTimeInSec = 5
```
jryb, line 958: minimum requeue time?

gaurav-splunk: When can this be negative?

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -89,6 +89,7 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	if len(cr.Spec.AppFrameworkConfig.AppSources) != 0 {
 		err := initAndCheckAppInfoStatus(client, cr, &cr.Spec.AppFrameworkConfig, &cr.Status.AppContext)
 		if err != nil {
+			cr.Status.AppContext.IsDeploymentInProgress = false
```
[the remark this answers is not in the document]
sgontla: No, the behavior is same, except, we just starting to use this flag, hence resetting it.

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -144,6 +145,11 @@ func ApplyClusterMaster(client splcommon.ControllerClient, cr *enterpriseApi.Clu
 	if cr.Status.Phase == splcommon.PhaseReady {
 		if cr.Status.AppContext.AppsSrcDeployStatus != nil {
 			markAppsStatusToComplete(client, cr, &cr.Spec.AppFrameworkConfig, cr.Status.AppContext.AppsSrcDeployStatus)
+			// Schedule one more reconcile in next 5 seconds, just to cover any latest app framework config changes
```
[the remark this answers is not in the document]
sgontla: Correct.

pkg/splunk/enterprise/searchheadcluster.go
sgontla: Correct. Deployer will not come to Phase ready unless the SHC is up and apps are really installed.

pkg/splunk/enterprise/util.go
sgontla: Correct, this is the default we use across the reconciliations, so using the same value here. I added this, just to make sure we are not getting into the -ve values when installing larger apps.

pkg/splunk/enterprise/clustermaster.go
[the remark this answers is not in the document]
gaurav-splunk: so what happens during the next reconcile? We wont do anything since `IsDeploymentInProgress == false` ?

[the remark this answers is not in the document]
sgontla: Good question. On the next reconcile, `initAndCheckAppInfoStatus()` will proceed to check, if the config changed, OR if the timer expired.

pkg/splunk/enterprise/util.go
sgontla: Say, if the polling interval is 1 min, and If we are installing a large app like ES App, we are not setting the `LastAppInfoCheckTime` for as high as 15 minutes. In such cases, it helps.

gaurav-splunk (approved):

jryb (approved):

smohan-splunk (approved):

PR #509, opened 2021-09-13
Bias language exceptions

mgaldino-splunk:


test/smoke/cluster_master_sites_response.go
```diff
@@ -15,7 +15,7 @@
 package smoke
 
 // ClusterMasterSitesResponse is a representation of the sites managed by a Splunk cluster-manager
-// Endpoint: /services/cluster/master/sites
```
mgaldino-splunk, line 18: Some endpoints have been migrated to use manager as 
[here](https://docs.splunk.com/Documentation/Splunk/8.2.2/RESTREF/RESTcluster#cluster.2Fmanager.2Fsites). But not all. 

Our general approach was to migrate all comments while using constants for literals (non-comments).

pkg/splunk/common/urls_testingcode.go
```diff
@@ -0,0 +1,403 @@
...
..."splunk-stack1-cluster-master-service"}],"resources":{"limits":{"cpu":"4","memory":"8Gi"},"requests":{"cpu":"100m","memory":"512Mi"}},"volumeMounts":[{"name":"mnt-splunk-etc","mountPath":"/opt/splunk/etc"},{"name":"mnt-splunk-var","mountPath":"/opt/splunk/var"},{"name":"mnt-splunk-secrets","mountPath":"/mnt/splunk-secrets"}],"livenessProbe":{"exec":{"command":["/sbin/checkstate.sh"]},"initialDelaySeconds":300,"timeoutSeconds":30,"periodSeconds":30},"readinessProbe":{"exec":{"command":["/bin/grep","started","/opt/container_artifact/splunk-container.state"]},"initialDelaySeconds":10,"timeoutSeconds":5,"periodSeconds":5},"imagePullPolicy":"IfNotPresent"}],"securityContext":{"runAsUser":41812,"fsGroup":41812},"affinity":{"podAntiAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"weight":100,"podAffinityTerm":{"labelSelector":{"matchExpressions":[{"key":"app.kubernetes.io/instance","operator":"In","values":["splunk-test-monitoring-console"]}]},"topologyKey":"kubernetes.io/hostname"}}]}},"schedulerName":"default-scheduler"}},"serviceName":"splunk-test-monitoring-console-headless","podManagementPolicy":"Parallel","updateStrategy":{"type":"OnDelete"}},"status":{"replicas":0}}`
```
mgaldino-splunk, line 264: @pdhanoyasplunk & @jambrosiano I was puzzled to find a way to optimize the Body responses but did not find a solution without requiring a lot of effort/refactoring. Please feel free to add any suggestions.

.biased_lang_exclude
```diff
@@ -0,0 +1,4 @@
+pkg/splunk/common/urls.go
+pkg/splunk/common/urls_testingcode.go
+pkg/splunk/common/paths.go
+deploy/olm-catalog/
```
mgaldino-splunk, line 4: This line adds the Legacy CRDs as exceptions for the Bias Language process.

pkg/splunk/common/util_test.go
```diff
@@ -572,12 +572,12 @@ func TestGetLabels(t *testing.T) {
...
 
-	test("indexer", "cluster-master", "t1", "t1", map[string]string{
+	test("indexer", CM, "t1", "t1", map[string]string{
 		"app.kubernetes.io/managed-by": "splunk-operator",
 		"app.kubernetes.io/component":  "indexer",
-		"app.kubernetes.io/name":       "cluster-master",
+		"app.kubernetes.io/name":       CM,
 		"app.kubernetes.io/part-of":    "splunk-t1-indexer",
-		"app.kubernetes.io/instance":   "splunk-t1-cluster-master",
+		"app.kubernetes.io/instance":   ManagerT1,
```
smohan-splunk, line 580: Naming convention has been changed.. any reason why we just didn't use "cluster-manager" & "splunk-t1-cluster-manager" ?

test/c3/appframework/appframework_test.go
```diff
@@ -413,8 +413,8 @@ var _ = Describe("c3appfw test", func() {
 			testenvInstance.Log.Info("Verify Apps are installed Locally on CM and Deployer by running Splunk CLI commands for app", "version", appVersion)
 			testenv.VerifyAppInstalled(deployment, testenvInstance, testenvInstance.GetName(), podNames, appListV1, false, "enabled", false, false)
 
-			// Verify apps are not copied in /etc/master-apps/ on CM and /etc/shcluster/ on Deployer (therefore not installed on peers and on SH)
-			testenvInstance.Log.Info("Verify Apps are NOT copied to /etc/master-apps/ on CM and /etc/shcluster/ on Deployer for app", "version", appVersion, "App List", appFileList)
+			// Verify apps are not copied in /etc/manager-apps/ on CM and /etc/shcluster/ on Deployer (therefore not installed on peers and on SH)
+			testenvInstance.Log.Info("Verify Apps are NOT copied to "+splcommon.ManagerApps+" on CM and /etc/shcluster/ on Deployer for app", "version", appVersion, "App List", appFileList)
 			testenv.VerifyAppsCopied(deployment, testenvInstance, testenvInstance.GetName(), podNames, appListV1, false, true)
```
smohan-splunk, line 418: For consistency, can we replace
`"on CM and /etc/shcluster/ on Deployer for app"`
with
`"on CM and "+splcommon.SHCApps+" on Deployer for app"`

test/c3/appframework/appframework_test.go
```diff
@@ -413,8 +413,8 @@ var _ = Describe("c3appfw test", func() {
 			testenvInstance.Log.Info("Verify Apps are installed Locally on CM and Deployer by running Splunk CLI commands for app", "version", appVersion)
 			testenv.VerifyAppInstalled(deployment, testenvInstance, testenvInstance.GetName(), podNames, appListV1, false, "enabled", false, false)
 
-			// Verify apps are not copied in /etc/master-apps/ on CM and /etc/shcluster/ on Deployer (therefore not installed on peers and on SH)
-			testenvInstance.Log.Info("Verify Apps are NOT copied to /etc/master-apps/ on CM and /etc/shcluster/ on Deployer for app", "version", appVersion, "App List", appFileList)
+			// Verify apps are not copied in /etc/manager-apps/ on CM and /etc/shcluster/ on Deployer (therefore not installed on peers and on SH)
```
smohan-splunk, line 416: Is it really "/etc/manager-apps/" now? `splcommon.ManagerApps` still points to `"etc/master-apps"` in paths.go.  May be its better to modify the comment to remove the bias language

test/testenv/appframework_utils.go
```diff
@@ -49,7 +50,7 @@ func GenerateAppSourceSpec(appSourceName string, appSourceLocation string, appSo
 // GetPodAppStatus Get the app install status and version number
 func GetPodAppStatus(deployment *Deployment, podName string, ns string, appname string, clusterWideInstall bool) (string, string, error) {
 	// For clusterwide install do not check for versions on deployer and cluster-manager as the apps arent installed there
-	if clusterWideInstall && (strings.Contains(podName, "-cluster-master-") || strings.Contains(podName, "-deployer-")) {
+	if clusterWideInstall && (strings.Contains(podName, splcommon.CMDashed) || strings.Contains(podName, "-deployer-")) {
 		logf.Log.Info("Pod skipped as install is Cluter-wide", "PodName", podName)
```
smohan-splunk, line 54: For consistency, can we also add `DeployerDashed`?

smohan-splunk (approved):

test/c3/appframework/appframework_test.go
mgaldino-splunk: Resolved in the latest commit.

pkg/splunk/common/urls.go
```diff
@@ -0,0 +1,134 @@
...
+	//Config = "/config"
+	Config = "/config"
+
+	//LocalsPeer = "/localslave"
+	LocalsPeer = "/localslave"
+
+	//Licenser = "/licenser"
+	Licenser = "/licenser"
+
+	//QM = Question Mark
```
sgontla, line 126: Nit: Adding this comment for this entire file. If the idea behind this tokenization is to toggle the correct keyword (ex. Master vs. Manager)based on the Splunk version, then, probably, we can just limit to those terms(ex. master, slave, etc..), so that this will not be bloating going forward. 
Alternatively, If you prefer to go with this format, I think, it will be better to have some sorted order of the RHS value, so that, when somebody wants to frame a new URI/URL, they know where to look for :-)

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -117,18 +117,18 @@ func TestGetClusterMasterStatefulSet(t *testing.T) {
...
+	test(splcommon.TestGetCMStatefulSet)
```
sgontla, line 120: Returned STS spec is the result of the config we are passing above, and there is not much we can leverage with this change(?). Any reason why you want to move it to a common place?

Same comment for other places.

pkg/splunk/enterprise/indexercluster_test.go
```diff
@@ -143,7 +143,7 @@ func TestGetClusterMasterClient(t *testing.T) {
 	c := spltest.NewMockClient()
 	mgr.c = c
 	cm := mgr.getClusterMasterClient()
-	if cm.ManagementURI != "https://splunk--cluster-master-service.test.svc.cluster.local:8089" {
+	if cm.ManagementURI != splcommon.ManagerServiceTest {
```
sgontla, line 146: I feel it is more readable if you can just replace the `master` keyword alone with variable substitution.  Same comment with other places.

pkg/splunk/common/urls_testingcode.go
```diff
@@ -0,0 +1,409 @@
...
+	PeerDecommission = LocalURL + DecommissionAPI
+
+	//ApplyBundleTest = "https://localhost:8089/services/cluster/master/control/default/apply"
+	ApplyBundleTest = LocalURL + ApplyBundleAPI
+
+	//ManagerInfoTest = "https://localhost:8089/services/cluster/master/info?count=0&output_mode=json"
+	ManagerInfoTest = LocalURL + ManagerInfoAPI + TestCommonOutput
+
+	//ManagerInfoTestJSON = "https://localhost:8089/services/cluster/master/info?output_mode=json"
+	ManagerInfoTestJSON = LocalURL + ManagerInfoAPI + JSONOutput
```
sgontla, line 28: Considering my other comments, assuming if we are going with this kind of tokenization, wondering if a naming convention like `UriServiersClusterMasterInfo`, `UriServicesClusterMasterPeers`, `UriServicesClusterMasterPeersCount` etc..  which gives us better readability wherever these are being used.

pkg/splunk/common/urls.go
gaurav-splunk: I agree with Subba, as much as it looks nice tokenizing the strings, it sort of is an over-kill TBH. You could have just tokenized the string in demand i.e. master vs manager or slave vs peer, etc.

pkg/splunk/enterprise/util_test.go
```diff
@@ -119,7 +119,7 @@ func TestGetLicenseMasterURL(t *testing.T) {
 	want = []corev1.EnvVar{
 		{
 			Name:  "SPLUNK_LICENSE_MASTER_URL",
-			Value: "splunk-stack1-license-master-service.test.svc.cluster.local",
+			Value: splcommon.Stack1LMClusterLocal,
```
gaurav-splunk, line 122: probably better to create/use `fmt.Sprintf()` format to create the string?
So lets say, 
```
in pkg/splunk/common ----
LMClusterLocal= "splunk-%s-license-master-service.test.svc.cluster.local"

and here in this pkg ----
Value: fmt.Sprintf(splcommon.LMClusterLocal, "stack1")
```

pkg/splunk/enterprise/util_test.go
```diff
@@ -105,7 +105,7 @@ func TestGetLicenseMasterURL(t *testing.T) {
 	want := []corev1.EnvVar{
 		{
 			Name:  "SPLUNK_LICENSE_MASTER_URL",
-			Value: "splunk-stack1-license-master-service",
+			Value: splcommon.Stack1LMService,
```
gaurav-splunk, line 108: Same as below comment

pkg/splunk/common/urls.go
mgaldino-splunk: Hi @gaurav-splunk & @sgontla, this is good feedback. The main advantage of tokenization was to reduce the amount of hardcoded Paths & URLs. So in the future new constants can be defined in a single place. This is especially true for URLs that are reused many times across the code, but surely an over-kill for single usages.

But I also understand the readability concern since we no longer see a full URL in the code and the sense of an over-kill/bloating with too many tokenizations.

I think we should discuss these two options and see what the team thinks is a better approach moving forward.

I opened this sample PR for comparison: https://github.com/splunk/splunk-operator/pull/540/files

Let me know if you think that's better :D

pkg/splunk/enterprise/clustermaster_test.go
mgaldino-splunk: This is because in the response there are cases with biased language such as master/slave.

pkg/splunk/enterprise/indexercluster_test.go
mgaldino-splunk: Let's discuss this suggested approach. I opened this mock PR for comparison: https://github.com/splunk/splunk-operator/pull/540/files

pkg/splunk/common/urls_testingcode.go
mgaldino-splunk: Sounds like a good idea to use RHS naming convention. Will change if we keep this approach.

mgaldino-splunk: Changed the naming convention in the latest commit. Will complete the changes after our discussion later today.

pkg/splunk/common/urls.go
```diff
@@ -0,0 +1,134 @@
...
+
+	//*****************
+	//   Base Hosts
+	//*****************
+
+	//Port8089 = ":8089"
+	Port8089 = ":8089"
+
+	//LocalURL = "https://localhost:8089"
+	LocalURL = "https://localhost" + Port8089
```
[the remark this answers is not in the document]
mgaldino-splunk: Done

pkg/splunk/enterprise/indexercluster_test.go
mgaldino-splunk: Resolved in the latest commit which we reduced the amount tokenizations and improved the naming convention.

pkg/splunk/common/urls_testingcode.go
mgaldino-splunk: Resolved in the latest commit which we reduced the amount tokenizations and improved the naming convention.

pkg/splunk/common/paths.go
```diff
@@ -0,0 +1,23 @@
...
+	PeerApps = "etc/slave-apps"
+
+	//ManagerApps = "etc/master-apps"
+	ManagerApps = "etc/master-apps"
+
+	//SHCluster = "etc/shcluster/apps"
+	SHCluster = "etc/shcluster"
+
+	//SHClusterApps = "etc/shcluster/apps"
+	SHClusterApps = SHCluster + "/apps"
```
sgontla, line 22: Nit: Can we name them with the convention like: `PeerAppsLoc`, `ManagerAppsLoc`, etc.. so that it will be easy to maintain in future?

pkg/splunk/common/urls.go
```diff
@@ -0,0 +1,54 @@
...
+	APIClusterManagerApplyBundle = APICLusterManagerServices + "/control/default/apply"
+
+	//APIClusterManagerGetHealth = "/services/cluster/master/health?"
+	APIClusterManagerGetHealth = APICLusterManagerServices + "/health?"
+
+	//APIClusterManagerGetSites = "/services/cluster/master/sites?"
+	APIClusterManagerGetSites = APICLusterManagerServices + "/sites?"
+
+	//APIClusterManagerGetSearchHeads = "/services/cluster/master/searchheads?"
+	APIClusterManagerGetSearchHeads = APICLusterManagerServices + "/searchheads?"
```
sgontla, line 37: For all the items listed here, I think we better name them Uri, than Api, as these two are different?

mgaldino-splunk: Updated in the latest commit. Also based on our conversation we added these changes:
1) Corrected URI vs URL usage
2) Removed generic URLs from testing file to ease future usage
3) Added variable ID/counters using fmt.Sprintft() when needed.

PR #515, opened 2021-09-17
cspl-1256: API for the file copy from Operator Pod to any other CR pods

sgontla:
- API that accepts Pod details, source, and destination file paths.
- Sanity check the source and destination paths
- Uses the tar on the sending side, and untar on the receiving side.


sgontla:
Though it is not a test to derive the thruput numbers, tested copying a 10GB file, and it took around 76 seconds, which is around 1Gbps.

`{"level":"info","ts":1631905269.5589511,"logger":"splunk.enterprise.CopyFileToPod","msg":"Normalized file paths","podName":"splunk-s2apps-standalone-0","namespace":"default","srcPath":"/tmp/10GB.spl","destPath":"/init-apps/splunkFwdApps/","src path":"/tmp/10GB.spl","dest path":"/init-apps/splunkFwdApps/10GB.spl"}
{"level":"info","ts":1631905269.6798935,"logger":"splunk.enterprise.CopyFileToPod","msg":"Directory on Pod exists","podName":"splunk-s2apps-standalone-0","namespace":"default","srcPath":"/tmp/10GB.spl","destPath":"/init-apps/splunkFwdApps/","(test -d) result":0}
{"level":"info","ts":1631905345.6247246,"logger":"splunk.enterprise.ApplyStandalone","msg":"File Copied Successfully.","name":"s2apps","namespace":"default","localFile":"/tmp/10GB.spl","podFile":"/init-apps/splunkFwdApps/","resp":"","kind":"Standalone"}`

```
Every 1.0s: ls -l                                                            splunk-s2apps-standalone-0: Fri Sep 17 19:06:24 2021

total 16836808
-rw-r--r-- 1 splunk splunk 10240000000 Sep 17 18:57 10GB.spl
-rw-r--r-- 1 splunk splunk  7000883200 Sep 17 18:55 COPYING
-rw-r--r-- 1 splunk splunk         790 Jun 22 22:43 appSplunkForwarder.tgz

```

pkg/splunk/util/util.go
```diff
@@ -139,7 +138,7 @@ func generateHECToken() []byte {
 }
 
 // PodExecCommand execute a shell command in the specified pod
-func PodExecCommand(c splcommon.ControllerClient, podName string, namespace string, cmd []string, stdin string, tty bool, mock bool) (string, string, error) {
+func PodExecCommand(c splcommon.ControllerClient, podName string, namespace string, cmd []string, streamOptions *remotecommand.StreamOptions, tty bool, mock bool) (string, string, error) {
```
jryb, line 141: Just wondering what the particular reason was to move the `streamOptions` parameter to the callers responsibility?

jryb (approved):

pkg/splunk/util/util.go
sgontla: Good question. StreamOptions take the 'io.Reader' interface.  Existing code is taking string as stdin, which is converted as interface `strings.Reader`.  Now, with this API, we are passing the `PipeReader`. When I changed it to the generic `io.Reader`, it didn't go well, due to compatibility issues. Also, typecasting at the calling place didn't help.  So, felt this is the clean way, as the caller knows what will be the Stdin input.

smohan-splunk (approved):

gaurav-splunk (approved):

smohan-splunk (approved):

PR #527, opened 2021-10-02
CSPL-1334: Added framework for downloading multiple apps in parallel on operator pod

gaurav-splunk:
**Description**
This is PR is the first part of the bigger work in Phase-3 of the app framework to install apps in the following ways - 
1. Download apps on the operator pod at a specific volume mounted on the operator pod i.e. at this location **/opt/splunk/appframwork**. We then create sub-directories under this path to download apps per CR.
2. Copy downloaded apps to Splunk pod.
3. In the case of cluster scoped apps, untar the apps at the desired location on the splunk pod
4. Issue app installation command to either install the app on standalone pod or issue bundle push command in clustering scenario.

This PR targets only the download part of it.

**Key features as part of this PR**
1. Introduced new config in the CR spec to specify the maximum number of app downloads in parallel, i.e. *maxConcurrentAppDownloads*
2. Introduced new App install states to track the app install state. Right now it only tracks the download of apps but eventually, we would like to track the copy and untar of apps too.
3. Introduced a new Workerpool to initialize and spawn workers which will do the actual work of installing the apps in parallel.
4. The only eligibility for the app to be downloaded is if we have enough disk space available to download the app on operato pod.

**Sample operator logs**
Here are some sample operator logs showing the download of apps in parallel on operator pod - 
```
{"level":"info","ts":1633128050.2653751,"logger":"splunk.enterprise.handleAppInstall","msg":"Installing apps for app source","name":"s1","namespace":"default","appSource":"admin3"}
{"level":"info","ts":1633128050.2655275,"logger":"splunk.enterprise.getAvailableDiskSpace","msg":"current available disk space in GB","volume mount":"/opt/splunk/appframework/","available disk space":7}
{"level":"info","ts":1633128050.2655592,"logger":"splunk.enterprise.handleAppInstallPerAppSrc","msg":"Starting the work to install apps..","name":"s1","namespace":"default"}
{"level":"info","ts":1633128050.2655864,"logger":"splunk.enterprise.isAppAlreadyDownloaded","msg":"App not present on operator pod","app name":"smart-store-106.tgz"}
{"level":"info","ts":1633128050.2655964,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-106.tgz","old install state":"Download Not Started","new install state":"Download In Progress"}
{"level":"info","ts":1633128050.26564,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"Started install of app","worker id":5,"appName":"smart-store-106.tgz"}
{"level":"info","ts":1633128050.2657602,"logger":"splunk.enterprise.isAppAlreadyDownloaded","msg":"App not present on operator pod","app name":"smart-store-216.tgz"}
{"level":"info","ts":1633128050.2657707,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-216.tgz","old install state":"Download Not Started","new install state":"Download In Progress"}
{"level":"info","ts":1633128050.2657928,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"Started install of app","worker id":1,"appName":"smart-store-216.tgz"}
{"level":"info","ts":1633128050.2658784,"logger":"splunk.enterprise.isAppAlreadyDownloaded","msg":"App not present on operator pod","app name":"smart-store-816.tgz"}
{"level":"info","ts":1633128050.2658901,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-816.tgz","old install state":"Download Not Started","new install state":"Download In Progress"}
{"level":"info","ts":1633128050.265914,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"Started install of app","worker id":2,"appName":"smart-store-816.tgz"}
{"level":"info","ts":1633128050.2659957,"logger":"splunk.enterprise.isAppAlreadyDownloaded","msg":"App not present on operator pod","app name":"smart-store.tgz"}
{"level":"info","ts":1633128050.2660065,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store.tgz","old install state":"Download Not Started","new install state":"Download In Progress"}
{"level":"info","ts":1633128050.2660308,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"Started install of app","worker id":3,"appName":"smart-store.tgz"}
{"level":"info","ts":1633128050.2661073,"logger":"splunk.enterprise.isAppAlreadyDownloaded","msg":"App not present on operator pod","app name":"updated_app-66.tgz"}
{"level":"info","ts":1633128050.2661238,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"updated_app-66.tgz","old install state":"Download Not Started","new install state":"Download In Progress"}
{"level":"info","ts":1633128050.2661467,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"Started install of app","worker id":4,"appName":"updated_app-66.tgz"}
{"level":"info","ts":1633128050.2685301,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"s1","namespace":"default","volume":"volume_repo_3","bucket":"gaurav-sample-bucket","bucket path":"folder3/admin/"}
{"level":"info","ts":1633128050.268645,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1633128050.2696228,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"s1","namespace":"default","volume":"volume_repo_3","bucket":"gaurav-sample-bucket","bucket path":"folder3/admin/"}
{"level":"info","ts":1633128050.2697139,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1633128050.2702298,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"s1","namespace":"default","volume":"volume_repo_3","bucket":"gaurav-sample-bucket","bucket path":"folder3/admin/"}
{"level":"info","ts":1633128050.2703483,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1633128050.2707462,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"s1","namespace":"default","volume":"volume_repo_3","bucket":"gaurav-sample-bucket","bucket path":"folder3/admin/"}
{"level":"info","ts":1633128050.2708347,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1633128050.2721314,"logger":"splunk.enterprise.GetRemoteStorageClient","msg":"Creating the client","name":"s1","namespace":"default","volume":"volume_repo_3","bucket":"gaurav-sample-bucket","bucket path":"folder3/admin/"}
{"level":"info","ts":1633128050.2722175,"logger":"splunk.client.InitAWSClientSession","msg":"AWS Client Session initialization successful.","region":"us-west-2","TLS Version":"TLS 1.2"}
{"level":"info","ts":1633128050.325256,"logger":"splunk.client.DownloadApp","msg":"File downloaded","remoteFile":"folder3/admin/smart-store-816.tgz","localFile":"/opt/splunk/appframework/s1/local/admin3/smart-store-816.tgz_4cd06eb80b131a88bb473dd298e000c4","numBytes: ":4608}
{"level":"info","ts":1633128050.3252954,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-816.tgz","old install state":"Download In Progress","new install state":"Download Complete"}
{"level":"info","ts":1633128050.3253083,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"finished job","worker id":2}
{"level":"info","ts":1633128050.3253117,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"availableDiskSpace after freeing up space","worker id":2,"availableDiskSpace":8266391552}
{"level":"info","ts":1633128050.333039,"logger":"splunk.client.DownloadApp","msg":"File downloaded","remoteFile":"folder3/admin/smart-store.tgz","localFile":"/opt/splunk/appframework/s1/local/admin3/smart-store.tgz_376eba5f53063c5a23aa52229609d97e","numBytes: ":4608}
{"level":"info","ts":1633128050.3330684,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store.tgz","old install state":"Download In Progress","new install state":"Download Complete"}
{"level":"info","ts":1633128050.333077,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"finished job","worker id":3}
{"level":"info","ts":1633128050.3330812,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"availableDiskSpace after freeing up space","worker id":3,"availableDiskSpace":8266391552}
{"level":"info","ts":1633128050.3363197,"logger":"splunk.client.DownloadApp","msg":"File downloaded","remoteFile":"folder3/admin/updated_app-66.tgz","localFile":"/opt/splunk/appframework/s1/local/admin3/updated_app-66.tgz_dab7e59daacce62b9b1b273d1fcb38ef","numBytes: ":5632}
{"level":"info","ts":1633128050.3363578,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"updated_app-66.tgz","old install state":"Download In Progress","new install state":"Download Complete"}
{"level":"info","ts":1633128050.336365,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"finished job","worker id":4}
{"level":"info","ts":1633128050.3363698,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"availableDiskSpace after freeing up space","worker id":4,"availableDiskSpace":8266391552}
{"level":"info","ts":1633128050.355428,"logger":"splunk.client.DownloadApp","msg":"File downloaded","remoteFile":"folder3/admin/smart-store-216.tgz","localFile":"/opt/splunk/appframework/s1/local/admin3/smart-store-216.tgz_4119ad7df0881d49cfc059603702dba3","numBytes: ":4608}
{"level":"info","ts":1633128050.3554573,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-216.tgz","old install state":"Download In Progress","new install state":"Download Complete"}
{"level":"info","ts":1633128050.3554647,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"finished job","worker id":1}
{"level":"info","ts":1633128050.355469,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"availableDiskSpace after freeing up space","worker id":1,"availableDiskSpace":8266391552}
{"level":"info","ts":1633128050.3601468,"logger":"splunk.client.DownloadApp","msg":"File downloaded","remoteFile":"folder3/admin/smart-store-106.tgz","localFile":"/opt/splunk/appframework/s1/local/admin3/smart-store-106.tgz_ac7b30b04ed7d1bd64e8e308d3a59cfc","numBytes: ":4608}
{"level":"info","ts":1633128050.360188,"logger":"splunk.enterprise.setAppInstallState","msg":"Setting the install state for app","app name":"smart-store-106.tgz","old install state":"Download In Progress","new install state":"Download Complete"}
{"level":"info","ts":1633128050.360196,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"finished job","worker id":5}
{"level":"info","ts":1633128050.3602002,"logger":"splunk.enterprise.AppInstallWorkerPool.run()","msg":"availableDiskSpace after freeing up space","worker id":5,"availableDiskSpace":8266391552}
```


pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,184 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package enterprise
+
+import (
+	"fmt"
+	"io/fs"
```
gaurav-splunk, line 19: Because of this package, the compilation is failing while trying to run UTs. In my env everything passes since my go version is 1.16.x but the CircleCI server has go version 1.13.x and I think this is why it is failing. @pdhanoyasplunk anything we can do to get around this error? Maybe update the go version on CircleCI?

pkg/splunk/common/names.go
```diff
@@ -83,6 +83,9 @@ const (
 
 	// MaxAppsRepoPollInterval sets the polling interval to one day
 	MaxAppsRepoPollInterval int64 = 60 * 60 * 24
+
+	// AppDownloadVolume is hte mount volume on the operator pod to temporarily download apps
```
jryb, line 87: spelling nit: the

pkg/splunk/enterprise/configuration.go
```diff
@@ -1151,6 +1173,17 @@ func ValidateAppFrameworkSpec(appFramework *enterpriseApi.AppFrameworkSpec, appC
 		appContext.AppsRepoStatusPollInterval = splcommon.MaxAppsRepoPollInterval
 	}
 
+	if appContext.AppsStatusMaxConcurrentAppDownloads <= 0 {
+		scopedLog.Info("Invalid value of maxConcurrentAppDownloads", "configured value", appContext.AppsStatusMaxConcurrentAppDownloads, "Defaulting to 5.")
+		appContext.AppsStatusMaxConcurrentAppDownloads = 5
```
jryb, line 1178: Should we make this a const instead of hardcoding a 5 here?

pkg/splunk/enterprise/configuration.go
```diff
@@ -1151,6 +1173,17 @@ func ValidateAppFrameworkSpec(appFramework *enterpriseApi.AppFrameworkSpec, appC
...
+	}
+
+	// check whether the temporary volume to download apps is mounted or not on the operator pod
+	if _, err := os.Stat(splcommon.AppDownloadVolume); os.IsNotExist(err) {
+		scopedLog.Error(err, "directory /opt/splunk/appframework/ needs to be mounted on operator pod to download apps. Please mount this as a separate volume on operator pod.")
```
jryb, line 1183: Use `AppDownloadVolume` in the error message.

pkg/splunk/enterprise/util.go
```diff
@@ -762,13 +961,14 @@ func AddOrUpdateAppSrcDeploymentInfoList(appSrcDeploymentInfo *enterpriseApi.App
 			if appList[idx].AppName == appName {
 				found = true
 				if appList[idx].ObjectHash != *remoteObj.Etag || appList[idx].RepoState == enterpriseApi.RepoStateDeleted {
-					scopedLog.Info("App change detected.", "App name: ", appName, "marking for an update")
+					scopedLog.Info("App change detected,marking for an update.", "App name", appName)
```
jryb, line 964: Thanks for fixing these!

pkg/splunk/enterprise/util.go
```diff
@@ -1044,11 +1245,20 @@ func initAndCheckAppInfoStatus(client splcommon.ControllerClient, cr splcommon.M
 				return err
 			}
 
+			// TODO: gaurav, this needs to be removed eventually in favor of downloading apps locally
 			_, _, err = ApplyAppListingConfigMap(client, cr, appFrameworkConf, appStatusContext.AppsSrcDeployStatus)
 			if err != nil {
 				return err
 			}
 
+			// TODO: gaurav, this will eventually go outside this if condition,
```
jryb, line 1254: So does this mean that we will handle all app downloads/installs in a single iteration of the reconcile loop?  Do we wait on all worker threads to finish before this call returns?  Will this block servicing other reconciles?  I'm fine with the caveat but we should probably make that explicitly know in this comment.

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,184 @@
...
+	"os"
+	"sync"
+
+	enterpriseApi "github.com/splunk/splunk-operator/pkg/apis/enterprise/v2"
+	splclient "github.com/splunk/splunk-operator/pkg/splunk/client"
+)
+
+// AppInstallWorkerPool is a pool of workers to install apps in parallel
+type AppInstallWorkerPool struct {
```
jryb, line 28: Nice work!  Will this struct be able to handle all parameters needed to download and install apps for a given CR?  Will the pool be used across CRs or per CR?

pkg/splunk/enterprise/configuration.go
gaurav-splunk: Yup, right good point!

pkg/splunk/enterprise/util.go
gaurav-splunk: As of right now, yes we are doing everything in one reconcile and will wait until all the downloads are finished. @sgontla is going to introduce the yield logic through which we can continue the work across reconciles.

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: This is per CR

pkg/splunk/enterprise/util.go
```diff
@@ -232,6 +234,203 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+func getAppInstallState(appInfo *enterpriseApi.AppDeploymentInfo) enterpriseApi.AppInstallState {
+	return appInfo.AppInstallStatus.AppInstallState
+}
+
+// createAppDownloadDir creates the app download directory on the operator pod
+func createAppDownloadDir(path string) error {
+	scopedLog := log.WithName("createAppDownloadDir").WithValues("path", path)
+	_, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		errDir := os.MkdirAll(path, 0755)
```
sgontla, line 303: Do we need to use `Mkdir()` here? Otherwise, If the customer doesn't configure the vol, then MkdirAll can start going to the container storage?

pkg/splunk/enterprise/util.go
```diff
@@ -232,6 +234,203 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+		if err != nil {
+			return err
+		}
+
+		// get the app source spec to get the volume configuration
+		var vol enterpriseApi.VolumeSpec
+		appSrc, err := getAppSrcSpec(appFrameworkConfig.AppSources, appSrcName)
+		if err != nil {
+			return err
+		}
```
sgontla, line 402: Now that this code is going to be yielding and re-entering. `appFrameworkConfig` may be changing across the reconcile entries. In that case, I guess, we need to handle it more gracefully, than treating it as an error.  Probably, for now, we can add a comment here?
I think this is also a test case we need to cover in our functional testing. CC @pdhanoyasplunk

pkg/splunk/enterprise/util.go
```diff
@@ -232,6 +234,203 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+
+		vol, err = splclient.GetAppSrcVolume(appSrc, appFrameworkConfig)
+		if err != nil {
+			return err
+		}
```
sgontla, line 407: Same as above

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,184 @@
...
+	remoteSize := *object.Size
+	if localSize != remoteSize {
+		err = fmt.Errorf("size of app on operator pod does not match size on remote storage. localSize: %d, remoteSize:%d", localSize, remoteSize)
+		scopedLog.Error(err, "need to re-download app")
+		return false
+	}
+
+	// If we reached here, this means app was downloaded successfuly to operator pod,
+	// mark this as DownloadComplete
+	setAppInstallState(app, enterpriseApi.DownloadComplete)
```
sgontla, line 124: Nit: To make the semantics clean, the caller of this API can call `setAppInstallState`, when true

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,184 @@
...
+	return true
+}
+
+// Start kicks off the workers to do the job of app installation
+func (w *AppInstallWorkerPool) Start() {
+	// start the workers
+	for i := 1; i <= int(w.NumOfWorkers); i++ {
+		go w.Run(i, w.JobsChan)
+	}
```
sgontla, line 133: I think we need to handle this in a more controlled way, otherwise the go routines may end prematurely, even before the jobs queue is filled up? It may not be an issue, as long as the main thread is not doing any blocking calls before adding the jobs to the channel.
Alternatively, maybe moved to the line after 171, if that works?. I think addressing it in a more controlled way(wait for the main thread to give an indication) is preferable, such that, we don't need to worry in case even if dealing with parallel executions involving multiple CPUs.

pkg/apis/enterprise/v2/common_types.go
```diff
@@ -288,6 +314,7 @@ type AppDeploymentInfo struct {
 	Size             uint64              `json:"Size,omitempty"`
 	RepoState        AppRepoState        `json:"repoState"`
 	DeployStatus     AppDeploymentStatus `json:"deployStatus"`
+	AppInstallStatus AppInstallStatus    `json:"appInstallStatus"`
```
sgontla, line 317: I know, this is the first code we are getting in, but, going forward, I think we need to break up the `AppInstallStatus` into multiple phases(probably slightly different naming, say `podInstall`, `podCopy`, `operatorDownload`, etc..), and each phase pointing to separate auxiliary status info.  This way, it will be clear to move the phases on error conditions. For example, if the phase is `podInstall`, and if the app package is not found on the Pod, can be moved to phase `podCopy`. If the app package is not found in the phase `podCopy`, the phase can be moved to 'operatorDownload` etc..

Also,  it will be much easy to decouple the workers, and also to accommodate for any future functionality needs, if we want to introduce in between.

gaurav-splunk: Yeah that's the plan going forward, and that is the reason we created a struct for the states. Sorry for the terminology but in `AppInstallStatus ` we can have something like `AppDownloadstate`, `AppCopyState` and `AppInstallState` for things to be more clear.

pkg/splunk/enterprise/util.go
gaurav-splunk: If the customer doesn't configure the volume, we would fail in the validation stage itself and would not even reach here.

pkg/splunk/enterprise/util.go
```diff
@@ -232,6 +234,203 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+
+	scopedLog := log.WithName("handleAppInstall").WithValues("cr", cr.GetName(), "namespace", cr.GetNamespace())
+
+	var scope string
+	for appSrcName, appDeployInfo := range appDeployContext.AppsSrcDeployStatus {
+		scopedLog.Info("Installing apps for app source", "appSource", appSrcName)
+
+		scope = getAppSrcScope(appFrameworkConfig, appSrcName)
+
+		localPath := filepath.Join(splcommon.AppDownloadVolume, cr.GetName(), scope, appSrcName) + "/"
```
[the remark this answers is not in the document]
gaurav-splunk: That check is already been taken care of in the app framework spec validation in the beginning.

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: Yeah I think either way its fine  TBH :)

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: Yeah, so that is the interesting thing about channels in goroutine. The workers won't exit prematurely since they are waiting on jobs in the jobs channel. And unless you close the channel they(goroutines) won't exit. And you only close the channel once all the jobs are sent in the channel, hence I think we are safe here.

sgontla: Ahh, then it makes sense.

pkg/splunk/enterprise/workerpool.go
sgontla: `isAppAlreadyDownloaded` sounds like `true/false` return kind :-). That is why, either we better rename it to something more logical OR, handle it at the calling place :-)

pkg/splunk/enterprise/util.go
sgontla: Are you referring to the error in line# 320? If that is the case, probably, we can modify the error msg to reflect, something like to configure the volume as suggested in one of our docs

pkg/splunk/enterprise/util.go
smohan-splunk: @sgontla Trying to understand your comment better: `getAppSrcSpec` should retrieve the app source for the CR whose reconcile we are in, and if that fails, we should throw an error I would assume. Is there a specific scenario you had in mind when you mentioned handle more gracefully?

pkg/splunk/enterprise/util.go
```diff
@@ -1044,11 +1245,20 @@ func initAndCheckAppInfoStatus(client splcommon.ControllerClient, cr splcommon.M
 				return err
 			}
 
+			// TODO: gaurav, this needs to be removed eventually in favor of downloading apps locally
 			_, _, err = ApplyAppListingConfigMap(client, cr, appFrameworkConf, appStatusContext.AppsSrcDeployStatus)
```
smohan-splunk, line 1249: Why are we not already removing/commenting this now? Wouldn't he configMap update interfere with Phase 3 functioning?

pkg/splunk/enterprise/workerpool.go
smohan-splunk: Good point .. wonder if we need a pool of pools to avoid OOM etc?

pkg/splunk/common/names.go
```diff
@@ -83,6 +83,9 @@ const (
...
+	AppDownloadVolume = "/opt/splunk/appframework/"
```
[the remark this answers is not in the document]
gaurav-splunk: Yeah probably we can !

pkg/splunk/enterprise/util.go
gaurav-splunk: We will remove this once we have all the pieces ready to install the app/s via operator pod. We are just in the download stage at the moment.

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: I am sorry but did not understand the concerns here. @jryb @smohan-splunk ?

pkg/splunk/enterprise/util.go
sgontla: @smohan-splunk, my comment was for the same CR. Say, if the app framework config changes, while the app installation is already in progress.

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -684,6 +684,11 @@ spec:
                         description: Remote Storage Volume name
                         type: string
                     type: object
+                  maxConcurrentAppDownloads:
```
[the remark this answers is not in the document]
gaurav-splunk: Created following story for this - https://jira.splunk.com/browse/CSPL-1365. CC: @kashok-splunk

deploy/olm-catalog/splunk/1.0.2/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -684,6 +684,11 @@ spec:
                         description: Remote Storage Volume name
                         type: string
                     type: object
+                  maxConcurrentAppDownloads:
```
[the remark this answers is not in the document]
gaurav-splunk: So what do you propose we do about this in this PR?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2740,11 +2740,11 @@ spec:
                                 format: int64
                                 type: integer
                               appInstallStatus:
-                                description: AppInstallState represents the current
+                                description: AppInstallStatus represents the current
                                   install state of the app
                                 properties:
                                   appInstallState:
```
smohan-splunk, line 2746: Should this be AppDownloadState instead of AppInstallState. There is a mismatch between the param name and the description. Same applies to other CRDs too

gaurav-splunk: huh strange, let me check again.

pkg/splunk/enterprise/workerpool.go
jryb: I didn't have a particular concern just wondering if each pool is per CR or per operator instance.  If it's per CR then is there a concern that there could be too many threads open for operators that are managing a number of CRs?  My guess would be that this likely wouldn't be an issue of running out of threads or memory, but might be worth a thought.

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,191 @@
...
+
+// isAppAlreadyDownloaded checks if the app is already downloaded on the operator pod by checking a couple of things -
+// 1. is the app bundle present on the operator pod &&
+// 2. if the size on disk matches the size reported from remote storage
+func (w *AppInstallWorkerPool) isAppAlreadyDownloaded(app *enterpriseApi.AppDeploymentInfo) bool {
+	scopedLog := log.WithName("isAppAlreadyDownloaded").WithValues("app name", app.AppName)
+	object := getRemoteObjectFromS3Response(app.AppName, w.S3Reponse)
+	localAppFileName := getLocalAppFileName(w.DownloadPath, app.AppName, *object.Etag)
+
+	var fileInfo os.FileInfo
```
gaurav-splunk, line 104: Just for reference, I got the build issue on CircleCI resolved by using `os.FileInfo` instead of `fs.FileInfo` and hence importing package `os` rather than `io/fs`. So, for older versions of golang(<1.16.x) there was `FileInfo` struct which was implemented in `os` package but in newer versions(>=1.16.x) it is implemented under `io/fs` package. Hence instead of using  `fs.FileInfo` I had to use `os.FileInfo` which is backward compatible too in the sense that for newer versions, FileInfo is mapped to `fs.FileInfo`.

sgontla (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -244,6 +246,211 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+	for _, object := range s3Response.Objects {
+		rcvdAppName := extractAppNameFromKey(*object.Key)
+		if rcvdAppName == appName {
+			return object
+		}
+	}
+	return nil
+}
+
+// appDownloadStateAsStr cnoverts the state enum to corresponding string
```
smohan-splunk, line 284: Nit: "converts"

pkg/splunk/enterprise/util.go
```diff
@@ -244,6 +246,211 @@ func GetSmartstoreRemoteVolumeSecrets(volume enterpriseApi.VolumeSpec, client sp
...
+	case enterpriseApi.DownloadInProgress:
+		return "Download In Progress"
+	case enterpriseApi.DownloadComplete:
+		return "Download Complete"
+	default:
+		return "Download Error"
+	}
+}
+
+// setAppDownloadState sets tha app install state for an app
```
smohan-splunk, line 298: Nit: sets app download state. Applies to `getAppDownloadState` too

pkg/splunk/enterprise/util.go
smohan-splunk: Interesting point. I wonder if there is an easy way to determine if the `appFrameworkConfig.AppSources` has changed midway of App installations, and hence its ok for `getAppSrcSpec` to fail.

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -684,6 +684,11 @@ spec:
                         description: Remote Storage Volume name
                         type: string
                     type: object
+                  maxConcurrentAppDownloads:
+                    description: Maximum number of apps that can be downloaded at
```
smohan-splunk, line 688: Am I right in saying the `maxConcurrentAppDownloads` will be 1 global config which is applicable  for all App sources of the CR?

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,191 @@
...
+	var jobsAdded int
+	// add the apps to the jobs channel so that workers can pick them up
+	for {
+		for index := 0; index < len(w.JobList); index++ {
+			app := w.JobList[index]
+
+			switch getAppDownloadState(app) {
+			case enterpriseApi.DownloadNotStarted:
+				// check if we have already downloaded the app
+				if w.isAppAlreadyDownloaded(app) {
```
smohan-splunk, line 147: Under what circumstances would a successfully downloaded app have status as `DownloadNotStarted`?

pkg/splunk/enterprise/workerpool.go
```diff
@@ -0,0 +1,191 @@
...
+		Scope:              scope,
+		S3ClientMgr:        s3ClientMgr,
+		S3Reponse:          s3Response,
+		availableDiskSpace: availableDiskSpace,
+		wg:                 new(sync.WaitGroup),
+	}
+	return w
+}
+
+// Run is the main work that workers do to install the apps.
```
smohan-splunk, line 58: Am I right in saying the main work is to download the apps and not install?

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
gaurav-splunk: Yes

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2755,6 +2774,11 @@ spec:
                       are different.
                     format: int64
                     type: integer
+                  appsStatusMaxConcurrentAppDownloads:
```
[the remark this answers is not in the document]
gaurav-splunk: Possibly, let me give it a shot and see if we can do away without it

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: As of now, it's download, but we plan to extend it to copy and other stuff too. If we plan to create separate threads for other tasks, then I can rename it to suggest it is for Download only.

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: Under rare scenario, where while we have downloaded the app but have not yet set the download state as DownloadComplete and the operator crashed/restarted. Then we again come here to check if we had already downloaded the app, if yes then we don't need to redownload again.

deploy/olm-catalog/splunk/1.0.2/enterprise.splunk.com_clustermasters_crd.yaml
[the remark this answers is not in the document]
gaurav-splunk: ok, I'll remove the field from this OLM directory for now and add a task to add it back after our next release.

smohan-splunk (approved):

PR #544, opened 2021-10-15
CSPL-1363: Added UTs for parallel app downloads

gaurav-splunk:


pkg/splunk/enterprise/workerpool.go
```diff
@@ -127,29 +128,32 @@ func (w *AppInstallWorkerPool) Start() {
...
+				jobsParsed++
 				continue
 			default:
 				continue
 			}
 		}
 
 		// all the jobs are added to the channel, we are done here
-		if jobsAdded == len(w.JobList) {
+		if jobsParsed == len(w.JobList) {
```
jryb, line 148: So if all jobs are in `DownloadInProgress` or `DownloadError` we will never reach this condition, correct?

pkg/splunk/enterprise/workerpool.go
```diff
@@ -127,29 +128,32 @@ func (w *AppInstallWorkerPool) Start() {
...
 			break
 		}
 	}
 
-	// close the jobs channel now
-	close(w.JobsChan)
+	// close the jobs channel now, only if it is not closed yet
+	if actualJobsAdded {
+		close(w.JobsChan)
```
jryb, line 155: For my knowledge: if no jobs are added to the channel, it never gets opened?

pkg/splunk/enterprise/workerpool_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// remove the apps locally to test download again
+	for idx := 0; idx < len(appNames); idx++ {
+		appFileName := appNames[idx] + "_" + objectHashes[idx]
+		appLoc := filepath.Join(localPath, appFileName)
+		os.Remove(appLoc)
+	}
+
+	// Reduce the availableDiskSpace to test scenario where we dont have enough disk space.
+	workPool = NewAppInstallWorkerPool(splcommon.DefaultMaxConcurrentAppDownloads, appDeployInfoList,
```
jryb, line 209: Is this testing the case when we have enough space to download and push only one app at a time so they have to be queued up serially?

pkg/splunk/enterprise/workerpool_test.go
```diff
@@ -0,0 +1,322 @@
...
+// initialize some data structures to be used in this file
+func init() {
+	appSrcName = "appSrc1"
+	scope = "local"
+	crName = "dummyStandalone"
+	randomTime = "1633975734"
+	storageClass = "gp2"
+
+	appNames = []string{"app1.tgz", "app2.tgz", "app3.tgz", "app4.tgz"}
+	objectHashes = []string{"abcd1111", "efgh2222", "ijkl3333", "wxyz4444"}
```
jryb, line 39: Do we have a test case that cover an updated hash/app?

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: yup, right

pkg/splunk/enterprise/workerpool_test.go
gaurav-splunk: I think that is covered as part of the `util` framework. We reach at this code if there are any new apps or updated apps and hence we download them.

pkg/splunk/enterprise/workerpool_test.go
gaurav-splunk: This is testing the case where we are able to push few apps and then for the rest of them we have to wait till we have enough space. And then once we have the available space, we will pick them up from the channel

pkg/splunk/enterprise/workerpool.go
gaurav-splunk: So `DownloadInProgress` is just a transient state before the actual download begins. If due to some reason download is taking more time, we will loop through and will increment the jobsParsed count once the download is complete.
But, about the `DownloadError` state, I think you are right. We probably have to join `DownloadComplete` and `DownloadError` in a single case. Let me check and fix it.

PR #563, opened 2021-10-29
CSPL-1384: Parallel downloads with app framework scheduler

gaurav-splunk:


pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -538,9 +691,18 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
+	availableDiskSpace, err := getAvailableDiskSpace()
+	if err != nil {
+		return err
+	}
+	afwPipeline.availableDiskSpace = availableDiskSpace
+
 	// Start the download phase manager
 	afwPipeline.phaseWaiter.Add(1)
-	go afwPipeline.downloadPhaseManager()
+	go afwPipeline.downloadPhaseManager(appDeployContext)
```
sgontla, line 705: Let us have the reference for appDeployContext,  at the overall pipeline level, so that other phases also can have access to it. Other PR is covering this change.

pkg/splunk/enterprise/util.go
```diff
@@ -1172,6 +1129,31 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+	// check if the app is present on operator pod
+	file, err := os.Open(localAppFileName)
+
+	if os.IsNotExist(err) {
+		scopedLog.Info("App not present on operator pod")
+		return false
+	}
+
+	// update the state to be download complete
+	updatePplnWorkerPhaseInfo(downloadWorker.appDeployInfo, enterpriseApi.PhaseDownload, 0, enterpriseApi.AppPkgDownloadComplete)
```
sgontla, line 1151: Nit: Can we do this step at the calling place of this function, so that this function can be re-used across the phases?

pkg/splunk/enterprise/util.go
```diff
@@ -1172,6 +1129,31 @@ func setupAppInitContainers(client splcommon.ControllerClient, cr splcommon.Meta
...
+	scopedLog := log.WithName("isAppAlreadyDownloaded").WithValues("app name", downloadWorker.appDeployInfo.AppName)
+
+	scope := getAppSrcScope(downloadWorker.afwConfig, downloadWorker.appSrcName)
+	kind := downloadWorker.cr.GetObjectKind().GroupVersionKind().Kind
+
+	localPath := filepath.Join(splcommon.AppDownloadVolume, "downloadedApps", downloadWorker.cr.GetNamespace(), kind, downloadWorker.cr.GetName(), scope, downloadWorker.appSrcName) + "/"
+	localAppFileName := getLocalAppFileName(localPath, downloadWorker.appDeployInfo.AppName, downloadWorker.appDeployInfo.ObjectHash)
+
+	// check if the app is present on operator pod
+	file, err := os.Open(localAppFileName)
```
sgontla, line 1143: nit: os.stat() may be light, as we don't need to close it?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -273,12 +281,160 @@ func getPhaseInfoByPhaseType(worker *PipelineWorker, phaseType enterpriseApi.App
...
 
+// updatePplnWorkerPhaseInfo updates the in-memory PhaseInfo
+func updatePplnWorkerPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, phaseType enterpriseApi.AppPhaseType, retryCount int32, statusType enterpriseApi.AppPhaseStatusType) {
+	scopedLog := log.WithName("updatePplnWorkerPhaseInfo").WithValues("appName", appDeployInfo.AppName)
+
+	scopedLog.Info("changing the status", "old status", appPhaseStatusAsStr(appDeployInfo.PhaseInfo.Status), "new status", appPhaseStatusAsStr(statusType))
+	appDeployInfo.PhaseInfo.Phase = phaseType
+	appDeployInfo.PhaseInfo.RetryCount = retryCount
+	appDeployInfo.PhaseInfo.Status = statusType
+}
```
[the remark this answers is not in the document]
sgontla: Ya, with respect to the status it is fine. But,  the idea is, it should not touch the `Phase`. I know, here you are writing the same phase again, but we don't want to give that provision, just to avoid any misuses in future.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -290,17 +446,13 @@ downloadPhase:
...
 				if checkIfWorkerIsEligibleForRun(downloadWorker, phaseInfo, enterpriseApi.AppPkgDownloadComplete) {
 					downloadWorker.waiter = &pplnPhase.workerWaiter
-					select {
-					case pplnPhase.msgChannel <- downloadWorker:
-						scopedLog.Info("Download worker got a run slot", "name", downloadWorker.cr.GetName(), "namespace", downloadWorker.cr.GetNamespace(), "App name", downloadWorker.appDeployInfo.AppName, "digest", downloadWorker.appDeployInfo.ObjectHash)
-						downloadWorker.isActive = true
-					default:
-						downloadWorker.waiter = nil
-					}
+					pplnPhase.msgChannel <- downloadWorker
```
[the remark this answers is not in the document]
sgontla: Ah ok. I had a similar situation, hence, as part of other PR, i made the msgChannel a buffered channel, so this should not be an issue anymore.

pkg/apis/enterprise/v2/common_types.go
```diff
@@ -71,24 +71,6 @@ const (
 	DeployStatusError
 )
 
-// AppDownloadState represents the download status of app
```
jryb, line 74: This is now rolled into the app deployment status, correct?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -273,12 +280,177 @@ func getPhaseInfoByPhaseType(worker *PipelineWorker, phaseType enterpriseApi.App
...
+					// increment the retry count and mark this app as download pending
+					updatePplnWorkerPhaseInfo(appDeployInfo, appDeployInfo.PhaseInfo.RetryCount+1, enterpriseApi.AppPkgDownloadPending)
+					continue
+				}
+
+				// get the S3ClientMgr instance
+				s3ClientMgr, _ := getS3ClientMgr(*downloadWorker.client, downloadWorker.cr, downloadWorker.afwConfig, downloadWorker.appSrcName)
+
+				// increment the number of active workers
+				activeWorkers++
```
jryb, line 422: Do we need locks around `activeWorkers` here as well?

pkg/splunk/enterprise/afwscheduler_test.go
```diff
@@ -833,7 +847,506 @@ func TestAfwGetReleventStatefulsetByKind(t *testing.T) {
...
+		},
+	}
+	sts := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "splunk-s1-standalone",
+			Namespace: "test",
+		},
+	}
+	sts.Spec.Replicas = new(int32)
+	*sts.Spec.Replicas = 1
```
jryb, line 1224: Do we need a test with multiple standalone replicas?  Or maybe this was part of another commit or I missed it.

pkg/splunk/enterprise/util.go
```diff
@@ -338,117 +343,63 @@ func getAvailableDiskSpace() (uint64, error) {
...
-		}
+	vol, err = splclient.GetAppSrcVolume(*appSrc, appFrameworkConfig)
+	if err != nil {
+		scopedLog.Error(err, "unable to get volume spec")
+		return nil, err
 	}
 
-	return err
+	s3ClientWrapper := splclient.S3Clients[vol.Provider]
+	initFunc := s3ClientWrapper.GetS3ClientInitFuncPtr()
```
jryb, line 392: Not a comment on this functionality/MR but the S3 lib in general.  I wonder if we can save the s3Client for future use.  Just thinking out loud but this might be why we see so many IAM connections, because we create a new s3 session every time?

pkg/apis/enterprise/v2/common_types.go
gaurav-splunk: Yeah as part of PhaseInfo

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: Probably not here since `scheduleDownloads` is only a single thread

pkg/splunk/enterprise/util.go
gaurav-splunk: I have been thinking about it as well for some time now. Let me create a separate task for it so that we can track it properly.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -273,12 +280,177 @@ func getPhaseInfoByPhaseType(worker *PipelineWorker, phaseType enterpriseApi.App
...
+
+downloadWork:
+	for {
+		if activeWorkers <= maxWorkers {
+			select {
+			case _, channelOpen := <-ppln.sigTerm:
+				if !channelOpen {
+					scopedLog.Info("Received the termination request from the scheduler")
+					break downloadWork
+				}
```
sgontla, line 372: Can we use the msgChannel close as the exit criteria, that way the control is just in between the parent and child, to accommodate any future clean-up activity(if required) for the parent?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -534,10 +707,19 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
-	afwPipeline = initAppInstallPipeline()
+	afwPipeline = initAppInstallPipeline(appDeployContext)
 
 	var clusterScopedApps []*enterpriseApi.AppDeploymentInfo
 
+	afwEntryTime := time.Now().Unix()
+
+	// get the current available disk space for downloading apps on operator pod
+	availableDiskSpace, err := getAvailableDiskSpace()
+	if err != nil {
```
sgontla, line 718: Now that we have a return here, can you please move the defer function at line 817 here, so that we are sure that all the memory is released? I should have done it in my original PR, but seems like I missed it.

pkg/splunk/enterprise/util.go
sgontla: Just a question for  my understanding, is it safe to assume that, once we come out of the scheduler, all the connections are closed?

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: Yes that defer function is too far below, needs to be somewhere in the beginning of this function.

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: Sure

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: I think its better if I use the counting semaphore method as is used by Subba in his PR, that way we will have a consistent code and we will not this field

sgontla (approved):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2739,22 +2739,13 @@ spec:
...
                               appName:
                                 type: string
                               auxPhaseInfo:
-                                description: Used by App framework to track the copy
-                                  and install status for each replica member of standalone
-                                  CR. Each Pod's phase info is mapped to its ordinal
-                                  value.
+                                description: Used to track the copy and install status
+                                  for each replica member. Each Pod's phase info is
+                                  mapped to its ordinal value. Ignored, once the DeployStatus
```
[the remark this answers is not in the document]
gaurav-splunk: I think what Subba meant here was actually the ordinal value , as in here `splunk-s1-standalone-0` the ordinal value is `0`.

smohan-splunk (approved):

PR #586, opened 2021-11-12
CSPL-1443: Remove references to old MCPodReady function.

pdhanoyasplunk:
Remove references to old MCPodReady function. 
Added test to integration pipeline which were previously skipped.
Added changes to resolve MC test flakiness

### DIFF
--- a/test/monitoring_console/monitoring_console_test.go
+++ b/test/monitoring_console/monitoring_console_test.go
@@ -389,12 +389,12 @@ var _ = Describe("Monitoring Console test", func() {
 				2. Deploy Monitoring Console
 				3. Wait for Monitoring Console status to go back to READY
 				4. Verify SH are configured in MC Config Map
-				5. Verify SH are configured in peer strings on MC Pod
-				6. Verify Monitoring Console Pod has peers in Peer string on MC Pod
+				5. VerifyMonitoring Console Pod has Search Heads in Peer strings
+				6. Verify Monitoring Console Pod has peers(indexers) in Peer string
 				7. Scale SH Cluster
-				8. Scale Indexer Count on One Site
+				8. Scale Indexer Count
 				9. Add a standalone
-				10. Verify Standalone is configured in MC Config Map
+				10. Verify Standalone is configured in MC Config Map and Peer String
 				11. Verify SH are configured in MC Config Map and Peers String
 				12. Verify Indexers are configured in Peer String
 			*/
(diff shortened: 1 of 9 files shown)


test/scaling_test/scaling_test.go
```diff
@@ -50,7 +50,7 @@ var _ = Describe("Scaling test", func() {
 	})
 
 	Context("Standalone deployment (S1)", func() {
-		It("scaling_test: Can Scale Up and Scale Down Standalone CR", func() {
+		It("scaling_test, integration: Can Scale Up and Scale Down Standalone CR", func() {
 
 			standalone, err := deployment.DeployStandalone(deployment.GetName(), "", "")
 			Expect(err).To(Succeed(), "Unable to deploy standalone instance ")
```
tpereirasplunk, line 56: On line 61 we check for  'testenv.GetConfiguredPeers' which also runs on MC pod. Think we need to remove that too

tpereirasplunk (approved):

kashok-splunk (approved):

test/monitoring_console/monitoring_console_test.go
```diff
@@ -381,7 +381,6 @@ var _ = Describe("Monitoring Console test", func() {
 		})
 	})
 
-	// TO BE READDED WHEN CSPL-1379 fixed
 	Context("Clustered deployment (C3 - clustered indexer, search head cluster)", func() {
 		It("monitoringconsole, smoke: MC can configure SHC, indexer instances after scale up and standalone in a namespace", func() {
 			/*
```
jambrosiano, line 386: Nit: I don't really get what are some of these steps:
"5. Verify SH are configured in peer strings on MC Pod" What is "peer strings" here? Is it actually like "Check Search Head Pods in Monitoring Console Config Map"?
"6. Verify Monitoring Console Pod has peers in Peer string on MC Pod" Basically here it says "MC pod" twice, that's redundant.
"8. Scale Indexer Count on One Site" It's a single site cluster, "on One Site" could be removed

test/monitoring_console/monitoring_console_test.go
```diff
@@ -501,7 +500,7 @@ var _ = Describe("Monitoring Console test", func() {
 			testenv.SearchHeadClusterReady(deployment, testenvInstance)
 
 			// Wait for MC to go to Updating Phase
```
jambrosiano, line 502: Nit: If we now wait for Pending instead of Updating, would be more clear to edit that comment above too

smohan-splunk (approved):

test/monitoring_console/monitoring_console_test.go
pdhanoyasplunk: These are old comments not as part of this change however good idea I can clean up.

pdhanoyasplunk: . Verify SH are configured in peer strings on MC Pod" What is "peer strings" here? Is it actually like "Check Search Head Pods in Monitoring Console Config Map"?
MC has `configured peer string` in one of the files thats how it configures SH's, CM, Depoyer, standalone and indexers. We check this value and compare with pods in the namespace to verify they are added to MC.

MC config map check are separate

jambrosiano (approved):

PR #590, opened 2021-11-16
CSPL-1332: Cluster scoped playbook impementation for indexer cluster

gaurav-splunk:
This PR targets the following things - 
1. Implement a runbook for cluster scoped apps for indexer cluster
2. Runbook will trigger a bundle push and will check the status of bundle push in each reconcile.

### DIFF
--- a/deploy/olm-catalog/splunk/1.0.3/splunk.v1.0.3.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/splunk/1.0.3/splunk.v1.0.3.clusterserviceversion.yaml
@@ -120,8 +120,7 @@ spec:
       - kind: Secrets
         version: v1
       version: v1beta1
-    - displayName: Standalone
-      kind: ClusterMaster
+    - kind: ClusterMaster
       name: clustermasters.enterprise.splunk.com
       resources:
       - kind: StatefulSets
@@ -137,6 +136,7 @@ spec:
       - kind: Secrets
         version: v1
       version: v1alpha3
+      displayName: Standalone
     - displayName: Standalone
       kind: ClusterMaster
       name: clustermasters.enterprise.splunk.com
--- a/deploy/olm-catalog/splunk/1.0.3/splunk.v1.0.3.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/splunk/1.0.3/splunk.v1.0.3.clusterserviceversion.yaml
@@ -120,7 +120,8 @@ spec:
       - kind: Secrets
         version: v1
       version: v1beta1
-    - kind: ClusterMaster
+    - displayName: Standalone
+      kind: ClusterMaster
       name: clustermasters.enterprise.splunk.com
       resources:
       - kind: StatefulSets
@@ -136,7 +137,6 @@ spec:
       - kind: Secrets
         version: v1
       version: v1alpha3
-      displayName: Standalone
     - displayName: Standalone
       kind: ClusterMaster
       name: clustermasters.enterprise.splunk.com
--- a/pkg/apis/enterprise/v3/common_types.go
+++ b/pkg/apis/enterprise/v3/common_types.go
@@ -313,10 +313,8 @@ type AppSrcDeployInfo struct {
 type BundlePushStageType int
 
 const (
-	// BundlePushNotStarted indicates the initial state of bundle push
-	BundlePushNotStarted BundlePushStageType = iota + 1
 	// BundlePushPending waiting for all the apps to be copied to the Pod
-	BundlePushPending
+	BundlePushPending BundlePushStageType = iota + 1
 	// BundlePushInProgress indicates bundle push in progress
 	BundlePushInProgress
 	// BundlePushComplete bundle push completed
@@ -326,7 +324,7 @@ const (
 // BundlePushTracker used to track the bundle push status
 type BundlePushTracker struct {
 	// Represents the current stage. Internal to the App framework
-	BudlePushStage BundlePushStageType `json:"bundlePushStage,omitempty"`
+	BundlePushStage BundlePushStageType `json:"bundlePushStage,omitempty"`
 
 	// RetryCount defines the number of retries completed so far
 	RetryCount int32 `json:"retryCount,omitempty"`
--- a/pkg/apis/enterprise/v3/common_types.go
+++ b/pkg/apis/enterprise/v3/common_types.go
@@ -313,9 +313,11 @@ type AppSrcDeployInfo struct {
 type BundlePushStageType int
 
 const (
+	// BundlePushNotStarted indicates the initial state of bundle push
+	BundlePushNotStarted BundlePushStageType = iota + 1
 	// BundlePushPending waiting for all the apps to be copied to the Pod
-	BundlePushPending BundlePushStageType = iota + 1
-	// BundlePushInProgress indicates bundle push to complete
+	BundlePushPending
+	// BundlePushInProgress indicates bundle push in progress
 	BundlePushInProgress
 	// BundlePushComplete bundle push completed
 	BundlePushComplete
@@ -325,6 +327,9 @@ const (
 type BundlePushTracker struct {
 	// Represents the current stage. Internal to the App framework
 	BudlePushStage BundlePushStageType `json:"bundlePushStage,omitempty"`
+
+	// RetryCount defines the number of retries completed so far
+	RetryCount int32 `json:"retryCount,omitempty"`
 }
 
 // AppDeploymentContext for storing the Apps deployment information
--- a/pkg/splunk/enterprise/afwscheduler.go
+++ b/pkg/splunk/enterprise/afwscheduler.go
@@ -888,8 +888,17 @@ func isAppInstallationCompleteOnStandaloneReplicas(auxPhaseInfo []enterpriseApi.
 	return true
 }
 
+// isClusterScoped checks whether current cr is a SHC or a CM
+func isClusterScoped(kind string) bool {
+	return kind == "ClusterMaster" || kind == "SearchHeadCluster"
+}
+
 // checkIfBundlePushNeeded confirms if the bundle push is needed or not
-func checkIfBundlePushNeeded(clusterScopedApps []*enterpriseApi.AppDeploymentInfo) bool {
+func checkIfBundlePushNeeded(bundlePushState enterpriseApi.BundlePushStageType, kind string, clusterScopedApps []*enterpriseApi.AppDeploymentInfo) bool {
+	if !isClusterScoped(kind) || bundlePushState == enterpriseApi.BundlePushComplete {
+		return false
+	}
+
 	for _, appDeployInfo := range clusterScopedApps {
 		if appDeployInfo.PhaseInfo.Phase != enterpriseApi.PhasePodCopy || appDeployInfo.PhaseInfo.Status != enterpriseApi.AppPkgPodCopyComplete {
 			return false
@@ -897,6 +906,14 @@ func checkIfBundlePushNeeded(clusterScopedApps []*enterpriseApi.AppDeploymentInf
 	}
 
 	return true
+}
+
+// checkIfBundlePushIsDone checks if the bundle push is done, if there are cluster scoped apps
+func checkIfBundlePushIsDone(kind string, bundlePushState enterpriseApi.BundlePushStageType) bool {
+	if !isClusterScoped(kind) || bundlePushState == enterpriseApi.BundlePushComplete {
+		return true
+	}
+	return false
 }
 
 // initPipelinePhase initializes a given pipeline phase
@@ -1187,9 +1204,10 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
 	// Wait for the yield function to finish
 	afwPipeline.phaseWaiter.Add(1)
 	go func(afwEntryTime int64) {
+		kind := cr.GetObjectKind().GroupVersionKind().Kind
 		for {
 			bundlePushState := getBundlePushState(afwPipeline)
-			if afwEntryTime+maxRunTimeBeforeAttemptingYield < time.Now().Unix() || (afwPipeline.isPipelineEmpty() && bundlePushState == enterpriseApi.BundlePushComplete) {
+			if afwEntryTime+maxRunTimeBeforeAttemptingYield < time.Now().Unix() || (afwPipeline.isPipelineEmpty() && checkIfBundlePushIsDone(kind, bundlePushState)) {
 				scopedLog.Info("Yielding from AFW scheduler", "time elapsed", time.Now().Unix()-afwEntryTime)
 
 				// Trigger termination by closing the channel
@@ -1197,13 +1215,19 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
 				afwPipeline.phaseWaiter.Done()
 				break
 			} else {
-				if bundlePushState != enterpriseApi.BundlePushComplete && checkIfBundlePushNeeded(clusterScopedApps) {
+				if checkIfBundlePushNeeded(bundlePushState, kind, clusterScopedApps) {
 					// Trigger the bundle push playbook: CSPL-1332, CSPL-1333
 
 					podExecClient := splutil.GetPodExecClient(client, cr, targetPodName)
-					kind := cr.GetObjectKind().GroupVersionKind().Kind
-					playBookContext := getPlayBookContext(client, cr, afwPipeline, targetPodName, kind, podExecClient)
 
+					playBookContext := getPlayBookContext(client, cr, afwPipeline, targetPodName, kind, podExecClient)
+					if playBookContext == nil {
+						err = fmt.Errorf("playBookContext is nil")
+						scopedLog.Error(err, "invalid playBookContext")
+						// sleep for one second
+						time.Sleep(1 * time.Second)
+						continue
+					}
 					// run the playbook to issue bundle push command on the pod
 					err = playBookContext.runPlayBook()
 					if err != nil {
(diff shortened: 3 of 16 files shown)


pkg/apis/enterprise/v1/zz_generated.deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
gaurav-splunk, line 1: @pdhanoyasplunk do I need add these back in all the files?

pkg/apis/enterprise/v3/common_types.go
```diff
@@ -325,6 +327,9 @@ const (
 type BundlePushTracker struct {
 	// Represents the current stage. Internal to the App framework
 	BudlePushStage BundlePushStageType `json:"bundlePushStage,omitempty"`
```
smohan-splunk, line 329: Nit: BundlePushStage

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -869,6 +877,131 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
+}
+
+// getIdxcPlayBookContext returns the idxc playbook context
+func getIdxcPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObject, afwPipeline *AppInstallPipeline, podName string) *IdxcPlayBookContext {
+	return &IdxcPlayBookContext{
+		client:        client,
+		cr:            cr,
+		afwPipeline:   afwPipeline,
+		targetPodName: podName,
+		podExecClient: getIdxcPodExecClient(client, cr, applyIdxcBundleCmdStr, podName),
```
sgontla, line 907: nit: do we need to check nil value for podExecClient, to avoid invalid mem references later?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -869,6 +877,131 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
+func (idxcPlayBookContext *IdxcPlayBookContext) isBundlePushComplete() bool {
+	scopedLog := log.WithName("isBundlePushComplete").WithValues("crName", idxcPlayBookContext.cr.GetName(), "namespace", idxcPlayBookContext.cr.GetNamespace())
+
+	stdOut, stdErr, err := idxcPlayBookContext.podExecClient.runPodExecCommand()
+	if err != nil || stdErr != "" {
+		scopedLog.Error(err, "show cluster-bundle-status failed", "stdout", stdOut, "stderr", stdErr)
+		return false
+	}
+
+	if !strings.Contains(stdOut, "cluster_status=None") {
```
sgontla, line 959: nit: instead of contains, is it possible to do a pattern match for `"cluster_status=None"` ?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -869,6 +877,131 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
+	scopedLog.Info("running command to apply cluster bundle")
+	stdOut, stdErr, err := idxcPlayBookContext.podExecClient.runPodExecCommand()
+	if err != nil || stdErr != "OK\n" {
+		err = fmt.Errorf("error while applying cluster bundle. stdout: %s, stderr: %s, err: %v", stdOut, stdErr, err)
+		scopedLog.Error(err, "failed to apply cluster bundle")
+		return err
+	}
+
+	setBundlePushState(afwPipeline, enterpriseApi.BundlePushInProgress)
+	return nil
```
sgontla, line 1002: nit: Since you wrote `isBundlePushComplete`  for checking the status, does it make sense to have another function to trigger the bundle push. 
Also, a switch/if-else-if block may serve better, as it is driven by the bundle push status?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -920,7 +1053,7 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
 			}
 
 			// Track the cluster scoped apps to track the bundle push
-			if appSrc.Scope == enterpriseApi.ScopeCluster && deployInfoList[i].DeployStatus != enterpriseApi.DeployStatusComplete {
+			if appSrc.Scope == enterpriseApi.ScopeCluster && appDeployContext.BundlePushStatus.BudlePushStage != enterpriseApi.BundlePushComplete {
```
sgontla, line 1056: Any reason why we are considering apps that were already installed as well for the tracking?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -961,21 +1094,42 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
 	}
 	scopedLog.Info("List of cluster scoped apps(appName:digest) for this reconcile entry", "apps", clusterAppsList)
 
+	targetPodName := getApplicablePodNameForWorker(cr, 0)
```
sgontla, line 1097: nit: Now that it has other use, probably, we can change the name to something like `getApplicablePodNameForAppFramwork()`?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -961,21 +1094,42 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
+
 	// Wait for the yield function to finish
 	afwPipeline.phaseWaiter.Add(1)
 	go func(afwEntryTime int64) {
 		for {
-			if afwEntryTime+maxRunTimeBeforeAttemptingYield < time.Now().Unix() || afwPipeline.isPipelineEmpty() {
+			if afwEntryTime+maxRunTimeBeforeAttemptingYield < time.Now().Unix() || (afwPipeline.isPipelineEmpty() && getBundlePushState(afwPipeline) != enterpriseApi.BundlePushPending) {
```
sgontla, line 1103: Doesn't the state check need to be `BundlePushComplete`?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -961,21 +1094,42 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
+					if afwPipeline.playBookContext == nil {
+						afwPipeline.playBookContext = getPlayBookContext(client, cr, afwPipeline, targetPodName, kind)
+					}
+					// run the playbook to issue bundle push command on the pod
+					err = afwPipeline.playBookContext.runPlayBook()
+					if err != nil {
+						// If the error is due to a bundle which is already present, don't do anything.
+						// In the next reconcile we will mark it as bundle push complete
+						if strings.Contains(err.Error(), idxcBundleAlreadyPresentStr) {
+							continue
```
sgontla, line 1124: nit: the continue will miss the sleep, but hope it's ok for now, as this code move to the install handler.

pkg/splunk/enterprise/util.go
```diff
@@ -314,6 +314,41 @@ func appPhaseStatusAsStr(status enterpriseApi.AppPhaseStatusType) string {
 	}
 }
 
+// bundlePushStateAsStr converts the bundle push state enum to corresponding string
+func bundlePushStateAsStr(state enterpriseApi.BundlePushStageType) string {
+	switch state {
+	case enterpriseApi.BundlePushNotStarted:
+		return "Bundle Push Not Started"
```
sgontla, line 321: Do we need this state, as we can jump to the Pending, oncee we detect cluster scoped apps?

pkg/splunk/enterprise/util.go
```diff
@@ -314,6 +314,41 @@ func appPhaseStatusAsStr(status enterpriseApi.AppPhaseStatusType) string {
...
+}
+
+// setBundlePushState sets the bundle push state to the new state
+func setBundlePushState(afwPipeline *AppInstallPipeline, state enterpriseApi.BundlePushStageType) {
+	scopedLog := log.WithName("setBundlePushState")
+
+	afwPipeline.pplnMutex.Lock()
+	scopedLog.Info("Setting the bundle push state", "old state", bundlePushStateAsStr(afwPipeline.appDeployContext.BundlePushStatus.BudlePushStage), "new state", bundlePushStateAsStr(state))
+	afwPipeline.appDeployContext.BundlePushStatus.BudlePushStage = state
+	afwPipeline.pplnMutex.Unlock()
```
sgontla, line 340: A lock is not required, as all the bundle push to be handled on on thread?

pkg/splunk/enterprise/util.go
```diff
@@ -314,6 +314,41 @@ func appPhaseStatusAsStr(status enterpriseApi.AppPhaseStatusType) string {
...
+}
+
+// getBundlePushState returns the current bundle push state
+func getBundlePushState(afwPipeline *AppInstallPipeline) enterpriseApi.BundlePushStageType {
+	var state enterpriseApi.BundlePushStageType
+	afwPipeline.pplnMutex.Lock()
+	state = afwPipeline.appDeployContext.BundlePushStatus.BudlePushStage
+	afwPipeline.pplnMutex.Unlock()
+	return state
```
sgontla, line 349: same as above?

pkg/splunk/enterprise/util.go
```diff
@@ -896,9 +931,14 @@ func handleAppRepoChanges(client splcommon.ControllerClient, cr splcommon.MetaOb
...
+		scope := getAppSrcScope(appFrameworkConfig, appSrc)
 		// 2.2 Check for any App changes(Ex. A new App source, a new App added/updated)
-		if AddOrUpdateAppSrcDeploymentInfoList(&appSrcDeploymentInfo, s3Response.Objects) {
-			appsModified = true
+		appsModified = AddOrUpdateAppSrcDeploymentInfoList(&appSrcDeploymentInfo, s3Response.Objects)
+
+		// if some apps were modified or added, and we have cluster scoped apps,
+		// then initialize the bundle push state
+		if appsModified && scope == enterpriseApi.ScopeCluster {
+			appDeployContext.BundlePushStatus.BudlePushStage = enterpriseApi.BundlePushNotStarted
```
sgontla, line 941: I feel this is redundant, as we are already preparing all the `cluster` scoped apps in the scheduler, and that is where we can set the stage to pending directly.

pkg/splunk/enterprise/util.go
```diff
@@ -1578,3 +1618,59 @@ func validateMonitoringConsoleRef(c splcommon.ControllerClient, revised *appsv1.
...
+		return false
+	}
+
+	for _, app := range appsList {
+		if app.PhaseInfo.Status != enterpriseApi.AppPkgPodCopyComplete {
+			return false
+		}
+	}
+	return true
+}
```
sgontla, line 1659: As mentioned in my earlier comment, we have this info already with the cluster scoped apps tracking in place.

deploy/olm-catalog/splunk/1.0.3/splunk.v1.0.3.clusterserviceversion.yaml
```diff
@@ -120,7 +120,8 @@ spec:
       - kind: Secrets
         version: v1
       version: v1beta1
-    - kind: ClusterMaster
+    - displayName: Standalone
```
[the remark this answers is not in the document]
gaurav-splunk: Yeah not sure how this changed :/

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -716,6 +716,14 @@ podCopyPhase:
 
 		time.Sleep(200 * time.Millisecond)
 	}
+
+	// check if all the cluster scoped apps have been copied to the pod
```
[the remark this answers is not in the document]
gaurav-splunk: Yes, we will only initiate the bundle push once all the cluster scoped apps are copied.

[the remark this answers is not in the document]
smohan-splunk: I mean, `podCopyPhaseManager()` is called for both cluster & local correct? the checked you have added is intended for cluster scoped apps but the code itself will execute for local too?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -961,21 +1094,42 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
 	}
 	scopedLog.Info("List of cluster scoped apps(appName:digest) for this reconcile entry", "apps", clusterAppsList)
 
+	targetPodName := getApplicablePodNameForWorker(cr, 0)
+
 	// Wait for the yield function to finish
 	afwPipeline.phaseWaiter.Add(1)
 	go func(afwEntryTime int64) {
 		for {
```
smohan-splunk, line 1102: Nit: may be read `getBundlePushState` once and use it in the if/else conditions

smohan-splunk (approved):

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -869,6 +877,131 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
+
+	appDeployContext := idxcPlayBookContext.afwPipeline.appDeployContext
+
+	// if the bundle push is already in progress, check the status
+	if appDeployContext.BundlePushStatus.BudlePushStage == enterpriseApi.BundlePushInProgress {
+		podExecClient := idxcPlayBookContext.podExecClient.(*IdxcPodExecClient)
+
+		// change the command string to run the show cluster-bundle-status command for idxc
+		podExecClient.command = fmt.Sprintf(idxcShowClusterBundleStatusStr, podExecClient.adminPwd)
+		if idxcPlayBookContext.isBundlePushComplete() {
```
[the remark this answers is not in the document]
gaurav-splunk: I wanted to do that but this messes up the test case and crashes at line 980 when we try to use our mock podExecClient. Hence I had to do it this way.

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: Yeah I will change to have apps not installed yet.

pkg/splunk/enterprise/types.go
```diff
@@ -113,6 +113,58 @@ type AppInstallPipeline struct {
 
 	// Refernce to app deploy context
 	appDeployContext *enterpriseApi.AppDeploymentContext
+
+	// reference to PlayBookImpl to run a specific playbook
+	playBookContext PlayBookImpl
```
sgontla, line 118: This is something I think I missed earlier, do we need this? I think it brings confusion when we have multiple cluster scoped playbooks. Ex. SHC, ES, ITSI, and additional logic to resolve the same.  On every reconcile entry, the logic has to figure out and prepare the correct playbook context to run the correct playbook, and already all those functions are being methods with those contexts as receivers, so, it is simple just to call them. We can discuss and finalize if we need this.

pkg/splunk/enterprise/util.go
```diff
@@ -1615,3 +1646,42 @@ func validateMonitoringConsoleRef(c splcommon.ControllerClient, revised *appsv1.
...
+		deployInfoList := appSrcDeployInfo.AppDeploymentInfoList
+		for i := range deployInfoList {
+			appSrc, err := getAppSrcSpec(appDeployContext.AppFrameworkConfig.AppSources, appSrcName)
+			if err != nil {
+				// Error, should never happen
+				scopedLog.Error(err, "Unable to find App src", "App src name", appSrcName, "App name", deployInfoList[i].AppName)
+				continue
+			}
+
+			if appSrc.Scope == enterpriseApi.ScopeCluster {
```
sgontla, line 1664: Since the mapping is `appSrcName` to `deployInfoList`, we can do the scope check above line# 1655

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -981,6 +981,143 @@ func afwGetReleventStatefulsetByKind(cr splcommon.MetaObject, client splcommon.C
...
+
+// runPodExecCommand runs the commands related to idxc bundle push
+func (podExecClient *PodExecClient) runPodExecCommand() (string, string, error) {
+	if podExecClient == nil {
+		err := fmt.Errorf("podExecClient is nil")
+		return "", "", err
+	}
+
+	streamOptions := &remotecommand.StreamOptions{
+		Stdin: strings.NewReader(podExecClient.command),
```
sgontla, line 1047: Can we make the command as input to this method, that way, the user has the option to re-use the same client with multiple commands. One of my earlier comments was to pass the bundle push command as part of the `isBundlePushComplete`. So, the caller fo the APIs need not worry about setting the command.

sgontla (changes requested):

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: I get your point but I think for the exact reason that users can re-use this API with different command/s, I embedded the command in the client itself, they can just change the command in the client and can re-use it. In either case we are relying on the caller of this API to set the command to be called in the podExec command.
Also, about your second point of passing the command in the `isBundlePushComplete` API, as I mentioned before too, I could not pass it inside the API since we require admin password to create the whole command string. Now, we need this sort of conversion i.e. `podExecClient := idxcPlayBookContext.podExecClient.(*PodExecClient)` to convert the generic `PodExecClientImpl` to `PodExecClient` so that we can access the `adminPwd` field. And if you are trying to use `mockPodExecClient` in UT, it will crash at the point of conversion to `PodExecClient`. Hence to cover the lines in `isBundlePushComplete`, I had to set the command before calling the podExec API. But let's discuss it more offline and then maybe we can reach a common ground :)

gaurav-splunk: NVM, I edited my previous comment, I got what you are saying and will change the code accordingly!

sgontla (approved):

PR #607, opened 2021-12-01
CSPL-1333: Implement playbook for SHC

gaurav-splunk:
This PR targets the following things - 
1. Implement a playbook for installing apps for SHC
2. Fix bug/s found as part of CSPL-1515 while downloading apps.

### DIFF
--- a/pkg/splunk/enterprise/afwscheduler.go
+++ b/pkg/splunk/enterprise/afwscheduler.go
@@ -1029,7 +1029,8 @@ func getPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObje
 }
 
 // triggerBundlePush triggers the bundle push operation for SHC
-func (shcPlayBookContext *SHCPlayBookContext) triggerBundlePush(cmd string) error {
+func (shcPlayBookContext *SHCPlayBookContext) triggerBundlePush() error {
+	cmd := fmt.Sprintf(applySHCBundleCmdStr, shcPlayBookContext.searchHeadCaptainURL)
 	stdOut, stdErr, err := shcPlayBookContext.podExecClient.RunPodExecCommand(cmd)
 	if err != nil || stdErr != "" {
 		err = fmt.Errorf("error while applying cluster bundle. stdout: %s, stderr: %s, err: %v", stdOut, stdErr, err)
@@ -1055,8 +1056,7 @@ func (shcPlayBookContext *SHCPlayBookContext) runPlayBook() error {
 		// run the command to apply cluster bundle
 		scopedLog.Info("running command to apply cluster bundle")
 
-		bundlePushCmd := fmt.Sprintf(applySHCBundleCmdStr, shcPlayBookContext.searchHeadCaptainURL)
-		err := shcPlayBookContext.triggerBundlePush(bundlePushCmd)
+		err := shcPlayBookContext.triggerBundlePush()
 		if err != nil {
 			scopedLog.Error(err, "failed to apply cluster bundle")
 			shcPlayBookContext.afwPipeline.appDeployContext.BundlePushStatus.RetryCount++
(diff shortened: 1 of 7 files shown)


pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -343,6 +343,13 @@ func (downloadWorker *PipelineWorker) download(pplnPhase *PipelinePhase, s3Clien
 	err = s3ClientMgr.DownloadApp(remoteFile, localFile, appDeployInfo.ObjectHash)
 	if err != nil {
 		scopedLog.Error(err, "unable to download app", "appName", appName)
+
+		// remove the local file
+		err = os.RemoveAll(localFile)
```
gaurav-splunk, line 348: This is one of the bugs found as part of CSPL-1515 where if we are not able to download the app/s due to any reason, we would be left with stale 0 sized files on the operator pod. And in the next reconcile we would just check that the file is present on operator pod(without checking the size) and would move to podCopy and then to install phase. Then at the install phase we would fail to install the app/s as it is of size 0. Hence we need to remove the file on the operator pod.

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -173,17 +173,18 @@ func ApplySearchHeadCluster(client splcommon.ControllerClient, cr *enterpriseApi
...
-		// Schedule one more reconcile in next 5 seconds, just to cover any latest app framework config changes
-		if cr.Status.AppContext.IsDeploymentInProgress {
-			cr.Status.AppContext.IsDeploymentInProgress = false
-			return result, nil
-		}
-	}
-
 	// no need to requeue if everything is ready
 	if cr.Status.Phase == splcommon.PhaseReady {
+		if cr.Status.AppContext.AppsSrcDeployStatus != nil && cr.Status.DeployerPhase == splcommon.PhaseReady {
```
gaurav-splunk, line 178: This is put here because we should only push the bundle once both Deployer and SHC members are ready or else we would keep trying and fail to push the bundle until the SHC members come up.

pkg/splunk/enterprise/util.go
```diff
@@ -396,7 +396,7 @@ func getRemoteObjectKey(cr splcommon.MetaObject, appFrameworkConfig *enterpriseA
 	volumePath = volumePath[index+1:]
 	location := appSrc.Location
 
-	remoteObjectKey = volumePath + location + appName
+	remoteObjectKey = filepath.Join(volumePath, location, appName)
```
gaurav-splunk, line 399: This is another fix as part of CSPL-1515 where if we miss `/` in the volume `path` and/or in the `location` in the appSources, we were not creating correct remoteObjectKey, which would eventually result in download failure.

pkg/splunk/enterprise/searchheadcluster.go
sgontla: Can we move this check to the playbook, such that the local installs are not blocked? Probably, just exiting the playbook.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1012,14 +1021,66 @@ func getPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObje
...
+	}
+	return nil
+}
+
+// runPlayBook will implement the bundle push logic for SHC
+func (shcPlayBookContext *SHCPlayBookContext) runPlayBook() error {
+	scopedLog := log.WithName("runPlayBook").WithValues("crName", shcPlayBookContext.cr.GetName(), "namespace", shcPlayBookContext.cr.GetNamespace())
+
+	cr := shcPlayBookContext.cr.(*enterpriseApi.SearchHeadCluster)
+	if cr.Status.Phase != splcommon.PhaseReady {
```
gaurav-splunk, line 1046: Added the check here so that we dont call the playBook unless SHC is ready

sgontla (approved):

smohan-splunk (approved):

pkg/splunk/enterprise/names.go
```diff
@@ -70,6 +70,8 @@ const (
 
 	manualAppUpdateCMStr = "splunk-%s-manual-app-update"
 
+	applySHCBundleCmdStr = "/opt/splunk/bin/splunk apply shcluster-bundle -target https://%s:8089 -auth admin:`cat /mnt/splunk-secrets/password` --answer-yes -push-default-apps true"
```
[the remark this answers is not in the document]
gaurav-splunk: As discussed, as of now we are hardcoding it everywhere in the code. Hence I'll leave it as is for now. Maybe we can take it up as part of some future code refactoring.

smohan-splunk (approved):

PR #615, opened 2021-12-09
CSPL-1328: Handle long running SHC bundle push

gaurav-splunk:


pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1028,12 +1028,28 @@ func getPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObje
...
+func (shcPlayBookContext *SHCPlayBookContext) isBundlePushComplete() bool {
+	scopedLog := log.WithName("isBundlePushComplete").WithValues("crName", shcPlayBookContext.cr.GetName(), "namespace", shcPlayBookContext.cr.GetNamespace())
+
+	cmd := fmt.Sprintf("cat %s", shcBundlePushStatusCheckFile)
+	// check the content of the status file
+	stdOut, stdErr, err := shcPlayBookContext.podExecClient.RunPodExecCommand(cmd)
+	if err != nil || stdErr != "" || !strings.Contains(stdOut, shcBundlePushCompleteStr) {
+		scopedLog.Error(err, "checking the status of SHC Bundle Push failed", "stdout", stdOut, "stderr", stdErr)
+		return false
+	}
```
smohan-splunk, line 1041: As discussed:

* The file should be cleaned up after the push command return contents are written to it & have been checked
* The contents should also be read for any errors too, so that the push can be retried in subsequent reconciles

smohan-splunk (approved):

pkg/splunk/enterprise/names.go
```diff
@@ -70,7 +70,11 @@ const (
 
 	manualAppUpdateCMStr = "splunk-%s-manual-app-update"
 
-	applySHCBundleCmdStr = "/opt/splunk/bin/splunk apply shcluster-bundle -target https://%s:8089 -auth admin:`cat /mnt/splunk-secrets/password` --answer-yes -push-default-apps true"
+	applySHCBundleCmdStr = "/opt/splunk/bin/splunk apply shcluster-bundle -target https://%s:8089 -auth admin:`cat /mnt/splunk-secrets/password` --answer-yes -push-default-apps true &> %s &"
+
+	shcBundlePushCompleteStr = "Bundle has been pushed successfully to all the cluster members.\n"
+
+	shcBundlePushStatusCheckFile = "/opt/splunk/etc/shcluster_bundle_status.txt"
```
sgontla, line 77: Can we use a location somewhere from the /init-apps/, that way to avoid broader visibility of this file?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1052,27 +1120,40 @@ func (shcPlayBookContext *SHCPlayBookContext) runPlayBook() error {
...
+			setBundlePushState(shcPlayBookContext.afwPipeline, enterpriseApi.BundlePushComplete)
+
+			// reset the retry count
+			shcPlayBookContext.afwPipeline.appDeployContext.BundlePushStatus.RetryCount = 0
+
+			// set the state to install complete for all the cluster scoped apps
+			setInstallStateForClusterScopedApps(appDeployContext, enterpriseApi.AppPkgInstallComplete)
+		} else if err != nil {
+			scopedLog.Error(err, "there was an error in SHC bundle push, will retry again.")
+			return err
```
sgontla, line 1139: nit: `return` is redundant. Either if-else-if or explicit returns will avoid confusion.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1028,12 +1028,80 @@ func getPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObje
...
+	// check if there is an error returned from running the pod exec command
+	if err != nil || stdErr != "" {
+		scopedLog.Error(err, "checking the status of SHC Bundle Push failed", "stdout", stdOut, "stderr", stdErr)
+		// reset the bundle push state to Pending, so that we retry again.
+		setBundlePushState(shcPlayBookContext.afwPipeline, enterpriseApi.BundlePushPending)
+
+		// remove the status file too, so that we dont have any stale status
+		removeErr := shcPlayBookContext.removeSHCBundlePushStatusFile()
+		if removeErr != nil {
+			err = removeErr
```
sgontla, line 1061: Do we need to append the existing error? If not, we are overriding it(?)

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1028,12 +1028,80 @@ func getPlayBookContext(client splcommon.ControllerClient, cr splcommon.MetaObje
...
+		err = fmt.Errorf("there was an error in applying SHC Bundle, err=\"%v\"", stdOut)
+		scopedLog.Error(err, "SHC Bundle push status file reported an error while applying bundle")
+
+		// reset the bundle push state to Pending, so that we retry again.
+		setBundlePushState(shcPlayBookContext.afwPipeline, enterpriseApi.BundlePushPending)
+
+		// remove the status file too, so that we dont have any stale status
+		removeErr := shcPlayBookContext.removeSHCBundlePushStatusFile()
+		if removeErr != nil {
+			err = removeErr
```
sgontla, line 1083: nit: same here.

pkg/splunk/enterprise/names.go
gaurav-splunk: I did not use that since in the future we were going to get rid of that location or at least change the name of that mount point.

sgontla: We better keep the Operator context out of the ../etc and ../var partitions. Otherwise, if Ansible/Docker/Operator keep adding, it will be kind of invading :-).  /init-apps/ is in our control and if it changes, we have to take care of this . Probably, we can offset from a global variable.

sgontla (approved):

PR #634, opened 2021-12-21
CSPL-1169: Apps should be correctly installed on standalone pods when scaling up/down

gaurav-splunk:
**Description**
In case of Standalone CR type, during scaling up, the new pods would not have the already installed apps that were installed on the existing pods.

**Solution**
During reconcile, at the moment when we decide if we are scaling up/down, update the AuxPhaseInfo for standalone CR, so that when afw scheduler runs, it knows to install apps for these new pods.

### DIFF
--- a/pkg/splunk/controller/statefulset.go
+++ b/pkg/splunk/controller/statefulset.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 
+	enterpriseApi "github.com/splunk/splunk-operator/pkg/apis/enterprise/v3"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


pkg/splunk/controller/statefulset.go
```diff
@@ -317,16 +317,22 @@ func isCurrentCROwner(cr splcommon.MetaObject, currentOwners []metav1.OwnerRefer
 	return reflect.DeepEqual(currentOwners[0], splcommon.AsOwner(cr, false))
 }
 
-// IsStatefulSetScalingUp checks if we are currently scaling up
-func IsStatefulSetScalingUp(client splcommon.ControllerClient, cr splcommon.MetaObject, name string, desiredReplicas int32) (bool, error) {
+// IsStatefulSetScalingUpOrDown checks if we are currently scaling up or down
+func IsStatefulSetScalingUpOrDown(client splcommon.ControllerClient, cr splcommon.MetaObject, name string, desiredReplicas int32) (bool, bool, error) {
```
smohan-splunk, line 321: As discussed can we use a enum to indicate the Scale up/down rather than 2 bools)

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -514,6 +518,36 @@ downloadPhase:
 	}
 }
 
+// isAppAlreadyInstalled checks whether app is already installed on the pod
+func (ctx *localScopePlaybookContext) isAppAlreadyInstalled(appInstallStatusFilePath string) bool {
```
smohan-splunk, line 522: As discussed, ideally this check shouldn't be needed for existing pods. To be checked under what exact scenario this is needed?

pkg/splunk/enterprise/standalone.go
```diff
@@ -140,22 +140,24 @@ func ApplyStandalone(client splcommon.ControllerClient, cr *enterpriseApi.Standa
...
 			// for all the app sources so that we add all the app sources in configMap.
 			cr.Status.AppContext.IsDeploymentInProgress = true
-			appStatusContext := cr.Status.AppContext
+
 			for appSrc := range appStatusContext.AppsSrcDeployStatus {
 				changeAppSrcDeployInfoStatus(appSrc, appStatusContext.AppsSrcDeployStatus, enterpriseApi.RepoStateActive, enterpriseApi.DeployStatusComplete, enterpriseApi.DeployStatusPending)
+				changePhaseInfo(cr.Spec.Replicas, appSrc, appStatusContext.AppsSrcDeployStatus)
 			}
 
 			// Now apply the configMap will full app listing.
```
smohan-splunk, line 163: Lets remove `ApplyAppListingConfigMap`

pkg/splunk/enterprise/util.go
```diff
@@ -886,6 +886,68 @@ func checkIfAppSrcExistsWithRemoteListing(appSrc string, remoteObjListingMap map
...
+	}
+}
+
+func removeStaleEntriesFromAuxPhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus map[string]enterpriseApi.AppSrcDeployInfo) {
+	scopedLog := log.WithName("changePhaseInfo")
+
+	if appSrcDeploymentInfo, ok := appSrcDeployStatus[appSrc]; ok {
+		appDeployInfoList := appSrcDeploymentInfo.AppDeploymentInfoList
+		for idx := range appDeployInfoList {
+			if len(appDeployInfoList[idx].AuxPhaseInfo) != 0 {
```
sgontla, line 939: Ni: Just to be on safe side, it may be a good check to see the length of the slice is greater than `desiredReplicas`

sgontla (approved):

smohan-splunk (approved):

PR #640, opened 2022-01-05
CSPL-1529: Fix a bug where ClusterMaster was not watching for changes in configMap

gaurav-splunk:
**Problem Statement**
When we deploy a ClusterMaster along with any other CR such as SHC or Standalone with manual app update enabled(OR in other words, polling disabled), manual app update was not getting triggered when we changed the configMap to poll for the changes on the remote storage.

**Root Cause**
Because of a recent PR for merge from develop to feature branch, we reverted to the old code where ClusterMaster would not watch for changes in configMap. Because of this bug, any change in the manual app update configMap was not getting picked up by ClusterMaster.

**Solution**
1. Add configMap as a watch type to ClusterMaster CR.
2. Also enable in-tests related to manual app update in test/ folder.

### DIFF
--- a/pkg/controller/add_clustermaster.go
+++ b/pkg/controller/add_clustermaster.go
@@ -49,7 +49,7 @@ func (ctrl ClusterManagerController) GetInstance() splcommon.MetaObject {
 
 // GetWatchTypes returns a list of types owned by the controller that it would like to receive watch events for
 func (ctrl ClusterManagerController) GetWatchTypes() []runtime.Object {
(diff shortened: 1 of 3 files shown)


smohan-splunk (approved):
The PR which introduced the regression: https://github.com/splunk/splunk-operator/pull/578

sgontla (approved):

smohan-splunk:
Int tests have been locally verified

bendermar:
For me it seems that the bug is again present in current version 2.1.1? I´ve tested on a new environment. The config map gets updated but changes not gets written to cluster manager.

PR #649, opened 2022-01-12
CSPL-1570: Fix a bug where if we scale standalone from 1 to any greater replica number, we were setting incorrect AuxPhase for replica 0

gaurav-splunk:
**Problem**
When trying to scale a standalone CR from 1 to any greater number of replicas, we would try to reinstall the app/s on pod-0, which would result in installation failure and then setting the wrong AuxPhaseInfo for that pod.

**Solution**
Set AuxPhaseInfo for pod-0 to be Phase = PhaseInstall and Status = AppPkgInstallComplete. This way we wont be attempting to install the app/s again on that pod.

### DIFF
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -886,23 +886,31 @@ func checkIfAppSrcExistsWithRemoteListing(appSrc string, remoteObjListingMap map
 	return false
 }
 
-// updateAuxPhaseInfo updates the AuxPhaseInfo
-func updateAuxPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, desiredReplicas int32) {
+// resetAuxPhaseInfo resets the AuxPhaseInfo slices with desired phase and status
+func resetAuxPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, replicas int32, phase enterpriseApi.AppPhaseType, status enterpriseApi.AppPhaseStatusType) {
+	for i := 0; i < int(replicas); i++ {
+		appDeployInfo.AuxPhaseInfo[i].Phase = phase
+		appDeployInfo.AuxPhaseInfo[i].Status = status
+	}
+}
+
+// createAuxPhaseInfoForNewPods appends the new AuxPhaseInfo for new pods
+func createAuxPhaseInfoForNewPods(appDeployInfo *enterpriseApi.AppDeploymentInfo, desiredReplicas int32, phase enterpriseApi.AppPhaseType, status enterpriseApi.AppPhaseStatusType) {
 
 	auxPhaseInfoLen := len(appDeployInfo.AuxPhaseInfo)
 
 	for i := auxPhaseInfoLen; i < int(desiredReplicas); i++ {
 		phaseInfo := enterpriseApi.PhaseInfo{
-			Phase:      enterpriseApi.PhasePodCopy,
-			Status:     enterpriseApi.AppPkgPodCopyPending,
+			Phase:      phase,
+			Status:     status,
 			RetryCount: 0,
 		}
 		appDeployInfo.AuxPhaseInfo = append(appDeployInfo.AuxPhaseInfo, phaseInfo)
 	}
 }
 
-// changePhaseInfo changes PhaseInfo and AuxPhaseInfo for each app to desired state
-func changePhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus map[string]enterpriseApi.AppSrcDeployInfo) {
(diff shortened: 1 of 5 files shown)


sgontla:
Can we also cover some UT for the combinations of PhaseInfo values?

pkg/splunk/enterprise/util.go
```diff
@@ -913,16 +913,23 @@ func changePhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus ma
 				continue
 			}
 
+			// CSPL-1570: If this is the first time we are scaling up from replicas=1, then set the AuxPhaseInfo to Install complete.
```
smohan-splunk, line 916: General question - why don't we always keep the `AuxPhaseInfo` up-to-date, irrespective of whether or not we are scaling up? If we have this already in place, we could just skip over the replicas which already have the status as `AppPkgInstallComplete` ?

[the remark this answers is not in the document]
gaurav-splunk: This is not an issue when scaling up from replicas > 1 because we will already have correct AuxPhaseInfo(with status as `AppPkgInstallComplete`) for both replicas the first time when afw scheduler runs. At the time of scaling from, let's say from 2 to 3, we will only update the AuxPhaseInfo for pod-2 and only install the app/s on that pod.

gaurav-splunk: As I have replied below in other comment, `AuxPhaseInfo` is always up-to-date. This was a special and only case left that was yet to be handled.

pkg/splunk/enterprise/util.go
```diff
@@ -913,16 +913,26 @@ func changePhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus ma
 				continue
 			}
 
+			// CSPL-1570: Set the AuxPhaseInfo to Install complete for all the existing pods.
+			// This is done so that we don't try to reinstall the app again on the existing pod.
+			if len(appDeployInfoList[idx].AuxPhaseInfo) == 0 &&
```
sgontla, line 918: To make the scale-up completely decoupled from the scheduler, Please remove this check as well. Then, at line 922, we can reset the slice, OR, update the existing slice. 

Also, looking at the overall flow, one another case is(rare case, scale-up, when in the middle of installation),  if the Main phase info is not Install complete:
1. Do not alter the existing auxPhaseInfo
2. Still set the PodCopy phase for new pod entries.


I feel, with these changes, it will be independent, and self-sufficient with respect to the scale-up scenario.  With that, the scheduler will get the calls in 3 different flows, and all of them are handled in the same way, and no special handle for any of these cases. i.e (1. First-time entry, once after learning new apps, 2. Re-entry across reconciles, 3. scale-up case.)

pkg/splunk/enterprise/util.go
```diff
@@ -902,7 +902,7 @@ func updateAuxPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, desiredR
 }
 
 // changePhaseInfo changes PhaseInfo and AuxPhaseInfo for each app to desired state
-func changePhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus map[string]enterpriseApi.AppSrcDeployInfo) {
+func changePhaseInfo(currentReplicas, desiredReplicas int32, appSrc string, appSrcDeployStatus map[string]enterpriseApi.AppSrcDeployInfo) {
```
sgontla, line 905: Nit: You may want to name it as `adjustAfwPhaseInfoForScaleup()` or something like that.

pkg/splunk/enterprise/util.go
```diff
@@ -913,16 +913,26 @@ func changePhaseInfo(desiredReplicas int32, appSrc string, appSrcDeployStatus ma
...
 
 			// set the status to download pending
 			appDeployInfoList[idx].PhaseInfo.Status = enterpriseApi.AppPkgDownloadPending
 
-			if len(appDeployInfoList[idx].AuxPhaseInfo) != 0 {
-				// update the aux phase info
-				updateAuxPhaseInfo(&appDeployInfoList[idx], desiredReplicas)
-			}
+			// update the aux phase info
+			updateAuxPhaseInfo(&appDeployInfoList[idx], desiredReplicas)
```
sgontla, line 934: Nit: you may want to rename to something like `createAuxPhaseInfoForNewPods() ` or something similar.

pkg/splunk/enterprise/util.go
gaurav-splunk: Done

sgontla (approved):

gaurav-splunk:
Closing this PR since this is not needed and the fix for this bug is being implemented in another PR as part of some other miscellaneous improvemens.

PR #660, opened 2022-01-21
fix: For Minio connections trust admin to pick protocol

rfaircloth-splunk:
When using minio inside of a k8s cluster or in a network with other mitigations such as a cloud vpc with host to host network encryption use of TLS may be undesirable (double encryption) and can limit observability.


smohan-splunk (approved):
Has been reviewed by security team. As a followup task, a new test case for cert validation to be introduced for https

PR #662, opened 2022-01-24
CSPL-1552: Update PodExecClient and MockPodExecClient to be used across the code

gaurav-splunk:
This PR consists of the following improvements - 
1. Implement a new **MockPodExecClient** which is used to handle different types of commands and their return values.
```
// MockPodExecClient mocks the PodExecClient
type MockPodExecClient struct {
	client             splcommon.ControllerClient
	cr                 splcommon.MetaObject
	targetPodName      string
	WantCmdList        []string
	GotCmdList         []string
	MockReturnContexts map[string]*MockPodExecReturnContext
}
```

2. Create **MockPodExecReturnContext** to have fields which represents return values of a PodExecClient - 
 
```
// MockPodExecReturnContext stores the return values for each podExec command
type MockPodExecReturnContext struct {
	StdOut string
	StdErr string
	Err    error
}
```
2. Update PodExecClient.RunPodExecCommand() to take `*remotecommand.StreamOptions` and  `[]string` as arguments so that we can provide custom commands and base commands(for e.g. "/bin/sh" or "tar -xf..") as per our requirements.
2. Pass in MockPodExecClient in the APIs from UTs for better code coverage. As a result of this change, we have removed the dependency of passing in `mock` boolean as a parameter to differentiate the code flow for UT v/s actual code.
3. Updated description of `appsRepoPollIntervalSecs` as part of CSPL-1572.
4. Hopefully this also improves the code coverage a bit :)


deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -665,8 +665,8 @@ spec:
                       App changes. The default value for this config is 1 hour(3600
                       sec), minimum value is 1 minute(60sec) and maximum value is
                       1 day(86400 sec). We assign the value based on following conditions
-                      -    1. If no value or 0 is specified then it will be defaulted
-                      to 1 hour.    2. If anything less than min is specified then
+                      -    1. If no value or 0 is specified then it means polling
```
sgontla, line 668: nit: `polling is disabled` --> `periodic polling is disabled` ?

pkg/splunk/test/util.go
```diff
@@ -42,3 +48,104 @@ func GetMockS3SecretKeys(name string) corev1.Secret {
...
+	StdErr string
+	Err    error
+}
+
+// MockPodExecClientHandler handles the MockPodExecClient
+type MockPodExecClientHandler struct {
+	WantMockPodExecClient []*MockPodExecClient
+	GotMockPodExecClient  []*MockPodExecClient
+	MockClients           map[string]*MockPodExecClient
+}
```
sgontla, line 64: Our original idea was, to have one mock client that takes a list of commands/patterns, and serves the required return values, also, support for the validation for those commands.  Just trying to understand why we need to go with multiple clients here.

pkg/splunk/test/util.go
```diff
@@ -42,3 +48,104 @@ func GetMockS3SecretKeys(name string) corev1.Secret {
...
+		// use a dummy mockPodExecClient
+		mockPodExecClient = &MockPodExecClient{}
+	}
+	// check if the mockPodExecClient is already added or not in the list of mockPodExecClients
+	var found bool
+	for i := range handler.GotMockPodExecClient {
+		if reflect.DeepEqual(handler.GotMockPodExecClient[i], mockPodExecClient) {
+			found = true
+			break
+		}
```
sgontla, line 136: Here, thought to compare the commands, OR command patterns.

pkg/splunk/test/util.go
```diff
@@ -42,3 +49,107 @@ func GetMockS3SecretKeys(name string) corev1.Secret {
...
+	var found bool
+	for i := range client.GotCmdList {
+		if reflect.DeepEqual(client.GotCmdList[i], command) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		client.GotCmdList = append(client.GotCmdList, command)
+	}
```
sgontla, line 142: Are we not allowing a command multiple times in the same API?

pkg/splunk/test/util.go
```diff
@@ -42,3 +49,107 @@ func GetMockS3SecretKeys(name string) corev1.Secret {
...
+
+	return mockPodExecReturnContext.StdOut, mockPodExecReturnContext.StdErr, mockPodExecReturnContext.Err
```
sgontla, line 144: Does this mock logic works only if the base command is `"/bin/sh"`?

pkg/splunk/test/util.go
```diff
@@ -42,3 +49,107 @@ func GetMockS3SecretKeys(name string) corev1.Secret {
...
+		if mockPodExecReturnContext == nil {
+			err := fmt.Errorf("mockPodExecReturnContext is nil")
+			return "", "", err
+		}
+	}
+
+	// check if the command is already added or not in the list of GotCmdList
+	var found bool
+	for i := range client.GotCmdList {
+		if reflect.DeepEqual(client.GotCmdList[i], command) {
```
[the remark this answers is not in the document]
gaurav-splunk: Yeah I will change it.

pkg/splunk/test/util.go
gaurav-splunk: The command can run multiple times in the same API, this will just keep track of all the different commands that run that API. This is helpful because you can run an API multiple times to test the negative and positive scenarios, and we dont want to append the same commands again and again.

pkg/splunk/test/util.go
gaurav-splunk: The logic from line 118-129 works only for command "bin/sh" at the moment and the reason is explained in the comment section. Everything else is common.

sgontla (approved):

smohan-splunk (approved):

PR #671, opened 2022-02-04
CSPL-1460: Phase-2 Code cleanup - Remove app listing configMap and init-container creation

gaurav-splunk:
As part of Phase-2 Code cleanup in this PR, we are focussing on the following things - 
1. Removing all the traces of app listing configMap
2. Removing creation of init container for app downloads on Splunk pod.
3. Changing the name of /init-apps/ volume.
4. Updating the test cases to reflect the above change

PR #672, opened 2022-02-07
CSPL-1579: Temp AppFramework download location on Operator pod

sgontla:
When the Operator is not configured with an explicit volume mount,
use the '/tmp/appframework/' from main memory as the temporary staging area for the app package downloads.

### DIFF
--- a/docs/AppFramework.md
+++ b/docs/AppFramework.md
@@ -349,6 +349,8 @@ When `appsRepoPollIntervalSeconds` is set to `0` for a CR, the App Framework wil
 
 ## Add a persistent storage volume to the Operator pod
 
+Note:- If the persistent storage volume is not configured for the Operator, by default, the App Framework uses the main memory(RAM) as the staging area for app package downloads. In order to avoid pressure on the main memory, it is strongly advised to use a persistent volume for the operator pod.
+
 1. Create the persistent volume used by the Operator pod to cache apps and add-ons:
 
 ```yaml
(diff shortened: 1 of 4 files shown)


docs/AppFramework.md
```diff
@@ -468,19 +482,19 @@ metadata:
   name: splunk-manual-app-update
   namespace: default
   ownerReferences:
-  - apiVersion: enterprise.splunk.com/v2
-    controller: false
-    kind: Standalone
-    name: s1
-    uid: ddb9528f-2e25-49be-acd4-4fadde489849
+    - apiVersion: enterprise.splunk.com/v2
```
smohan-splunk, line 485: Should this be v3?

sgontla: addressing it.

pkg/splunk/enterprise/types.go
```diff
@@ -56,6 +56,9 @@ const (
 
 	// SplunkMonitoringConsole is a single instance of Splunk monitor for mc
 	SplunkMonitoringConsole InstanceType = "monitoring-console"
+
+	// TmpAppDownloadDir is the Operator directory for app framework, when there is no explicit volume specified
+	TmpAppDownloadDir string = "/tmp/appframework/"
```
smohan-splunk, line 61: @sgontla Would it make sense to apply this [suggestion](https://github.com/splunk/splunk-operator/pull/671#discussion_r800919312) in this case of Operator too? Something like "/tmp/operator-staging/appframework/"
cc: @gaurav-splunk

sgontla: @smohan-splunk , for the operator, in the case of volume mount, we are already having that kind of directory hierarchy, i.e `/opt/splunk/appframework/`. We may not need the keyword `operator`, as it is already on the operator?

smohan-splunk (approved):

gaurav-splunk (approved):

PR #677, opened 2022-02-11
CSPL-1639: Fix a bug where streamoptions needs to be reset for running same command on pods

gaurav-splunk:
**Problem**
Due to code cleanup for phase-2, we removed all the references of init-container and changed the directory(to temporarily store the apps to be installed) from "/init-apps/" to "/operator/appframework/<appSrcName>". In the latest code, operator would create the latter directory on the splunk pod/s. This worked fine for CRs with 1 pod. But for CRs like Standalone which can scale up to > 1 pods, it was an issue where we were not creating the directory structure required for staging the apps on splunk pod/s.
This eventually resulted in failure during PodCopy phase of app framework where operator was trying to copy the apps from operator pod to splunk pod/s.

**Solution**
Reset the `Stdin` field of the `streamoptions` data structure passed to the podExecClient, so that for each pod we read from the correct offset of the reader pipe.

Here is the link to the detailed description about the problem - https://splunk.atlassian.net/browse/CSPL-1639?focusedCommentId=8556652

### DIFF
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -32,7 +32,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/tools/remotecommand"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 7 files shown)


sgontla (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -1443,22 +1441,33 @@ func checkIfFileExistsOnPod(cr splcommon.MetaObject, filePath string, podExecCli
 	return fileTestResult == 0
 }
 
+// resetStringReader resets the Stdin (of strings.Reader type) of the remotecommand.StreamOptions
```
smohan-splunk, line 1444: Is it more suitable to have `resetStringReader` part of  pkg/splunk/util/util.go?

smohan-splunk (approved):

PR #678, opened 2022-02-14
feat(helm): Support deployment of operator via helm

ryanfaircloth:
This PR adds support to deploy the splunk-operator using helm. The default mode will be cluster-wide (typical use) however using values.yaml single namespace can be selected. 

* Objects created are dynamically named in alignment with the helm release name
* Cluster-wide scope (default) does not require per namespace role binding

TODO: Publish the chart with CI
TODO: Update docs for this deployment option
TODO: the crds folder should be updated when crds are updated


smohan-splunk:
Helm support added through other merges, closing this PR

PR #680, opened 2022-02-16
CSPL-1642: fix a race condition due to which int tests were failing

gaurav-splunk:
**Problem**
One of the integration test was failing because the operator was crashing with the following crash - 
```
{"level":"info","ts":1644979534.06119,"logger":"splunk.enterprise.initAndCheckAppInfoStatus","msg":"App installation is already in progress. Not checking for any latest app repo changes","name":"c3appfw-qs9-l91","namespace":"c3appfw-qs9"}
E0216 02:45:34.064912       1 runtime.go:78] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
goroutine 1856 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x16b0880, 0x1b551a0})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/runtime/runtime.go:74 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0006941a0})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/runtime/runtime.go:48 +0x75
panic({0x16b0880, 0x1b551a0})
        /opt/hostedtoolcache/go/1.17.3/x64/src/runtime/panic.go:1038 +0x215
github.com/splunk/splunk-operator/pkg/splunk/enterprise.createOrUpdateAppUpdateConfigMap({0x7ff8d1d90380, 0xc00080bbc0}, {0x1be3d20, 0xc000b8c800})
        splunk-operator/pkg/splunk/enterprise/configuration.go:937 +0x499
github.com/splunk/splunk-operator/pkg/splunk/enterprise.initAppFrameWorkContext({0x7ff8d1d90380, 0xc00080bbc0}, {0x1be3d20, 0xc000b8c800}, 0xc000b8cdd0, 0xc000b8cee0)
        splunk-operator/pkg/splunk/enterprise/configuration.go:959 +0xd8
github.com/splunk/splunk-operator/pkg/splunk/enterprise.initAndCheckAppInfoStatus({0x7ff8d1d90380, 0xc00080bbc0}, {0x1be3d20, 0xc000b8c800}, 0xc000b8cdd0, 0xc000b8cee0)
        splunk-operator/pkg/splunk/enterprise/util.go:1200 +0x21b
github.com/splunk/splunk-operator/pkg/splunk/enterprise.ApplySearchHeadCluster({0x7ff8d1d90380, 0xc00080bbc0}, 0xc000b8c800)
        splunk-operator/pkg/splunk/enterprise/searchheadcluster.go:64 +0x28f
github.com/splunk/splunk-operator/pkg/controller.SearchHeadClusterController.Reconcile(...)
        splunk-operator/pkg/controller/add_searchheadcluster.go:59
github.com/splunk/splunk-operator/pkg/splunk/controller.splunkReconciler.Reconcile({{0x1bc5090, 0xc00080bbc0}, {0x1b90500, 0x29e91e8}}, {{{0xc000682ea0, 0x25e15086af2}, {0xc0006743c0, 0x485dd7}}})
        splunk-operator/pkg/splunk/controller/controller.go:121 +0x5a7
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0001e35c0, {0x1713d60, 0xc0006941a0})
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:256 +0x19c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0001e35c0)
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:232 +0xc5
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker(...)
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0x7ff8d1e60c78)
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/wait/wait.go:155 +0x67
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0x236d8, {0x1b6b500, 0xc0006a67b0}, 0x1, 0xc000c40480)
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/wait/wait.go:156 +0xb6
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000c011a0, 0x3b9aca00, 0x0, 0x0, 0x43dd85)
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/wait/wait.go:133 +0x89
k8s.io/apimachinery/pkg/util/wait.Until(0x0, 0xc000af6768, 0xc000af6778)
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.18.17/pkg/util/wait/wait.go:90 +0x25
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:193 +0x71e
panic: assignment to entry in nil map [recovered]
        panic: assignment to entry in nil map
```
Based on the above crash log, I could see that the there was a race condition where more than 1 CR reconciles were running in parallel, and they were trying to access and fill this map, **crKindMap**,  in the function **createOrUpdateAppUpdateConfigMap** - 
```
func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject) (*corev1.ConfigMap, error) {
	scopedLog := log.WithName("createOrUpdateAppUpdateConfigMap").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())

	crKindMap := make(map[string]string)
	var configMapData, status string
	var configMap *corev1.ConfigMap
	var err error
	var numOfObjects int

	kind := cr.GetObjectKind().GroupVersionKind().Kind

	configMapName := GetSplunkManualAppUpdateConfigMapName(cr.GetNamespace())
	namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: configMapName}

	configMap, err = splctrl.GetConfigMap(client, namespacedName)
	if err == nil {
		// If this CR is already an owner reference, then do nothing.
		// This can happen if we have already set this CR as ownerRef in the first time,
		// and we reach here again during the next reconcile.
		currentOwnerRef := configMap.GetOwnerReferences()
		for i := 0; i < len(currentOwnerRef); i++ {
			if reflect.DeepEqual(currentOwnerRef[i], splcommon.AsOwner(cr, false)) {
				return configMap, nil
			}
		}

		crKindMap = configMap.Data

		// get the number of instance types of this kind
		numOfObjects = getNumOfOwnerRefsKind(configMap, kind)
	}

	// prepare the configMap data OR
	// initialize the configMap data for this CR type,
	// if it did not exist before
	if _, ok := crKindMap[kind]; !ok {
		status = "off"
	} else {
		status = getManualUpdateStatus(client, cr, configMapName)
	}

	configMapData = fmt.Sprintf(`status: %s
refCount: %d`, status, numOfObjects+1)
	crKindMap[kind] = configMapData <<< This is the line where we crash <<<<<

	// Create/update the configMap to store the values of manual trigger per CR kind.
	configMap, err = ApplyManualAppUpdateConfigMap(client, cr, crKindMap)
	if err != nil {
		scopedLog.Error(err, "Create/update configMap for app update failed")
		return configMap, err
	}

	return configMap, nil
}
```
The line where we crash is this code - `crKindMap[kind] = configMapData`. Somehow, because of race condition, we were getting this crKindMap as nil and would crash while assigning a value in it.

**Solution**
Serialize the access to the configMap

### DIFF
--- a/.github/workflows/int-test-workflow.yml
+++ b/.github/workflows/int-test-workflow.yml
@@ -5,6 +5,7 @@ on:
       - develop
       - master
       - feature-appframework
+      - bugfix/race-condition-fix
 jobs:
   int-tests:
     strategy:
--- a/.github/workflows/int-test-workflow.yml
+++ b/.github/workflows/int-test-workflow.yml
@@ -5,7 +5,6 @@ on:
       - develop
       - master
       - feature-appframework
-      - bugfix/race-condition-fix
 jobs:
   int-tests:
     strategy:
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -827,8 +827,7 @@ func ApplyManualAppUpdateConfigMap(client splcommon.ControllerClient, cr splcomm
 	var err error
 	var newConfigMap bool
 
-	scopedLog.Info("Creating/Updating manual app update configMap")
-
+	scopedLog.Info("new configMap data", "crKindMap", crKindMap)
 	configMap, err = splctrl.GetConfigMap(client, namespacedName)
 	if err != nil {
 		configMap = splctrl.PrepareConfigMap(configMapName, cr.GetNamespace(), crKindMap)
@@ -841,15 +840,17 @@ func ApplyManualAppUpdateConfigMap(client splcommon.ControllerClient, cr splcomm
 	configMap.SetOwnerReferences(append(configMap.GetOwnerReferences(), splcommon.AsOwner(cr, false)))
 
 	if newConfigMap {
+		scopedLog.Info("Creating manual app update configMap")
 		err = splutil.CreateResourceLocked(client, configMap, mux)
 		if err != nil {
 			scopedLog.Error(err, "Unable to create the configMap", "name", configMapName)
 			return configMap, err
 		}
 	} else {
+		scopedLog.Info("Updating manual app update configMap")
 		err = splutil.UpdateResourceLocked(client, configMap, mux)
 		if err != nil {
-			scopedLog.Error(err, "Unable to create the configMap", "name", configMapName)
+			scopedLog.Error(err, "Unable to update the configMap", "name", configMapName)
 			return configMap, err
 		}
 	}
@@ -918,6 +919,7 @@ func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splc
 			}
 		}
 
+		scopedLog.Info("Existing confiMap data", "data", configMap.Data)
 		crKindMap = configMap.Data
 
 		// get the number of instance types of this kind
--- a/pkg/splunk/enterprise/configuration.go
+++ b/pkg/splunk/enterprise/configuration.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"sync"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -815,7 +816,7 @@ func AreRemoteVolumeKeysChanged(client splcommon.ControllerClient, cr splcommon.
 }
 
 // ApplyManualAppUpdateConfigMap applies the manual app update config map
-func ApplyManualAppUpdateConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject, crKindMap map[string]string) (*corev1.ConfigMap, error) {
+func ApplyManualAppUpdateConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject, crKindMap map[string]string, mux *sync.Mutex) (*corev1.ConfigMap, error) {
 
 	scopedLog := log.WithName("ApplyManualAppUpdateConfigMap").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
 
@@ -840,13 +841,13 @@ func ApplyManualAppUpdateConfigMap(client splcommon.ControllerClient, cr splcomm
 	configMap.SetOwnerReferences(append(configMap.GetOwnerReferences(), splcommon.AsOwner(cr, false)))
 
 	if newConfigMap {
-		err = splutil.CreateResource(client, configMap)
+		err = splutil.CreateResourceLocked(client, configMap, mux)
 		if err != nil {
 			scopedLog.Error(err, "Unable to create the configMap", "name", configMapName)
 			return configMap, err
 		}
 	} else {
-		err = splutil.UpdateResource(client, configMap)
+		err = splutil.UpdateResourceLocked(client, configMap, mux)
 		if err != nil {
 			scopedLog.Error(err, "Unable to create the configMap", "name", configMapName)
 			return configMap, err
@@ -891,10 +892,10 @@ func getManualUpdateRefCount(client splcommon.ControllerClient, cr splcommon.Met
 }
 
 // createOrUpdateAppUpdateConfigMap creates or updates the manual app update configMap
-func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject) (*corev1.ConfigMap, error) {
+func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splcommon.MetaObject, mux *sync.Mutex) (*corev1.ConfigMap, error) {
 	scopedLog := log.WithName("createOrUpdateAppUpdateConfigMap").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
 
-	crKindMap := make(map[string]string)
+	var crKindMap map[string]string
 	var configMapData, status string
 	var configMap *corev1.ConfigMap
 	var err error
@@ -926,6 +927,9 @@ func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splc
 	// prepare the configMap data OR
 	// initialize the configMap data for this CR type,
 	// if it did not exist before
+	if crKindMap == nil {
+		crKindMap = make(map[string]string)
+	}
 	if _, ok := crKindMap[kind]; !ok {
 		status = "off"
 	} else {
@@ -937,7 +941,7 @@ refCount: %d`, status, numOfObjects+1)
 	crKindMap[kind] = configMapData
 
 	// Create/update the configMap to store the values of manual trigger per CR kind.
-	configMap, err = ApplyManualAppUpdateConfigMap(client, cr, crKindMap)
+	configMap, err = ApplyManualAppUpdateConfigMap(client, cr, crKindMap, mux)
 	if err != nil {
 		scopedLog.Error(err, "Create/update configMap for app update failed")
 		return configMap, err
@@ -947,7 +951,7 @@ refCount: %d`, status, numOfObjects+1)
 }
 
 // initAppFrameWorkContext used to initialize the appframework context
-func initAppFrameWorkContext(client splcommon.ControllerClient, cr splcommon.MetaObject, appFrameworkConf *enterpriseApi.AppFrameworkSpec, appStatusContext *enterpriseApi.AppDeploymentContext) error {
+func initAppFrameWorkContext(client splcommon.ControllerClient, cr splcommon.MetaObject, appFrameworkConf *enterpriseApi.AppFrameworkSpec, appStatusContext *enterpriseApi.AppDeploymentContext, mux *sync.Mutex) error {
 	if appStatusContext.AppsSrcDeployStatus == nil {
 		appStatusContext.AppsSrcDeployStatus = make(map[string]enterpriseApi.AppSrcDeployInfo)
 		//Note:- Set version only at the time of allocating AppsSrcDeployStatus. This is important, so that we don't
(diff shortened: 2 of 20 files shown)


sgontla:
If it is a race condition, I think we need to have some kind of implementation for serialized access. Otherwise, though the fix may address the issue, for now, it may return in a different form(?)

gaurav-splunk:
> If it is a race condition, I think we need to have some kind of implementation for serialized access. Otherwise, though the fix may address the issue, for now, it may return in a different form(?)

You are right, initially, I put it in to test whether this resolves the failing test case, but now I am working on, probably using a mutex or some lock to serialize the access. My apologies, I should have mentioned it before pulling in reviewers.

.github/workflows/int-test-workflow.yml
```diff
@@ -5,6 +5,7 @@ on:
       - develop
       - master
       - feature-appframework
+      - bugfix/race-condition-fix
```
gaurav-splunk, line 8: I will revert this before merge

gaurav-splunk:
> > If it is a race condition, I think we need to have some kind of implementation for serialized access. Otherwise, though the fix may address the issue, for now, it may return in a different form(?)
> 
> You are right, initially, I put it in to test whether this resolves the failing test case, but now I am working on, probably using a mutex or some lock to serialize the access. My apologies, I should have mentioned it before pulling in reviewers.

This is good for review now!

pkg/splunk/enterprise/configuration.go
```diff
@@ -926,6 +930,9 @@ func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splc
 	// prepare the configMap data OR
 	// initialize the configMap data for this CR type,
 	// if it did not exist before
+	if crKindMap == nil {
```
smohan-splunk, line 933: With the serialized access, is this code still needed? Or can we revert to creation at top of the method?

pkg/splunk/enterprise/configuration.go
```diff
@@ -894,12 +895,15 @@ func getManualUpdateRefCount(client splcommon.ControllerClient, cr splcommon.Met
...
 
-	crKindMap := make(map[string]string)
+	var crKindMap map[string]string
 	var configMapData, status string
 	var configMap *corev1.ConfigMap
 	var err error
 	var numOfObjects int
+	var mux sync.Mutex
 
+	mux.Lock()
```
smohan-splunk, line 905: Can we move the following code to a new method which returns `configMap` and call them here & other places? that way we have a shorter block of code to protect, and should be faster too
```
mux.Lock()
defer mux.Unlock()
configMapName := GetSplunkManualAppUpdateConfigMapName(cr.GetNamespace())
namespacedName := types.NamespacedName{Namespace: cr.GetNamespace(), Name: configMapName}
configMap, err = splctrl.GetConfigMap(client, namespacedName)
```

pkg/splunk/enterprise/configuration.go
gaurav-splunk: We could revert it, but IMO think this is fine too!

pkg/splunk/enterprise/configuration.go
gaurav-splunk: This lock not only protects configMap but the crKindMap kind of things too

pkg/splunk/enterprise/configuration.go
smohan-splunk: Hmm..with this code, under what circumstances do we expect `crKindMap` be nil and need a new map created, especially considering line 924? One scenario I can think of is when `splctrl.GetConfigMap(client, namespacedName)` fails, but do we even want to proceed in that case?

sgontla:
Since the mutex is locally scoped to the function frame, it can't make the access autonomous.  What we need is a global lock, that serves re-entrancy(Ex. parallel runs from multiple controllers, multiple threads, etc...).
I think, we better create a map of locks to guard the resources that need serialized access(ex. Global secret object, manual trigger etc) , which will be initialized when the Operator started the first time. Whenever such a global scoped resource needs updates, then, it should happen with the lock on.

Right now, we have a global tracker for the storage, and i feel, we can re-use the same to introduce this new Map of locks
```
type globalResourceTracker struct {
	storage *storageTracker
}
```

gaurav-splunk:
Right now, just added some logs to debug the crash and see if our theory is valid. And also will be running only c3 test cases to get quicker test result.

pkg/splunk/enterprise/configuration.go
```diff
@@ -917,6 +919,7 @@ func createOrUpdateAppUpdateConfigMap(client splcommon.ControllerClient, cr splc
 			}
 		}
 
+		scopedLog.Info("Existing confiMap data", "data", configMap.Data)
```
smohan-splunk, line 922: nit:configMap

test/c3/appframework/appframework_test.go
```diff
@@ -2643,8 +2643,8 @@ var _ = Describe("c3appfw test", func() {
 		})
 	})
 
-	// CSPL-1647
-	XContext("Clustered deployment (C3 - clustered indexer, search head cluster)", func() {
+	// CSPL-1604
+	Context("Clustered deployment (C3 - clustered indexer, search head cluster)", func() {
 		It("integration, c3, appframeworkc3, appframework: can deploy a C3 SVA and have apps installed and updated locally on Cluster Manager and Deployer for manual polling", func() {
```
smohan-splunk, line 2648: There are more tests that need to be enabled back, both in this file as well as `test/m4/appframework/appframework_test.go`

gaurav-splunk: Yup just enabled one in m4. Let me know if there are other tests that need to be enabled also.

smohan-splunk:
```
grep "CSPL-1647" -rn *                      
test/m4/appframework/appframework_test.go:1047:	// CSPL-1647
test/m4/appframework/appframework_test.go:1231:	// CSPL-1647
test/c3/appframework/appframework_test.go:1642:	// CSPL-1647
test/c3/appframework/appframework_test.go:1826:	// CSPL-1647
```

pkg/splunk/enterprise/configuration.go
sgontla: This is a good question. I think it depends on how the json marshal takes care. At least, based on the original observations from this issue, seems like it never initializes the map(?), when the DB data is nil.

sgontla (approved):

pkg/splunk/enterprise/configuration.go
smohan-splunk: Well, original code had `crKindMap := make(map[string]string)` in the beginning of the function - if whatever reason the `crKindMap == nil ` succeeds post this, the map can still be nil post this new code to make the map. I think we need to understand better why `crKindMap` became nil in the original code

gaurav-splunk: Ok, I think I discussed this before also but let me refresh again. The original problem happened when a previous CR(of same kind in same namespace) was still getting terminated while a new one is coming up. There is a small window, where while deleting the previous CR(and if it is the last CR to be deleted), we remove its entry from the configMap BUT the configMap itself is not deleted yet by the garbage collector of K8s controller. In this case, the data section of the configMap is nil. And now if a new CR is coming up, we will go to line 910 and we will not return `err` since we have a configMap but as soon as we reach line 923, since `configMap.Data` = nil, `crKindMap is also nil and hence we would crash. With the current fix, we are covering that scenario.
Also, as far as line 926 is concerned, `numOfObjects` will be 0 anyways so we are good there too.
I hope this clears the exact scenario and how this fix addresses that problem.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -88,6 +88,16 @@ func ApplyMonitoringConsole(client splcommon.ControllerClient, cr *enterpriseApi
 
 	// check if deletion has been requested
 	if cr.ObjectMeta.DeletionTimestamp != nil {
+		// If this is the last of its kind getting deleted,
```
smohan-splunk, line 91: Why was this change needed, especially in the context of the issue being fixed?

pkg/splunk/enterprise/util.go
```diff
@@ -1188,6 +1217,57 @@ func getCleanObjectDigest(rawObjectDigest *string) (*string, error) {
 	return &cleanObjectHash, nil
 }
 
+// updateManualAppUpdateConfigMapLocked updates the manual app update config map
+func updateManualAppUpdateConfigMapLocked(client splcommon.ControllerClient, cr splcommon.MetaObject, appStatusContext *enterpriseApi.AppDeploymentContext, kind string, turnOffManualChecking bool) error {
```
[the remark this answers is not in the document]
smohan-splunk: I see what you are saying. But comparing both, most of the code related to accessing as well as updating the ConfigMap looks identical. May be a future opportunity to refine further

smohan-splunk (approved):

gaurav-splunk:
In the latest integration test run, the only test to fail is a known issue which is being tracked at https://splunk.atlassian.net/browse/CSPL-1663, rest all the test cases have passed.

PR #682, opened 2022-02-18
CSPL-1326: Update the File permissions for the cluster scoped apps

sgontla:
Splunk volumes are set with the file permissions +660 by default. When an app
package(with their original file permissions) is included as part of a bundle push,
that app is deployed to the cluster members.  However, if there is a pod reset happens,
then all those apps are set with +660 values. Due to this permissions change, whenever
there is a next bundle push, it is  unnecessarily including those apps as part of the bundle
one more time.

Changes are made such that, always set the permissions to +660, before doing a bundle push,
that way the pod reset scenario will not include the apps(with just file permission changes)
that were already part of the last bundle push

### DIFF
--- a/pkg/splunk/enterprise/afwscheduler.go
+++ b/pkg/splunk/enterprise/afwscheduler.go
@@ -598,17 +598,13 @@ func extractClusterScopedAppOnPod(worker *PipelineWorker, appSrcScope string, ap
 	cr := worker.cr
 	scopedLog := log.WithName("extractClusterScopedAppOnPod").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace(), "app name", worker.appDeployInfo.AppName)
 
-	var clusterAppsPath string
 	var stdOut, stdErr string
 	var err error
-	kind := worker.cr.GroupVersionKind().Kind
-	if kind == "SearchHeadCluster" {
-		clusterAppsPath = "/opt/splunk/etc/shcluster/apps/"
-	} else if kind == "ClusterMaster" {
-		clusterAppsPath = "/opt/splunk/etc/master-apps/"
-	} else {
-		// Do not return an error
-		scopedLog.Error(nil, "app extraction should not be called", "kind", kind)
+
+	clusterAppsPath := getClusterScopedAppsLocOnPod(worker.cr)
+	if clusterAppsPath == "" {
+		// This should never happen
+		scopedLog.Error(nil, "could not find the cluster scoped apps location on the Pod")
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/splunk/enterprise/afwscheduler_test.go
```diff
@@ -2130,41 +2135,50 @@ func TestIDXCRunPlaybook(t *testing.T) {
 		},
 	}
 
-	var mockPodExecClient *spltest.MockPodExecClient = &spltest.MockPodExecClient{}
+	var mockPodExecClient *spltest.MockPodExecClient = &spltest.MockPodExecClient{Cr: &cr}
 	mockPodExecClient.AddMockPodExecReturnContexts(podExecCommands, mockPodExecReturnContexts...)
 
 	playbookContext = getClusterScopePlaybookContext(c, &cr, afwPipeline, targetPodName, kind, mockPodExecClient)
 
+	// If the poxExec failed for changing the file permissions, should return an error
```
akondur, line 2143: Nit: podExec vs poxExec?

pkg/splunk/enterprise/util.go
```diff
@@ -1604,7 +1604,7 @@ func setInstallStateForClusterScopedApps(appDeployContext *enterpriseApi.AppDepl
 				deployInfoList[i].PhaseInfo.Phase = enterpriseApi.PhaseInstall
 				deployInfoList[i].PhaseInfo.Status = enterpriseApi.AppPkgInstallComplete
 				scopedLog.Info("Cluster scoped app installed", "app name", deployInfoList[i].AppName, "digest", deployInfoList[i].ObjectHash)
-			} else {
+			} else if deployInfoList[i].PhaseInfo.Phase != enterpriseApi.PhaseInstall || deployInfoList[i].PhaseInfo.Status != enterpriseApi.AppPkgInstallComplete {
```
akondur, line 1607: Is there a reason for this change?

akondur (approved):
Minor comments. Looks good Subba. If there is a github issue/jira can we link it here for reference?

https://splunk.atlassian.net/browse/CSPL-1250
https://splunk.atlassian.net/browse/CSPL-1326

pkg/splunk/enterprise/util.go
sgontla: Ya, it's not related to the changes here, but addressed as a minor issue. Without this change, we were unnecessarily logging errors for the apps, that were already installed successfully as part of earlier bundles.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1291,6 +1287,35 @@ func (shcPlaybookContext *SHCPlaybookContext) triggerBundlePush() error {
...
+}
+
+// adjustClusterAppsFilePermissions sets the file permissions to +550
+func adjustClusterAppsFilePermissions(podExecClient splutil.PodExecClientImpl) error {
+	dirPath := getClusterScopedAppsLocOnPod(podExecClient.GetCR())
+	if dirPath == "" {
+		return fmt.Errorf("invalid Cluster apps location")
+	}
+
+	cmd := fmt.Sprintf("chmod +660 -R %s", dirPath)
```
gaurav-splunk, line 1309: nit: please use a const variable instead of hard coded values

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1291,6 +1287,35 @@ func (shcPlaybookContext *SHCPlaybookContext) triggerBundlePush() error {
...
+		return idxcAppsLocationOnClusterManager
+	case "SearchHeadCluster":
+		return shcAppsLocationOnDeployer
+	default:
+		return ""
+	}
+}
+
+// adjustClusterAppsFilePermissions sets the file permissions to +550
+func adjustClusterAppsFilePermissions(podExecClient splutil.PodExecClientImpl) error {
```
gaurav-splunk, line 1303: I think it's better to create a more generic function so that anybody can use it to change file permission for any directory. So maybe something like `adjustFilePermissions(podExecClient splutil.PodExecClientImpl, dirPath string)` . So the callers of this API can set the diPath and this function would just set the file permissions for that path.

pkg/splunk/enterprise/afwscheduler_test.go
```diff
@@ -2130,41 +2135,50 @@ func TestIDXCRunPlaybook(t *testing.T) {
...
+	mockPodExecReturnContexts[0].StdErr = "Failed"
+	err = playbookContext.runPlaybook()
+	if err == nil {
+		t.Errorf("runPlaybook() should should return an error if the command %v execution fails", podExecCommands[0])
```
gaurav-splunk, line 2147: nit: s/should should return/should return

smohan-splunk (approved):

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: TBH it's not about 1 command but what we are trying to achieve and the usability in the future. But I leave it up to you if you want to change it or not :)

gaurav-splunk (approved):

pkg/splunk/enterprise/afwscheduler.go
sgontla: Agree, but here the use case is a bit narrow for this situation(recursive, append, permission value, etc..), without much visibility for future use cases. If we want to have a generic API, then need the options for all those R, +, and <value>. At this time we are not seeing much scope, leaving it this way for now. But, definitely, if we see a need, we can bring in those extra options and make it a generic API.

PR #692, opened 2022-03-05
CSPL-1661: Config options for the Phase try count and yield timer

sgontla:
- config option for phase max. try count.
- config option for scheduler yield timer.
- Also changed the retry count to try count, to be inline with the retry logic.


pkg/apis/enterprise/v3/common_types.go
```diff
@@ -276,6 +276,18 @@ type AppFrameworkSpec struct {
 	//    3. If anything more than the max value is specified then we set it to 1 day.
 	AppsRepoPollInterval int64 `json:"appsRepoPollIntervalSeconds,omitempty"`
 
+	// Period after which the scheduler tries to yield from the current reconcile entry
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=90
+	// +kubebuilder:validation:Minimum:=30
```
gaurav-splunk, line 282: is this auto-generated or done manually?

pkg/apis/enterprise/v3/common_types.go
```diff
@@ -397,7 +409,7 @@ type PhaseInfo struct {
 	// Status of the phase
 	Status AppPhaseStatusType `json:"status,omitempty"`
 	// RetryCount defines the number of retries completed so far
-	RetryCount int32 `json:"retryCount,omitempty"`
+	TryCount uint32 `json:"tryCount,omitempty"`
```
gaurav-splunk, line 412: Why are we using `TryCount` now vs `RetryCount`? In general, we use RetryCount if we are actually retrying `n` number of times. And other nit comment is to update the comment to reflect whether we are using retryCount or tryCount.
I think fundamentally, try means we only tried once whereas retry means we tried many times, isn't it?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -296,11 +295,11 @@ func getPhaseInfoByPhaseType(worker *PipelineWorker, phaseType enterpriseApi.App
 }
 
 // updatePplnWorkerPhaseInfo updates the in-memory PhaseInfo(specifically status and retryCount)
-func updatePplnWorkerPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, retryCount int32, statusType enterpriseApi.AppPhaseStatusType) {
+func updatePplnWorkerPhaseInfo(appDeployInfo *enterpriseApi.AppDeploymentInfo, retryCount uint32, statusType enterpriseApi.AppPhaseStatusType) {
```
gaurav-splunk, line 298: I think its better to keep parity in the sense that either keep retryCount or tryCount.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -568,15 +567,15 @@ func (ctx *localScopePlaybookContext) runPlaybook() error {
 	stdOut, stdErr, err := ctx.podExecClient.RunPodExecCommand(streamOptions, []string{"/bin/sh"})
 	// if the app was already installed previously, then just mark it for install complete
 	if stdErr != "" || err != nil {
-		phaseInfo.RetryCount++
-		scopedLog.Error(err, "local scoped app package install failed", "stdout", stdOut, "stderr", stdErr, "app pkg path", appPkgPathOnPod, "retry count", phaseInfo.RetryCount)
-		return fmt.Errorf("local scoped app package install failed. stdOut: %s, stdErr: %s, app pkg path: %s, retry count: %d", stdOut, stdErr, appPkgPathOnPod, phaseInfo.RetryCount)
+		phaseInfo.TryCount++
+		scopedLog.Error(err, "local scoped app package install failed", "stdout", stdOut, "stderr", stdErr, "app pkg path", appPkgPathOnPod, "retry count", phaseInfo.TryCount)
```
gaurav-splunk, line 571: ditto

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -2790,15 +2818,15 @@ spec:
...
-                                      description: RetryCount defines the number of
-                                        retries completed so far
-                                      format: int32
-                                      type: integer
                                     status:
                                       description: Status of the phase
                                       format: int32
                                       type: integer
+                                    tryCount:
+                                      description: RetryCount defines the number of
```
gaurav-splunk, line 2826: description does not match the field name

pkg/apis/enterprise/v3/common_types.go
```diff
@@ -276,6 +276,18 @@ type AppFrameworkSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=90
+	// +kubebuilder:validation:Minimum:=30
+	SchedulerYieldInterval uint64 `json:"schedulerYieldIntervalSeconds,omitempty"`
+
+	// Max. number of tries used by the scheduler for a pipeline phase
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=3
+	// +kubebuilder:validation:Minimum:=1
+	PhaseMaxTries uint32 `json:"phaseMaxTries,omitempty"`
```
gaurav-splunk, line 289: Are we doing any validation for these two params in lets say `ValidateAppFrameworkSpec` ?

gaurav-splunk (changes requested):

deploy/crds/enterprise.splunk.com_licensemasters_crd.yaml
```diff
@@ -694,6 +694,20 @@ spec:
                       same time
                     format: int64
                     type: integer
+                  phaseMaxTries:
+                    default: 3
+                    description: Max. number of tries used by the scheduler for a
```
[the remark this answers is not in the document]
sgontla: No, we haven't specified the details about the pipeline. Idea is to provide the default value and suggest this only when there is a real need to tune the config.

pkg/apis/enterprise/v3/common_types.go
sgontla: It is added manually.

pkg/apis/enterprise/v3/common_types.go
sgontla: Kubebuilder will validate it for us, so, we are not doing any explicit validation.

pkg/apis/enterprise/v3/common_types.go
sgontla: Right now, the Retry semantic is broken. When it fails for the first time, we are incrementing RetryCount to 1, similarly for the 2nd time, ... n times. When somebody looks at the CR status, it doesn't look correct. If we want to track just the retry count alone, then we need the additional logic to differentiate the first time try and then retry logic, and also need to do that consciously, as we are updating the CR on every reconcile. Hence, unifying it with just TryCount, which means how many times we are trying, and it implicitly covers the retry count.


(will address the nit comment in my next commit)

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
sgontla: I missed addressing the description part, will address it.

pkg/apis/enterprise/v3/common_types.go
sgontla: Hmm, semantically this is not correct right. Re-try is only a second time and onwards.

gaurav-splunk: Yeah, that is what I am saying too :). You tried the very first time and you failed and now you are retrying(for the second time) and the counter reflects how many times you retried. Am I still not making sense? :D

sgontla: My point is, say if we are configuring Max. retry count as 1, and when it fails in a phase, the retry count is marked as 1, same is updated in the CR, which means that when somebody dumps the CR, it kind of sounds like we retried one time and still failed. Actually, we never retried.
In addition to this, existing logic also fails, as the Max. the retry count is reached. 
As i mentioned earlier, for making the try vs retry scenarios clean, we need to differentiate if across the places, and it  increases unnecessary complexity. So, it is simple to go with just the try counter.

deploy/crds/enterprise.splunk.com_licensemasters_crd.yaml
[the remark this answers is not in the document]
smohan-splunk: This can be cryptic for someone looking to tweak. Either expose the config with clear explanation, or call it for internal/support purposes only or just don't expose it at all

smohan-splunk (approved):

gaurav-splunk (approved):

deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -637,6 +637,15 @@ spec:
                 description: Splunk Enterprise App repository. Specifies remote App
                   location and scope for Splunk App management
                 properties:
+                  appInstallPeriodSeconds:
+                    default: 90
+                    description: 'App installation period within a reconcile. Apps
```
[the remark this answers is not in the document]
sgontla: Seems like the SDK is inserting the quotes, whenever the text consists of any character other than the alphanumeric. I see this pattern in other places as well :-). So, we are good.

Issue #655, opened 2022-01-14
App framework S3 client multiple issues


rfaircloth-splunk:
the "minio" client requires tls however there are many use cases where tls would not be desirable such as when minio is deployed to k8s in the same cluster.

the aws s3 client uses a "hack" regex parse to find the region and passes empty string rather than usingthe env var AWS_REGION.

the aws s3 client should allow for a specified region as a new param. The aws s3 client should provide a param to indicate the endpoint should be provided as an argument to the sdk it is currently ignores this is required to support s3 on the edge and third party implementations like ceph

atopian:
This also creates some issues in AKS deployments, see https://github.com/aws/aws-cli/issues/5234

From the above, exposing AWS_EC2_METADATA_DISABLED as a variable will help in EKS/AKS where IMDS is a different implementation

atopian:
Additionally, a debug flag for the aws-cli would be awesome, as this took far longer to chase down without an easy way to increase verbosity in a cluster.

lindeskar:
Just ran into the limitation with `region` being extracted from the endpoint URL. Looks like the regex has "amazonaws.com" hardcoded, which makes using any other S3-compatible service fail?

Any way around this? Setting `AWS_REGION` using `extraEnv` does not apply on the init containers.

gaurav-splunk:
The issue with the `region` field is fixed in this PR: https://github.com/splunk/splunk-operator/pull/700. In upcoming releases, `region` is a configurable parameter in the CR spec.

akondur:
Fixed via the [PR](https://github.com/splunk/splunk-operator/pull/700).

PR #700, opened 2022-03-15
CSPL-1670: Add region as a configurable parameter in volume spec

gaurav-splunk:
**Problem**
Because of the issue mentioned [here](https://github.com/splunk/splunk-operator/issues/655), we were not able to pass in the region for a S3 compatible storage other than AWS and hence it would pass empty value as region.

**Solution**
Add `region` field as part of volume spec so that users can directly specify the region irrespective of the remote storage they are using.

### DIFF
--- a/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
+++ b/deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
@@ -707,6 +707,10 @@ spec:
                           description: 'App Package Remote Store provider. Supported
                             values: aws, minio'
                           type: string
+                        region:
+                          description: Region of the remote storage volume where apps
+                            reside
+                          type: string
                         secretRef:
                           description: Secret object name
                           type: string
@@ -1396,6 +1400,10 @@ spec:
                         provider:
                           description: 'App Package Remote Store provider. Supported
                             values: aws, minio'
+                          type: string
+                        region:
+                          description: Region of the remote storage volume where apps
+                            reside
                           type: string
                         secretRef:
                           description: Secret object name
@@ -2753,6 +2761,10 @@ spec:
                               description: 'App Package Remote Store provider. Supported
                                 values: aws, minio'
                               type: string
+                            region:
+                              description: Region of the remote storage volume where
+                                apps reside
+                              type: string
                             secretRef:
                               description: Secret object name
                               type: string
@@ -3004,6 +3016,10 @@ spec:
                         provider:
                           description: 'App Package Remote Store provider. Supported
                             values: aws, minio'
+                          type: string
+                        region:
+                          description: Region of the remote storage volume where apps
+                            reside
                           type: string
                         secretRef:
                           description: Secret object name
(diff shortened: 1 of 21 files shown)


deploy/crds/enterprise.splunk.com_clustermasters_crd.yaml
```diff
@@ -707,6 +707,9 @@ spec:
                           description: 'App Package Remote Store provider. Supported
                             values: aws, minio'
                           type: string
+                        region:
+                          description: Region on the remote storage where apps reside
```
sgontla, line 711: "Region of the remote storage volume where apps reside"

pkg/splunk/client/awss3client.go
```diff
@@ -52,23 +51,10 @@ type AWSS3Client struct {
 	AWSSecretAccessKey string
 	Prefix             string
 	StartAfter         string
-	Endpoint           string
 	Client             SplunkAWSS3Client
 	Downloader         SplunkAWSDownloadClient
 }
 
-// regex to extract the region from the s3 endpoint
```
smohan-splunk, line 60: Rather than completely removing the extraction, should we still maintain backward compatibility by attempting extraction if region in the CR spec hasn't been specified?

Otherwise, customers upgrading to this version without region explicitly set in their CRs would start seeing failures?

akondur (approved):

gaurav-splunk:
The only failure right now is in s1 tests, which is a known env issue.

pkg/splunk/client/awss3client.go
```diff
@@ -125,10 +123,15 @@ func InitAWSClientSession(region, accessKeyID, secretAccessKey string) SplunkAWS
...
-func NewAWSS3Client(bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string, endpoint string, fn GetInitFunc) (S3Client, error) {
+func NewAWSS3Client(bucketName string, accessKeyID string, secretAccessKey string, prefix string, startAfter string, region string, endpoint string, fn GetInitFunc) (S3Client, error) {
 	var s3SplunkClient SplunkAWSS3Client
 	var err error
-	region := GetRegion(endpoint)
+
+	// for backward compatibility, if `region` is not specified in the CR,
+	// then derive the region from the endpoint.
+	if region == "" {
+		region = GetRegion(endpoint)
```
smohan-splunk, line 133: What happens if region is not specified AND region can't be extracted from the endpoint .. should we return error? In other words, is region mandatory?

pkg/splunk/client/awss3client_test.go
```diff
@@ -35,7 +35,7 @@ func TestInitAWSClientWrapper(t *testing.T) {
 func TestNewAWSS3Client(t *testing.T) {
 
 	fn := InitAWSClientWrapper
-	awsS3Client, err := NewAWSS3Client("sample_bucket", "abcd", "xyz", "admin/", "admin", "https://s3.us-west-2.amazonaws.com", fn)
+	awsS3Client, err := NewAWSS3Client("sample_bucket", "abcd", "xyz", "admin/", "admin", "us-west-2", "https://s3.us-west-2.amazonaws.com", fn)
```
smohan-splunk, line 38: Can we also add test case for backward compatibility case?

pkg/splunk/client/awss3client.go
gaurav-splunk: Good point, let me make the change to return error in case we are not able to extract the region.

smohan-splunk (approved):

sgontla (approved):

Issue #689, opened 2022-03-01
Organization Compliance Requirement to have spec.securityContext.runAsNonRoot=true


cderocco5:
My organization requires that all statefulsets have spec.securityContext.runAsNonRoot=true. There needs to be a way to configure this setting through the CRD to set runAsNonRoot to true. e.g
spec: 
    template:
        spec:
            securityContext:
                 runAsNonRoot: true

akondur:
Hi @cderocco5 , the issue has been fixed via the linked [PR](https://github.com/splunk/splunk-operator/pull/703). 

Currently we are setting the runAsNonRoot to true by default considering that we do not want to run containers as root due to security concerns.

PR #703, opened 2022-03-17
CSPL-1652: Set runasnonroot to true by default

akondur:
Addressing the [github issue](https://github.com/splunk/splunk-operator/issues/689) 

Set podSecurityContext **runAsNonRoot** to true by default as running containers as root is not encouraged. If a customer needs to run as root, they can do so via a private operator build/binary.

Example standalone with new setting:
```
$ kubectl get sts splunk-test2-standalone -o yaml | grep -i nonroot
              f:runAsNonRoot: {}
        runAsNonRoot: true
```


smohan-splunk:
Why did we choose not to allow the option of configuring the podSecurityContext via CRDs, as requested in the gitHub issue? What would the impact of changing to runAsNonRoot by default be on existing customers when they upgrade?

sgontla (approved):

jryb (approved):

smohan-splunk (approved):

PR #713, opened 2022-03-25
CSPL-1677: Update yield logic to use context

gaurav-splunk:


pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1565,7 +1566,13 @@ func afwSchedulerEntry(client splcommon.ControllerClient, cr splcommon.MetaObjec
...
 	// Wait for the yield function to finish.
 	afwPipeline.phaseWaiter.Add(1)
-	go afwPipeline.afwYieldWatcher()
+
+	ctx, cancel := context.WithTimeout(
+		ctx,
+		time.Duration(afwPipeline.appDeployContext.AppFrameworkConfig.SchedulerYieldInterval)*time.Second,
+	)
+	defer cancel()
+	go afwPipeline.afwYieldWatcher(ctx)
```
sgontla, line 1575: I am wondering if the same child context can be passed to all the phase managers, in which case that can be used as a termination signal.

sgontla:
@vivekr-splunk  is propagating the context, as part of his PR: https://github.com/splunk/splunk-operator/pull/707. So, you can just use the local context as part of the `afwSchedulerEntry()` for now. That way, the changes will be just local to `afwscheduler.go`, and minimizes the work.

gaurav-splunk:
> @vivekr-splunk is propagating the context, as part of his PR: #707. So, you can just use the local context as part of the `afwSchedulerEntry()` for now. That way, the changes will be just local to `afwscheduler.go`, and minimizes the work.

I know Vivek is passing the context as part of his PR and that is the reason I initialized the context in the main reconcile logic so that the context he is passing in as argument can be reused when we bring in those changes.

pkg/splunk/enterprise/afwscheduler.go
gaurav-splunk: Yes, that is part of the next overall change in app framework code to use contexts. This is just an initial PR that introduces the context to this code. I will put the other change as an enhancement which requires some more code changes on the UTs too.

gaurav-splunk: Also, on a side note, the termination happens not only because of timeout but also if pipeline is empty. Hence, I think, using context.Done() only would not suffice.

sgontla:
> > @vivekr-splunk is propagating the context, as part of his PR: #707. So, you can just use the local context as part of the `afwSchedulerEntry()` for now. That way, the changes will be just local to `afwscheduler.go`, and minimizes the work.
> 
> I know Vivek is passing the context as part of his PR and that is the reason I initialized the context in the main reconcile logic so that the context he is passing in as an argument can be reused when we bring in those changes.

Looking at the changes in #707,  it has the required changes to propagate the context all the way to the `afwSchedulerEntry()` and also all the functions defined in afwscheduler.go.   So, ideally this PR can be on top of #707, with min. changes just alone to the `afwSchedulerEntry()` for now.

PR #717, opened 2022-03-30
CSPL-1682: Splunk pods get recycled after upgrading operator from 1.0…

vivekr-splunk:
- changes done secret comparison before updating ownership reference. 
- set requeueafter to 0 when requeue is false.


smohan-splunk (approved):

pkg/splunk/controller/configmap.go
```diff
@@ -130,7 +130,7 @@ func SetConfigMapOwnerRef(ctx context.Context, client splcommon.ControllerClient
 	currentOwnerRef := configMap.GetOwnerReferences()
 	// Check if owner ref exists
 	for i := 0; i < len(currentOwnerRef); i++ {
-		if reflect.DeepEqual(currentOwnerRef[i], splcommon.AsOwner(cr, false)) {
+		if reflect.DeepEqual(currentOwnerRef[i].UID, cr.GetUID()) {
```
sgontla, line 133: You mentioned that we can include the Kind as well. In that line of thought, just wondering, is it a good idea, just to exclude the API version alone?  A new API that compares all the fields except API version.
```
`// AsOwner returns an object to use for Kubernetes resource ownership references.
func AsOwner(cr MetaObject, isController bool) metav1.OwnerReference {
	return metav1.OwnerReference{
		APIVersion: cr.GetObjectKind().GroupVersionKind().GroupVersion().String(),
		Kind:       cr.GetObjectKind().GroupVersionKind().Kind,
		Name:       cr.GetObjectMeta().GetName(),
		UID:        cr.GetObjectMeta().GetUID(),
		Controller: &isController,
	}
}`
```

pkg/splunk/controller/statefulset.go
```diff
@@ -326,7 +326,7 @@ func DeleteReferencesToAutomatedMCIfExists(ctx context.Context, client splcommon
 
 //isCurrentCROwner returns true if current CR is the ONLY owner of the automated MC
 func isCurrentCROwner(cr splcommon.MetaObject, currentOwners []metav1.OwnerReference) bool {
-	return reflect.DeepEqual(currentOwners[0], splcommon.AsOwner(cr, false))
+	return reflect.DeepEqual(currentOwners[0].UID, cr.GetUID())
```
akondur, line 329: Same comment as above. Maybe we can create function to compare the four fields - kind, name, UID and controller.

pkg/splunk/util/secrets.go
```diff
@@ -121,7 +121,7 @@ func SetSecretOwnerRef(ctx context.Context, client splcommon.ControllerClient, s
 	currentOwnerRef := secret.GetOwnerReferences()
 	// Check if owner ref exists
 	for i := 0; i < len(currentOwnerRef); i++ {
-		if reflect.DeepEqual(currentOwnerRef[i], splcommon.AsOwner(cr, false)) {
+		if reflect.DeepEqual(currentOwnerRef[i].UID, cr.GetUID()) {
```
akondur, line 124: Same comment as above.

pkg/splunk/controller/configmap.go
vivekr-splunk: we do not need controller, we just need  UID. UID will uniquely identify the resource. please [refer](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=A%20Kubernetes%20systems%2Dgenerated%20string,(also%20known%20as%20UUIDs).).

pkg/splunk/controller/statefulset.go
vivekr-splunk: we do not need controller, we just need  UID. UID will uniquely identify the resource. please [refer](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=A%20Kubernetes%20systems%2Dgenerated%20string,(also%20known%20as%20UUIDs).).

pkg/splunk/controller/configmap.go
akondur: Does the UID remain the if we update the name, kind of the CR(I'm not sure if these are even possible)? If yes then comparing the UID's should be enough.

sgontla (approved):

akondur (approved):

PR #727, opened 2022-04-12
rebase changes with PR:707

tpereirasplunk:
Fix Issues with pipeline

Int test run: https://github.com/splunk/splunk-operator/runs/6011142014?check_suite_focus=true

### DIFF
--- a/test/testenv/deployment.go
+++ b/test/testenv/deployment.go
@@ -759,8 +759,6 @@ func (d *Deployment) DeployClusterMasterWithGivenSpec(ctx context.Context, name 
 	if err != nil {
 		return nil, err
 	}
-	// Verify standalone goes to ready state
-	ClusterManagerReady(ctx, d, d.testenv)
 	return deployed.(*enterpriseApi.ClusterMaster), err
 }
 
--- a/test/c3/appframework/appframework_test.go
+++ b/test/c3/appframework/appframework_test.go
@@ -210,7 +210,7 @@ var _ = Describe("c3appfw test", func() {
 			clusterManagerBundleHash := testenv.AppFrameWorkVerifications(ctx, deployment, testcaseEnvInst, allAppSourceInfo, splunkPodAge, "")
 
 			// Verify no pods reset by checking the pod age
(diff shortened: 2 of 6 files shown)


jambrosiano (approved):

pdhanoyasplunk (approved):

test/testenv/deployment.go
```diff
@@ -759,8 +759,6 @@ func (d *Deployment) DeployClusterMasterWithGivenSpec(ctx context.Context, name
 	if err != nil {
 		return nil, err
 	}
-	// Verify standalone goes to ready state
-	ClusterManagerReady(ctx, d, d.testenv)
```
smohan-splunk, line 763: Can you please clarify why this change?

tpereirasplunk: This was added as @vivekr-splunk indicated it is highly unlikely for customers to deploy both CM and SHC at the same time as we do in automation. They would wait for CM to be ready before doing the same.

However if we wait for CM to be ready most times the app are downloaded and about to be installed. We have tests that require us to interrupt the operator in the download phase which would be missed and the test would fail.

sgontla: @tpereirasplunk, the App framework should not come into the timelines and the way the SHC connects CM, as the Phase-3 app framework always waits for the corresponding Pod to come up, before starting the app framework. In phase-2 it may be influencing the sequence. I wonder if @vivekr-splunk  did this change considering the phase-2 app framework which was still the case with the 1.1.0 release.

tpereirasplunk: @sgontla . The issue with this change was that we waited for CM to be ready before deploying SHC. So by the time we have CM ready and SHC deployed, the phase on CM is post download phase.

And we needed to interrupt the operator in download phase for one of the test case. So we usually call the function to check if CM is ready after in a separate step

tpereirasplunk: This call is already happening later after all specs are deployed and would be a duplicate if made here

sgontla (approved):

smohan-splunk (approved):

Issue #83, opened 2020-05-27
Add imagePullSecrets parameter to common spec


mikedickey:
All of Splunk Operator's custom resources that provision pods should support an imagePullSecrets parameter, which could be used to provide secrets for pulling the pod images from private registries.

mikedickey:
Related to #74 in that it might be a good idea to tackle these together.

akondur:
Closing the issue as `imagePullSecrets` config is supported now.

PR #730, opened 2022-04-12
CSPL-1707 - Allow configuration of ImagePullSecrets

akondur:
Allow configuration of `imagePullSecrets`. Addressing [github issue](https://github.com/splunk/splunk-operator/issues/83)

**Testing**
Tested with this [image](https://hub.docker.com/layers/202397823/akondur/priv/splunkedge/images/sha256-b71afed56f234ee0e50febd23d0c991a3c34168b0902d9701fd8494cc097e673?context=repo) from a private repo.

Note: Without **imagePullSecrets** config the container image pull fails as no authorization is provided

**Standalone config with imagePullSecrets**
```
$ cat ~/example_standalone.yml
kind: Standalone
metadata:
  name: test3
  finalizers:
spec:
  image: akondur/priv:splunkedge
  imagePullSecrets:
  - name: regcred
```

**Secret for .dockerconfigjson configuration**
```
$ kubectl get secrets regcred -o wide
NAME      TYPE                             DATA   AGE
regcred   kubernetes.io/dockerconfigjson   1      6h15m
```

**StatefulSet picked up the change**
```
kubectl get sts splunk-test3-standalone -o yaml | grep -i imagePullSecrets -A 3
            f:imagePullSecrets:
              .: {}
              k:{"name":"regcred"}:
                .: {}
--
      imagePullSecrets:
      - name: regcred
      restartPolicy: Always
      schedulerName: default-scheduler
```
**Running Pods(successfully pulled image from a private repository)**
```
$ kubectl get pods
NAME                                                 READY   STATUS    RESTARTS   AGE
splunk-operator-controller-manager-fff796db8-vp2st   2/2     Running   0          3d22h
splunk-test3-standalone-0                            1/1     Running   0          6h13m
 
$ kubectl describe pods splunk-test3-standalone-0 | grep -i image
    Image:          akondur/priv:splunkedge
    Image ID:       docker-pullable://667741767953.dkr.ecr.us-west-2.amazonaws.com/akondur/splunk/splunk@sha256:b71afed56f234ee0e50febd23d0c991a3c34168b0902d9701fd8494cc097e673
```

- Added a change to sort the imagePullSecrets slice on creation of StatefulSet & identify changes in it on every reconcile.


config/crd/bases/enterprise.splunk.com_clustermasters.yaml
```diff
@@ -1151,6 +1151,19 @@ spec:
...
+              imagePullSecrets:
+                description: Sets imagePullSecrets if image is being pulled from a
+                  private registry See https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+                items:
+                  description: LocalObjectReference contains enough information to
+                    let you locate the referenced object inside the same namespace.
+                  properties:
+                    name:
+                      description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                        TODO: Add other useful fields. apiVersion, kind, uid?'
```
jryb, line 1163: Where's this TODO come from?  Is that from the kubernetes code?

pkg/splunk/enterprise/configuration.go
```diff
@@ -216,11 +214,41 @@ func validateCommonSplunkSpec(spec *enterpriseApi.CommonSplunkSpec) error {
...
+	scopedLog := reqLogger.WithName("ValidateImagePullSecrets").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+
+	// If no imagePullSecrets are configured
+	var nilImagePullSecrets []corev1.LocalObjectReference
+	if len(spec.ImagePullSecrets) == 0 {
+		spec.ImagePullSecrets = nilImagePullSecrets
+		return nil
+	}
+
+	// If configured, validated if the secret/s exist
```
jryb, line 240: So there can be multiple `imagePullSecrets`? When an image pull is needed does it just walk the list until one secret works?

pkg/splunk/util/secrets.go
```diff
@@ -495,14 +495,13 @@ func ApplyNamespaceScopedSecretObject(ctx context.Context, client splcommon.Cont
 }
 
 // GetSecretByName retrieves namespace scoped secret object for a given name
-func GetSecretByName(ctx context.Context, c splcommon.ControllerClient, cr splcommon.MetaObject, name string) (*corev1.Secret, error) {
+func GetSecretByName(ctx context.Context, c splcommon.ControllerClient, namespace string, crName string, name string) (*corev1.Secret, error) {
```
jryb, line 498: `crName` is just used for logging?

jryb (approved):

config/crd/bases/enterprise.splunk.com_clustermasters.yaml
akondur: Yes it was automatically added while generating CRDs(possible from the source code of kubernetes)

pkg/splunk/enterprise/configuration.go
akondur: Yes there can be multiple secrets in the  `imagePullSecret` config and it walks the list & stops when it finds a working one.

pkg/splunk/util/secrets.go
akondur: Yes, initially the API `GetSecretByName` had a parameter `cr splcommon.MetaObject` passed in only for logging. I have introduced the parameter `crName` to retain the logging functionality(perhaps I can change the name of the parameter to loggingName to make it more clear).

I have changed the implementation of the API to not use `splcommon.MetaObject` as the purpose of the API is to retrieve a secret by its `name`, `namespace`(has nothing to do with `splcommon.MetaObject`. This helps in making the API being accessible to areas of code where there are no objects of type `splcommon.MetaObject` involved.

jryb: Sounds good to me.

config/crd/bases/enterprise.splunk.com_clustermasters.yaml
```diff
@@ -1151,6 +1151,19 @@ spec:
                 - Always
                 - IfNotPresent
                 type: string
+              imagePullSecrets:
+                description: Sets imagePullSecrets if image is being pulled from a
+                  private registry See https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
```
smohan-splunk, line 1156: super nit: a '.' between registry and See

pkg/splunk/enterprise/configuration.go
```diff
@@ -308,11 +306,41 @@ func validateCommonSplunkSpec(spec *enterpriseApi.CommonSplunkSpec) error {
...
+	for _, secret := range spec.ImagePullSecrets {
+		_, err := splutil.GetSecretByName(ctx, c, cr.GetNamespace(), cr.GetName(), secret.Name)
+		if err != nil {
+			scopedLog.Error(err, "Secret in the imagePullSecrets config is not found ", secret.Name)
```
smohan-splunk, line 336: Just thinking aloud.. if the secret in imagePullSecrets is not found, should we just war, use default values and continue?

pkg/splunk/util/secrets.go
```diff
@@ -495,14 +495,13 @@ func ApplyNamespaceScopedSecretObject(ctx context.Context, client splcommon.Cont
 }
 
 // GetSecretByName retrieves namespace scoped secret object for a given name
-func GetSecretByName(ctx context.Context, c splcommon.ControllerClient, cr splcommon.MetaObject, name string) (*corev1.Secret, error) {
+func GetSecretByName(ctx context.Context, c splcommon.ControllerClient, namespace string, logHandle string, name string) (*corev1.Secret, error) {
```
smohan-splunk, line 498: Why did we change the arguments? Earlier approach of passing the cr & extracting the Name & NS in one place within GetSecretByName seemed a cleaner approach (rather than in all callers)

pkg/splunk/enterprise/configuration.go
```diff
@@ -308,11 +306,41 @@ func validateCommonSplunkSpec(spec *enterpriseApi.CommonSplunkSpec) error {
...
+		return err
+	}
+
 	setVolumeDefaults(spec)
 
 	return ValidateSpec(&spec.Spec, defaultResources)
 }
 
+// ValidateImagePullSecrets sets default values for imagePullSecrets if not provided
+func ValidateImagePullSecrets(ctx context.Context, c splcommon.ControllerClient, cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunkSpec) error {
```
smohan-splunk, line 321: Could we extract the spec from the cr here rather than passing it all the way from validateCommonSplunkSpec?

pkg/splunk/enterprise/configuration.go
```diff
@@ -516,9 +544,10 @@ func getSplunkStatefulSet(ctx context.Context, client splcommon.ControllerClient
 				Annotations: annotations,
 			},
 			Spec: corev1.PodSpec{
-				Affinity:      affinity,
-				Tolerations:   spec.Tolerations,
-				SchedulerName: spec.SchedulerName,
+				Affinity:         affinity,
+				Tolerations:      spec.Tolerations,
+				SchedulerName:    spec.SchedulerName,
+				ImagePullSecrets: spec.ImagePullSecrets,
```
sgontla, line 550: Can we not populate, when the list is empty?

pkg/splunk/enterprise/configuration.go
```diff
@@ -308,11 +306,41 @@ func validateCommonSplunkSpec(spec *enterpriseApi.CommonSplunkSpec) error {
 		return fmt.Errorf("Negative value (%d) is not allowed for Readiness probe intial delay", spec.ReadinessInitialDelaySeconds)
 	}
 
+	// if not provided, set default values for imagePullSecrets
+	err := ValidateImagePullSecrets(ctx, c, cr, spec)
+	if err != nil {
+		return err
+	}
+
```
sgontla, line 314: I think we can conditionally call, if the slice is not empty?

sgontla:
Just for my understanding, all the Splunk images being public, just curious how this is going to be used.

pkg/splunk/enterprise/configuration.go
akondur: The CR object here is of type [splcommon.MetaObject](https://github.com/splunk/splunk-operator/blob/93edb2ca16aba6bbab1ef3accd54532416db812f/pkg/splunk/common/types.go#L111) and the enterpriseApi.CommonSplunkSpec cannot be extracted from the same. The reason we have the common metaObject is to make sure that we are able to call into a common function for all CRs regardless of the type of CR.

pkg/splunk/util/secrets.go
akondur: See [here](https://github.com/splunk/splunk-operator/pull/730#discussion_r848975369). The API prototype was changed as the purpose of the API is to retrieve a secret by its name, namespace(has nothing to do with **splcommon.MetaObject**). This helps in making the API being accessible to areas of code where there are no objects of type splcommon.MetaObject involved.

pkg/splunk/enterprise/configuration.go
akondur: If `spec.ImagePullSecrets` is empty, it assigns an empty value which the StatefulSet will do by default?

pkg/splunk/enterprise/configuration.go
akondur: In order to retain the logic [here](https://github.com/splunk/splunk-operator/pull/730/files#diff-57923b610c6cabf2ed47c4c3b44b4b365366b4949242100ff5c9856e02562424R550), we need to assign `spec.ImagePullSecrets` a nil slice? I am not sure if the CR spec automatically assigns the nil slice in which case we don't need to assign(I have seen issues before where we ran into issues for not assigning values to all fields of CR spec? i.e in integration tests).

pkg/splunk/enterprise/configuration.go
akondur: Good point @smohan-splunk , the K8s client runs through the list of secrets to find a valid secret. The only problem would be when we don't find any of the secrets! Let me change the logic to throw an error only if NONE of them are found.

smohan-splunk:
> Just for my understanding, all the Splunk images being public, just curious how this is going to be used.

@sgontla Some details of the usecase [here](https://jira.splunk.com/browse/CSPL-734)

pkg/splunk/enterprise/configuration.go
sgontla: Ideally, the spec should have been tagged as `omitempty` which will be ignored by the API server. And I think this is important IMO. As all our docker images are public, the customer may not even configure the secrets, instead, he can just set up open read permissions for the repository. So, we should make this config optional. thoughts?

akondur: Currently, the config for `imagePullSecrets` from the CR spec is optional. But the statefulSet is always populated i.e even if we have no CR spec a nil slice is updated. Let's say there is no CR spec and you point it to a public repo image, that image from is pulled cleanly. I agree that we should only warn the user upon not finding any of the secrets in the config!

smohan-splunk (approved):

pkg/splunk/enterprise/configuration.go
sgontla: After thinking through, empty slice is required, considering the spec update cases, we are good here.

sgontla (approved):

PR #736, opened 2022-04-13
updated document changes for upgrade scenario

vivekr-splunk:
1. added upgrade script download step

### DIFF
--- a/docs/SplunkOperatorUpgrade.md
+++ b/docs/SplunkOperatorUpgrade.md
@@ -23,13 +23,19 @@ A Splunk Operator for Kubernetes upgrade might include support for a later versi
 ```
 wget -O splunk-operator-install.yaml https://github.com/splunk/splunk-operator/releases/download/1.1.0/splunk-operator-install.yaml
 ```
+
+2. Download the upgrade script.
 ​
-2. (Optional) Review the file and update it with your specific customizations used during your install. 
+```
+wget -O upgrade-to-1.1.0.sh https://github.com/splunk/splunk-operator/releases/download/1.1.0/upgrade-to-1.1.0.sh
+```
+
+3. (Optional) Review the file and update it with your specific customizations used during your install. 
 ​
-3. Upgrade the Splunk Operator.
+4. Upgrade the Splunk Operator.
 # Splunk Operator Upgrade
 
-Upgrading the Splunk operator to Version 1.1.0 is a new installation rather than an upgrade from the current operator. The older Splunk operator must be cleaned up before installing the new version. Script `tools/upgrade-to-1.1.0.sh` helps you to do the cleanup. The script expects the current namespace where the operator is installed and the path to the 1.1.0 manifest file. The script performs the following steps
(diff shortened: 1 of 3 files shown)


docs/SplunkOperatorUpgrade.md
```diff
@@ -23,13 +23,19 @@ A Splunk Operator for Kubernetes upgrade might include support for a later versi
...
 ```
+
+2. Download the upgrade script.
 ​
-2. (Optional) Review the file and update it with your specific customizations used during your install. 
+```
+wget -O upgrade-to-1.1.0.sh https://github.com/splunk/splunk-operator/releases/download/1.1.0/upgrade-to-1.1.0.sh
+```
+
+3. (Optional) Review the file and update it with your specific customizations used during your install.
```
akondur, line 33: What is file referring to here? It would be nice if we can specify that.

docs/SplunkOperatorUpgrade.md
```diff
@@ -23,13 +23,19 @@ A Splunk Operator for Kubernetes upgrade might include support for a later versi
 ```
 wget -O splunk-operator-install.yaml https://github.com/splunk/splunk-operator/releases/download/1.1.0/splunk-operator-install.yaml
 ```
+
+2. Download the upgrade script.
```
akondur, line 27: Exchange steps 2 & 3 just to keep consistency i.e when we say review the file it may be easier for the user to follow that they need to review `splunk-operator-install.yml`

docs/SplunkOperatorUpgrade.md
```diff
@@ -23,13 +23,19 @@ A Splunk Operator for Kubernetes upgrade might include support for a later versi
...
+```
+
+3. (Optional) Review the file and update it with your specific customizations used during your install. 
 ​
-3. Upgrade the Splunk Operator.
+4. Upgrade the Splunk Operator.
 # Splunk Operator Upgrade
 
-Upgrading the Splunk operator to Version 1.1.0 is a new installation rather than an upgrade from the current operator. The older Splunk operator must be cleaned up before installing the new version. Script `tools/upgrade-to-1.1.0.sh` helps you to do the cleanup. The script expects the current namespace where the operator is installed and the path to the 1.1.0 manifest file. The script performs the following steps
+Upgrading the Splunk operator to Version 1.1.0 is a new installation rather than an upgrade from the current operator. The older Splunk operator must be cleaned up before installing the new version. Script `upgrade-to-1.1.0.sh` helps you to do the cleanup. The script expects the current namespace where the operator is installed and the path to the 1.1.0 manifest file. The script performs the following steps
```
akondur, line 38: Maybe we can hyperlink the script here - `[upgrade-to-1.1.0.sh](https://github.com/splunk/splunk-operator/releases/download/1.1.0/upgrade-to-1.1.0.sh)`.

Also if there are other instance of referring to the script, do the same there.

akondur (approved):

docs/SplunkOperatorUpgrade.md
vivekr-splunk: I retained this line from previous release. I am not sure if there is any change they need to do. unless they want to change the namespace from `splunk-operator` to something else.

smohan-splunk (approved):

PR #752, opened 2022-04-21
CSPL-1601: Manual Polling

jambrosiano:
Adding test of manual polling doing these steps:
- Deploy M4 SVA with app source containing several apps, and AppsRepoPollInterval=0 to have manual polling
- Ensure the manual poll configMap splunk-namespace-manual-app-update has been created, apps have been installed.
- Add newer versions of apps in app source.
- Disable Manual polling by setting AppsRepoPollInterva>0, and setting back status to ON in config map.
- Verify manual polling has not been triggered in the next few seconds and apps not updated.
- Verify apps are eventually updated after the end of AppsRepoPollInterval.

### DIFF
--- a/test/m4/appframework/appframework_test.go
+++ b/test/m4/appframework/appframework_test.go
@@ -2066,7 +2066,187 @@ var _ = Describe("m4appfw test", func() {
 
 			// Verify repo state is set  to 2 (i.e app deleted from S3 bucket)
 			testenv.VerifyAppRepoState(ctx, deployment, testcaseEnvInst, cm.Name, cm.Kind, appSourceNameIdxc, 2, appFileName[0])
+		})
+	})
 
+	Context("Multi Site Indexer Cluster with Search Head Cluster (M4) with App Framework", func() {
+		It("integration, m4, appframeworkm4, appframework: can deploy a M4 SVA, install apps via manual polling, switch to periodic polling, verify apps are not updated before the end of AppsRepoPollInterval, then updated after", func() {
+
+			/* Test Steps
+			   ################## SETUP ####################
+			   * Upload V1 apps to S3
+			   * Create app source with local scope for M4 SVA, AppsRepoPollInterval=0 to set apps polling as manual
+			   * Prepare and deploy M4 CRD with app framework and wait for pods to be ready
+			   ########## INITIAL VERIFICATION #############
+			   * Verify Apps Downloaded in App Deployment Info
+			   * Verify Apps Copied in App Deployment Info
+			   * Verify App Package is deleted from Operator Pod
+			   * Verify Apps Installed in App Deployment Info
+			   * Verify App Package is deleted from Splunk Pod
+			   * Verify apps are installed locally on Cluster Manager and Deployer
+			   * Verify status is 'OFF' in config map for Cluster Master and Search Head Cluster
+			   ######### SWITCH FROM MANUAL TO PERIODIC POLLING ############
+			   * Set AppsRepoPollInterval to 180 seconds for Cluster Master and Search Head Cluster
+			   * Change status to 'ON' in config map for Cluster Master and Search Head Cluster
+			   ############### UPGRADE APPS ################
+			   * Upgrade apps in app sources
+			   * Wait for pods to be ready
+			   ############ UPGRADE VERIFICATION ##########
+			   * Verify apps are not updated before the end of AppsRepoPollInterval duration
+			   * Verify apps are updated after the end of AppsRepoPollInterval duration
+			*/
+
+			//################## SETUP ####################
+			// Upload V1 apps to S3 for Indexer Cluster
+			appVersion := "V1"
+			appFileList := testenv.GetAppFileList(appListV1)
+			testcaseEnvInst.Log.Info(fmt.Sprintf("Upload %s apps to S3 for Indexer Cluster", appVersion))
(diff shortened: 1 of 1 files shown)


smohan-splunk:
From my understanding, the test case is achieving:
- App installation is triggered when manual polling is enabled (via configMap) the first time
- App installation is not triggered when manual polling is disabled

Does it make sense to also have a scenario where App installation is successful when manually triggered after the first time?

test/m4/appframework/appframework_test.go
```diff
@@ -2066,7 +2066,207 @@ var _ = Describe("m4appfw test", func() {
...
 
+			// Ensure Indexer cluster configured as multisite
+			testenv.IndexerClusterMultisiteStatus(ctx, deployment, testcaseEnvInst, siteCount)
+
+			// Ensure Search Head Cluster go to Ready phase
+			testenv.SearchHeadClusterReady(ctx, deployment, testcaseEnvInst)
+
+			// Get Pod age to check for pod resets later
+			splunkPodAge = testenv.GetPodsStartTime(testcaseEnvInst.GetName())
+
```
tpereirasplunk, line 2215: The app upload for v2 should be before this.

And then we should be checking at this point that the apps are not upgraded even though we modified config map to "on". 

The point of the test case would be that since `AppsRepoPollInterval` is set to non-zero. Modifying the configmap would not trigger an app poll

jambrosiano:
> 



> From my understanding, the test case is achieving:
> 
> * App installation is triggered when manual polling is enabled (via configMap) the first time
> * App installation is not triggered when manual polling is disabled
> 
> Does it make sense to also have a scenario where App installation is successful when manually triggered after the first time?

This scenario is actually covered in the test named "integration, m4, appframeworkm4, appframework: can deploy a M4 SVA with App Framework enabled for manual poll"

test/m4/appframework/appframework_test.go
```diff
@@ -2066,7 +2066,198 @@ var _ = Describe("m4appfw test", func() {
...
+			Expect(err).To(Succeed(), fmt.Sprintf("Unable to upload %s apps to S3 test directory for Indexer Cluster", appVersion))
+			uploadedApps = append(uploadedApps, uploadedFiles...)
+
+			// Upload V2 apps to S3 for Search Head Cluster
+			testcaseEnvInst.Log.Info(fmt.Sprintf("Upload %s apps to S3 for Search Head Cluster", appVersion))
+			uploadedFiles, err = testenv.UploadFilesToS3(testS3Bucket, s3TestDirShc, appFileList, downloadDirV2)
+			Expect(err).To(Succeed(), fmt.Sprintf("Unable to upload %s apps to S3 test directory for Search Head Cluster", appVersion))
+			uploadedApps = append(uploadedApps, uploadedFiles...)
+
+			// Check for changes in App phase to determine if next poll has been triggered
```
pdhanoyasplunk, line 2179: Would test fail here since manual poll is OFF ?

tpereirasplunk (approved):

test/m4/appframework/appframework_test.go
```diff
@@ -2066,7 +2066,199 @@ var _ = Describe("m4appfw test", func() {
...
+			splunkPodAge = testenv.GetPodsStartTime(testcaseEnvInst.GetName())
+
+			//########## UPGRADE VERIFICATIONS ############
+			testcaseEnvInst.Log.Info(" Verify apps are not updated before the end of AppsRepoPollInterval duration")
+			appVersion = "V1"
+			testenv.AppFrameWorkVerifications(ctx, deployment, testcaseEnvInst, allAppSourceInfo, splunkPodAge, "")
+
+			// Wait for the end of AppsRepoPollInterval duration
+			testcaseEnvInst.Log.Info("Wait for the end of AppsRepoPollInterval duration")
+			time.Sleep(2 * time.Minute)
```
sgontla, line 2247: I think, before reaching here, we can wait for some time (say 30 seconds), and still, make sure that the app is v1?

test/m4/appframework/appframework_test.go
```diff
@@ -2066,7 +2066,199 @@ var _ = Describe("m4appfw test", func() {
...
+			uploadedApps = append(uploadedApps, uploadedFiles...)
+
+			// Ensure that the Cluster Manager goes to Ready phase
+			testenv.ClusterManagerReady(ctx, deployment, testcaseEnvInst)
+
+			// Ensure the Indexers of all sites go to Ready phase
+			testenv.IndexersReady(ctx, deployment, testcaseEnvInst, siteCount)
+
+			// Ensure Indexer cluster configured as multisite
+			testenv.IndexerClusterMultisiteStatus(ctx, deployment, testcaseEnvInst, siteCount)
```
sgontla, line 2232: nit: Why do we need to do this?, as there are no pod resets involved....

test/m4/appframework/appframework_test.go
jambrosiano: We make sure apps are still in V1 just before that actually, inside AppFrameWorkVerifications on line 2243

test/m4/appframework/appframework_test.go
jambrosiano: Indeed this can be removed.

tpereirasplunk (approved):

smohan-splunk (approved):

PR #753, opened 2022-04-22
CSPL-1729: Detect and update the init container image

sgontla:
Whenever there is an update to the init container image spec, detect it and
update the modified image as part of the updated pod template spec


pkg/splunk/controller/util.go
```diff
@@ -140,6 +140,17 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name
 			"revised", len(revised.InitContainers))
 		current.InitContainers = revised.InitContainers
 		result = true
+	} else {
+		for idx := range current.InitContainers {
+			// check Image
```
smohan-splunk, line 145: Question about the for loop (here as well as later in this function). Once the result has been set to true, is there any reason for continuing parsing through the loop for other containers (or even other if checks in the same iteration)? May be we just break instead?

akondur: One benefit of continuing to parse the loop is that we log the differences which might help while debugging.

pkg/splunk/controller/util.go
```diff
@@ -140,6 +140,17 @@ func MergePodSpecUpdates(current *corev1.PodSpec, revised *corev1.PodSpec, name
...
+			if current.InitContainers[idx].Image != revised.InitContainers[idx].Image {
+				scopedLog.Info("Init Container Images differ",
+					"current", current.InitContainers[idx].Image,
+					"revised", revised.InitContainers[idx].Image)
+				current.InitContainers[idx].Image = revised.InitContainers[idx].Image
+				result = true
+			}
```
akondur, line 152: Should we also check for differences in other fields such as `ports`, `resources`, `env variables` etc on the init container?

akondur (approved):

pkg/splunk/controller/util.go
smohan-splunk: Good point. We should really have a dump functionality under debug (at the end of the function, if result is true), if we want to log the differences for every container.

sgontla: Idea is to update the current object, before updating it. So, we update all the specs, wherever there is a change.

pkg/splunk/controller/util.go
sgontla: Right now, we are not allowing any of these parameters(ports/resources, env, etc) through the CR. Image is the only field that we are taking from the config parameter, so, addressing that selectively for now.

pkg/splunk/controller/util.go
smohan-splunk: Ah yes, I missed the updates. 
> Directly writing the revised spec as an update can cause the exception as the resource version and other metadata differ.
Why is this not a concern when `len(current.Containers) != len(revised.Containers)`

smohan-splunk: Had an offline discussion with @sgontla .

Current containers can contain updates which the Operator may not be interested in and can cause unnecessary resets & undesirable behavior. To avoid this, we need to only compare the pre-defined specs of interest and hence the logic. Further we agreed that the condition under which the if statement `len(current.Containers) != len(revised.Containers)` could get executed & its effects need to be evaluated further (for example when Istio is deployed in the same namespace)

smohan-splunk (approved):

Issue #748, opened 2022-04-19
splunk operator repeatedly emits golang backtrace during ApplyIndexerCluster


mdaniel:
# What?

The logger includes golang backtrace for "Indexer cluster is not ready" over and over

# Why does this matter?

The backtrace output makes it look like splunk-operator is panicking and causes the operator to use more log storage than necessary given how often that line is emitted

# What should happen instead?

Log output without the backtrace, such as

```
2022-04-19T21:28:15.461Z        ERROR   controller.indexercluster.ApplyIndexerCluster   Indexer cluster is not ready    {"reconciler group": "enterprise.splunk.com", "reconciler kind": "IndexerCluster", "name": "example", "namespace": "splunk", "name": "orvn", "namespace": "splunk"}
```

# Versions
* splunk/splunk-operator:1.1.0

akondur:
Closing issue as it is fixed via [PR](https://github.com/splunk/splunk-operator/pull/755)

PR #755, opened 2022-04-22
CSPL-1763 - Fixing github issue for logging error unnecessarily

akondur:
Fixing github issue - https://github.com/splunk/splunk-operator/issues/748


smohan-splunk (commented):
The concern also seems to be repeated logging with disk consumption. Does changing from Error to Info throttle the logging frequency?

akondur:
> #748

I believe the issue mentions that the **backtrace** hogs more of the disk than necessary. The frequency of the logs are not reduced as the log is necessary to indicate the status of the cluster in the CR

smohan-splunk (approved):

jryb (approved):

PR #756, opened 2022-04-22
CSPL-1421 Implement LicenseManager CRD

mgaldino-splunk:
**Implement a LicenseManager CRD without Bias Language.**

In this PR we implement a new CRD for License Manager, which works in parallel with the existing LicenseMaster. The new CRD will eventually replace the obsolete LicenseMaster, once the customers had enough time to perform the migration. 

The PR consists of implementing new independent controllers for the new CRD, adding dual support for the new and obsolete License Manager, and redirection of testing and automation to the new CRD.

**Independent controllers:**
```
splunk-operator/
├── api
│   └── v3
│       ├── licensemanager_types.go
│       ├── licensemaster_types.go
├── controllers
│   └── common
│       ├── licensemanager_controller.go
│       ├── licensemaster_controller.go
├── pkg/splunk
│   └── enterprise
│       ├── licensemanager.go
│       ├── licensemaster.go
```

**Dual Support for shared references:**
`pkg/splunk/enterprise/configuration.go`
```
if instanceType != SplunkLicenseManager && spec.LicenseManagerRef.Name != "" {
		licenseManagerURL := GetSplunkServiceName(SplunkLicenseManager, spec.LicenseManagerRef.Name, false)
		< code >		
} else if instanceType != SplunkObsoleteLicenseManager && spec.ObsoleteLicenseManagerRef.Name != "" {
		ObsoletelicenseManagerURL := GetSplunkServiceName(SplunkObsoleteLicenseManager, spec.ObsoleteLicenseManagerRef.Name, false)
        < code >
}
```

**Replacements / Redirections:**
```
- spec := enterpriseApi.LicenseMasterSpec
+ spec := enterpriseApi.LicenseManagerSpec
```

**Testing Automation:**
We choose the Multisite (M4) License Manager Test to run on both the new and the obsolete CRD. 
All other tests will only run for the new CRD.


Dockerfile
```diff
@@ -1,6 +1,17 @@
 # Build the manager binary
 FROM golang:1.17 as builder
 
+RUN apt-get update && apt-get install -y ca-certificates openssl
```
akondur, line 4: Is there a reason for the extra certificate installs?

api/v3/common_types.go
```diff
@@ -104,8 +104,11 @@ type CommonSplunkSpec struct {
 	// Full path or URL for a Splunk Enterprise license file
 	LicenseURL string `json:"licenseUrl"`
 
-	// LicenseMasterRef refers to a Splunk Enterprise license manager managed by the operator within Kubernetes
-	LicenseMasterRef corev1.ObjectReference `json:"licenseMasterRef"`
+	// ObsoleteLicenseManagerRef refers to a Splunk Enterprise license manager managed by the operator within Kubernetes
+	ObsoleteLicenseManagerRef corev1.ObjectReference `json:"licenseMasterRef"`
```
akondur, line 108: The json tag(licenseMasterRef) is different from the field name(ObsoleteLicenseManagerRef). 

Two cents on this: 

IMO the intention here is to:

1.  Keep the conf exposed to the customer for LM ref the same i.e `licenseMasterRef`
2. Have the code refer the same with an `Obsolete*` tag.

Naming it `ObsoleteLicenseManager` seems a bit confusing to me. Maybe we can name it `ObsoleteLicenseMaster` or `LicenseMaster` thus making it:
- Similar to the json tag 
- Making the names unique enough for us to come back and remove in the next phase when we don't support the older LM.
- If using `ObsoleteLicenseMaster` spells out explicitly that it is obsolete.

api/v3/licensemaster_types.go
```diff
@@ -55,12 +55,12 @@ type LicenseMasterStatus struct {
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// LicenseMaster is the Schema for a Splunk Enterprise license manager.
+// LicenseMaster is the Schema for a Splunk Enterprise Obsolete license master.
```
akondur, line 58: Should we match the terminology with the spec [here](https://github.com/splunk/splunk-operator/pull/756/files#diff-809254b55d0140ed411813756f78738cb5836f972a111952db1078caec2e9bc3R108)? i.e 'Splunk Enterprise Obsolete license manager'.

api/v3/common_types.go
mgaldino-splunk: This is because we need to rename it without using the Bias language `Master`.  

It's a bit confusing because we can't change the Ref as long as the LicenseMaster CRD is supported, but at the same time using `Master` for variable names breaks the BiasLang effort. 

I choose `Obsolete*` as a term to mark every code that can be removed once the support for the LicenseMaster is deprecated. But we can discuss/think about other naming options as well.

api/v3/licensemaster_types.go
mgaldino-splunk: Resolved in the latest commit.

Dockerfile
mgaldino-splunk: Removed in the latest commit. Let's validate with the integration tests if this is needed or not. 

I had a local issue in which I needed those. Also during my testing, I noted some int-tests failures because these. I will update once the tests finish.

.biased_lang_exclude
```diff
@@ -1,6 +1,41 @@
...
 pkg/splunk/common/urls_testingcode.go
 pkg/splunk/common/paths.go
+PROJECT
+ChangeLog.md
+.rgignore
+# Exceptions for Obsolete CRDs
+config/crd/bases/
+bundle/manifests/
+config/rbac/role.yaml
+#Exceptions for dual support of license manager
```
mgaldino-splunk, line 12: Using specific files for exceptions rather than folders such `pkg/*`
Now with the new CRDs, we're tightening the grip on `biasLang` workflow.

controllers/licensemaster_controller.go
```diff
@@ -38,8 +38,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// LicenseMasterReconciler reconciles a LicenseMaster object
-type LicenseMasterReconciler struct {
+// ObsoleteLicenseManagerReconciler reconciles a LicenseMaster object
```
mgaldino-splunk, line 41: Used the term `obsolete` to mark all places to be removed when the LicenseMaster CRD is deprecated.

pkg/splunk/enterprise/configuration.go
```diff
@@ -630,15 +630,24 @@ func updateSplunkPodTemplateWithConfig(ctx context.Context, client splcommon.Con
 			Value: spec.LicenseURL,
 		})
 	}
```
mgaldino-splunk, line 632: During the time we support both CRDs it's possible a CR has both `LicenseMasterRef` and `LicenseManagerRef`. In that case, we default to `LicenseManagerRef` to support the migration.

pkg/splunk/common/urls_testingcode.go
```diff
@@ -93,62 +81,18 @@ const (
 	TestURLPeerHeadlessDecommission = "https://splunk-stack1-indexer-0.splunk-stack1-indexer-headless.test.svc.cluster.local:8089" + URIPeerDecommission
 )
 
-// ***** License Manager *****
```
mgaldino-splunk, line 96: By switching the testing code to `*Manager` we no longer need these exceptions. 
In the coming sprints, we will also review APIs that ES now supports with `*Manager`

api/v3/common_types.go
akondur: One thing I wanted to check on is whether the bias lang efforts are considering the json tags as well i.e LicenseMasterRef ? If yes, then we could still use ObsoleteLicenseMaster everywhere since the bias lang efforts would be broken either way? It should be fine either way - just an implementation choice :)

mgaldino-splunk: Yes, the effort considers json tags, so that's true it will only be completed once they are all removed. Maybe in 3.0 release? :D

We avoided changing the json tags because of the impact on existing deployments. Also, the reference is used more often than the tag itself, so we picked the one with more impact.

mgaldino-splunk: @akondur just FYI we're rolling back the Obsolete naming to keep the original name LicenseMasterRef.

Dockerfile
```diff
@@ -28,7 +28,7 @@ ENV OPERATOR=/manager \
     USER_UID=1001 \
     USER_NAME=nonroot
 
-RUN yum -y install shadow-utils
```
mgaldino-splunk, line 31: Please ignore. It won't be merged. This is a workaround for the int-test failures with 
```
#6 0.618 Updating Subscription Management repositories.
#6 0.618 Unable to read consumer identity
#6 0.622
#6 0.622 This system is not registered with an entitlement server. You can use subscription-manager to register.
```

api/v3/common_types.go
```diff
@@ -107,6 +107,9 @@ type CommonSplunkSpec struct {
 	// LicenseMasterRef refers to a Splunk Enterprise license manager managed by the operator within Kubernetes
 	LicenseMasterRef corev1.ObjectReference `json:"licenseMasterRef"`
 
+	// LicenseManagerRef refers to a Splunk Enterprise license manager managed by the operator within Kubernetes
+	LicenseManagerRef corev1.ObjectReference `json:"licenseManagerRef"`
```
vivekr-splunk, line 111: change this to optional field so one of the field is selected


`//+optoinal
LicenseMasterRef corev1.ObjectReference `json:"licenseMasterRef,omitempty"`

//+optional
LicenseManagerRef corev1.ObjectReference `json:"licenseManagerRef,omitempty"``

test/licensemaster/obsoletelm_m4_test.go
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2018-2022 Splunk Inc. All rights reserved.
```
vivekr-splunk, line 1: Not sure if we need to change the name here to obsolete

config/crd/patches/additional_supported_versions_patch_licensemanagers.yaml
```diff
@@ -0,0 +1,78 @@
+- op: add
```
vivekr-splunk, line 1: remove this we do not need patch for older version as this will be new CR

config/crd/bases/enterprise.splunk.com_licensemasters.yaml
```diff
@@ -1165,6 +1165,43 @@ spec:
                 - Always
                 - IfNotPresent
                 type: string
+              licenseManagerRef:
```
vivekr-splunk, line 1168: please check if this is optional field.

vivekr-splunk (changes requested):

controllers/licensemaster_controller.go
```diff
@@ -38,8 +38,8 @@ import (
...
+type ObsoleteLicenseManagerReconciler struct {
```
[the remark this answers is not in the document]
mgaldino-splunk: Resolved in the latest commit.

config/crd/bases/enterprise.splunk.com_licensemasters.yaml
```diff
@@ -14,23 +14,24 @@ spec:
     listKind: LicenseMasterList
     plural: licensemasters
     shortNames:
-    - lm
+    - obsoletelm
```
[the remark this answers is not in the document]
mgaldino-splunk: Resolved in the latest commit.

api/v3/common_types.go
mgaldino-splunk: Marked both *Master and *Manager as optional in the latest commit.

test/licensemaster/obsoletelm_m4_test.go
mgaldino-splunk: Resolved in the latest commit.

config/crd/patches/additional_supported_versions_patch_licensemanagers.yaml
mgaldino-splunk: Done

api/v3/common_types.go
akondur: Great, thanks for letting me know @mgaldino-splunk

akondur (approved):

PROJECT
```diff
@@ -34,6 +34,15 @@ resources:
   kind: LicenseMaster
   path: github.com/splunk/splunk-operator/api/v3
   version: v3
+- api:
+    crdVersion: v1
```
smohan-splunk, line 38: Although this is the first version of the LicenseManager CR, in the past, we have tried to align the new CRs with versions of the existing CRs (MC CRD is an example).
cc: @kashok-splunk , @vivekr-splunk

mgaldino-splunk: Agreed. I will double-check if we need to change something. Currently, it looks like this new CRD is set to v3.

mgaldino-splunk: Confirmed we don't need additional changes for now as the CRD is on `V3`. I think that `v1` above is related to `apiextensions` version. Our CRD is referenced in line 45 as `version: v3`. 
cc: @kashok-splunk , @vivekr-splunk

smohan-splunk (approved):

api/v3/licensemaster_types.go
```diff
@@ -55,10 +55,10 @@ type LicenseMasterStatus struct {
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// LicenseMaster is the Schema for a Splunk Enterprise license manager.
+// LicenseMaster is the Schema for a Splunk Enterprise license master.
```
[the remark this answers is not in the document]
mgaldino-splunk: These are to avoid conflicts between the two CRDs. In a few cases, we had to move the master back to allow the new manager labels to be unique.  We chose to roll them back in the *master* files since they'll eventually be removed.

pkg/splunk/common/urls.go
```diff
@@ -98,6 +98,12 @@ const (
 	//LicenseManager = "license-master"
 	LicenseManager = "license-master"
 
+	// LicenseManagerRole = "splunk_license_master"
+	LicenseManagerRole = "splunk_license_master"
+
+	// LicenseManagerURL = "SPLUNK_LICENSE_MASTER_URL"
+	LicenseManagerURL = "SPLUNK_LICENSE_MASTER_URL"
+
```
jryb, line 106: Should we create a followup issue for this once they are fixed within `splunk-ansible`?

test/licensemaster/enterprise.lic
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<license>
+  <signature>kGsVWVk5lbo7V6hqKzzXSKmw0KegBMVD3d5J41ddH6Jv+jIEZlqXaz9b5A51HapnaBgi5N/OzWNUxMBVwkBqadeWjgOk385Z1Sn6M5HPq5NYcjrhYAn/zDIILtGw0vHusVLprA0qBsz7Jz9DaCcPgidIOZGC/f/F+k3p/f6IrqMU9X9fDi1xvdq25wIQsiaHZ+vUGEFTmPzu+m8T+pQPICOG8Kw0eYr3MAolpkTKGS6I7Sf8EAMe9j9h6uLI09SEyJdhhH+pmuxEXyaJ8gX58KRzCUVKu6SSTwg1uns43BTVQ85FD7j57c3aSQ05QkvraCYhSdM7YS7mV94r0XWCJw==</signature>
```
jryb, line 3: Do we really want this license in the code base?  I guess its expired now but still...

mgaldino-splunk: Good catch Jeff. Removing it in the latest commit.

.github/workflows/int-test-workflow.yml
```diff
@@ -5,6 +5,7 @@ on:
       - develop
       - master
       - feature-appframework
+      - CSPL-1421-LicenseManager
```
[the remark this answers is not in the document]
mgaldino-splunk: Correct. Before merging we will remove it.
We're keeping it for now just to run all Integration-test on this branch.

jryb (approved):

kashok-splunk (approved):

pkg/splunk/common/urls.go
mgaldino-splunk: Absolutely. We're also reviewing this exception list to remove all the APIs that now ES supports with `/manager/`. But the complete list will come with the ClusterManager CRD (next PR).

pkg/splunk/enterprise/licensemanager.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		scopedLog.Error(err, "Failed to validate license manager spec")
+		return result, err
+	}
+
+	// If needed, Migrate the app framework status
+	err = checkAndMigrateAppDeployStatus(ctx, client, cr, &cr.Status.AppContext, &cr.Spec.AppFrameworkConfig, true)
+	if err != nil {
+		return result, err
+	}
```
sgontla, line 59: Nit: This is not the case for us as there the License Manager CRD is coming for the first time, but, it may be useful to maintain the consistency.

api/v3/licensemaster_types.go
```diff
@@ -55,10 +55,10 @@ type LicenseMasterStatus struct {
...
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:path=licensemasters,scope=Namespaced,shortName=lm
+// +kubebuilder:resource:path=licensemasters,scope=Namespaced,shortName=lmaster
```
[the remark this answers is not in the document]
mgaldino-splunk: If the CR is already deployed this shouldn't have an impact. For new deployments, we're encouraging as much as possible to use the new CRD moving forward, we even considered not releasing the updated `master` CRD at all. 

That said I think it's better to keep it consistent/backwards compatible so the new CRD should use a new name while keeping `lm` for the old `master` CRD.  The downside is that `lm` will be lost after the `Master` is deprecated :(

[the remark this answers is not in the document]
mgaldino-splunk: Rolled back to `lm` in the latest commit.

config/crd/bases/enterprise.splunk.com_licensemasters.yaml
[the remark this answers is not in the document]
mgaldino-splunk: This happens because `licenseM*Ref` is part of the `CommonSplunkSpec` so it comes out in all Splunk CRDs. This is the same behavior as all other Refs e.g. `clusterMasterRef` is part of the [Standalone CRD](https://github.com/splunk/splunk-operator/blob/93edb2ca16aba6bbab1ef3accd54532416db812f/config/crd/bases/enterprise.splunk.com_standalones.yaml#L966). There was no way around it without bigger changes. 

To avoid these cases we need to break out `CommonSplunkSpec` to fit only what's common to all Splunk CRDs. Maybe this type of refactoring can be done in the future.

config/examples/advanced/c1.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: enterprise.splunk.com/v3
-kind: LicenseMaster
+kind: LicenseManager
```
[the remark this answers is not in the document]
mgaldino-splunk: We didn't want to duplicate all tests ( And the effort to support them). We chose a few relevant tests to run on both CRDs, and everything else only runs for the `*Manager` CRD. The coverage might be temporarily lower, but not by much. Let me get the numbers.

docs/CustomResources.md
```diff
@@ -8,7 +8,7 @@ you can use to manage Splunk Enterprise deployments in your Kubernetes cluster.
   - [Metadata Parameters](#metadata-parameters)
   - [Common Spec Parameters for All Resources](#common-spec-parameters-for-all-resources)
   - [Common Spec Parameters for Splunk Enterprise Resources](#common-spec-parameters-for-splunk-enterprise-resources)
-  - [LicenseMaster Resource Spec Parameters](#licensemaster-resource-spec-parameters)
```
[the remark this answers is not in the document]
mgaldino-splunk: We want to direct users to the new CRD as much as possible. The documentation was all switched to cover only the `Manager`

bundle/manifests/enterprise.splunk.com_clustermasters.yaml
```diff
@@ -3802,6 +3842,66 @@ spec:
     storage: true
     subresources:
       status: {}
+  - name: v1
```
mgaldino-splunk, line 3845: The bundle updates were generated after running `make bundle`. I'm validating if these should or not be part of the PR. 
CC: @vivekr-splunk , @sgontla

pkg/splunk/enterprise/licensemanager.go
mgaldino-splunk: Initially, both CRDs are identical to avoid confusion.

config/examples/advanced/c1.yaml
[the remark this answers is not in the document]
mgaldino-splunk: Coverage decreased (-0.01%) to 83.031%

sgontla (approved):

vivekr-splunk (approved):

.github/workflows/nightly-int-test-workflow.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test: [appframeworks1, appframeworkc3, appframeworkm4, secret, smartstore, monitoringconsole, scaling, crcrud, licensemaster]
```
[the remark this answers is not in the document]
mgaldino-splunk: For the testing code, the `licensemanager` now has tests for both CRDs.

pkg/splunk/enterprise/util.go
```diff
@@ -257,23 +257,45 @@ func getStandaloneExtraEnv(cr splcommon.MetaObject, replicas int32) []corev1.Env
 }
 
 // getLicenseManagerURL returns URL of license manager
-func getLicenseManagerURL(ctx context.Context, cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunkSpec) []corev1.EnvVar {
+func getLicenseManagerURL(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunkSpec) []corev1.EnvVar {
```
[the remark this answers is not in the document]
mgaldino-splunk: Will review and consider changes in the next PR.

PR #770, opened 2022-04-30
CSPL-1765: fixing a log format that is causing the panic

sgontla:
fixing a log format that is causing the panic

### DIFF
--- a/pkg/splunk/util/secrets.go
+++ b/pkg/splunk/util/secrets.go
@@ -482,8 +482,7 @@ func ApplyNamespaceScopedSecretObject(ctx context.Context, client splcommon.Cont
 	}
 
 	// Make data
-	scopedLog.Info("Namespace scoped secret does not exist, creating it ",
-		"and filling it with new values for all token types")
+	scopedLog.Info("Namespace scoped secret does not exist, creating and filling it with new values for all token types")
 	current.Data = make(map[string][]byte)
 	// Not found, update data by generating values for all types of tokens
 	for _, tokenType := range splcommon.GetSplunkSecretTokenTypes() {


jryb (approved):

pkg/splunk/util/secrets.go
```diff
@@ -482,8 +482,7 @@ func ApplyNamespaceScopedSecretObject(ctx context.Context, client splcommon.Cont
 	}
 
 	// Make data
-	scopedLog.Info("Namespace scoped secret does not exist, creating it ",
-		"and filling it with new values for all token types")
+	scopedLog.Info("Namespace scoped secret does not exist, creating and filling it with new values for all token types")
```
akondur, line 485: Is this format causing a panic? Are there logs for the panic?

sgontla: @akondur , right, this is causing the panic with the new SDK (and I guess mostly with the new lib packages). As the interface implementation expects the odd number of arguments (msg[, KV pairs]) format, it is doing a panic. In older releases, it was throwing a warning and continuing. 

I think, as a permanent fix, we need to add a format checking logic(something like `make log-fmt`), that validates all the log entries and finds out any deviations from:
1. error: even number of arguments
2. info: odd number of entries

sgontla: @jambrosiano, can you share a panic log example here.

akondur (approved):

pkg/splunk/util/secrets.go
akondur: To add to this maybe we should monitor operator logs during tests for any PANIC's? @pdhanoyasplunk @jambrosiano @tpereirasplunk wonder if this is possible.

smohan-splunk (approved):

Issue #765, opened 2022-04-28
improvement: add default-container annotation to splunk-operator


mdaniel:
Per [the documentation](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container), if the Pod were annotated with `kubectl.kubernetes.io/default-container`, then `kubectl logs deploy/splunk-operator-controller-manager` would have a reasonable default

# What happens right now?

```console
$ kubectl -n splunk-operator logs deploy/splunk-operator-controller-manager
error: a container name must be specified for pod splunk-operator-controller-manager-6465874c68-lxn95, choose one of: [kube-rbac-proxy manager]
```

# What should happen?

```console
$ kubectl -n splunk-operator patch -p '{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/default-container": "manager"}}}}}' deploy/splunk-operator-controller-manager
$ kubectl -n splunk-operator rollout status deploy/splunk-operator-controller-manager
deployment "splunk-operator-controller-manager" successfully rolled out
$ kubectl -n splunk-operator logs deploy/splunk-operator-controller-manager
2022-04-28T22:04:25.519Z        INFO    setup   Manager will be watching namespace      {"namespace": ""}
...
```

akondur:
Hi @mdaniel , when I tried the following commands:

```
kubectl -n splunk-operator patch -p '{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/default-container": "manager"}}}}}' deploy/splunk-operator-controller-manager

kubectl -n splunk-operator rollout status deploy/splunk-operator-controller-manager
```

I still see the following error:
```
$ kubectl -n splunk-operator logs deploy/splunk-operator-controller-manager                                                                                                                                                                                                                                                   [252/9274]
error: a container name must be specified for pod splunk-operator-controller-manager-66d5456c6b-n7bmn, choose one of: [kube-rbac-proxy manager]
```

However, if I patch the following annotation:
```
kubectl -n splunk-operator patch -p '{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/default-logs-container": "manager"}}}}}' deploy/splunk-operator-controller-manager
```

I am able to:

**Display logs from deployment without mentioning the container:**

```
$ kubectl -n splunk-operator logs deploy/splunk-operator-controller-manager
[...]

PR #777, opened 2022-05-02
CSPL- 1768 - Adding an annotation to define a default container

akondur:
Adding an annotation to define a default container for the operator deployment pod. Addressing github enhancement [request](https://github.com/splunk/splunk-operator/issues/765)

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -25,6 +25,8 @@ spec:
       labels:
         control-plane: controller-manager
         name: splunk-operator
+      annotations:
+        kubectl.kubernetes.io/default-logs-container: manager
     spec:
       securityContext:
         runAsUser: 1001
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -72,22 +74,9 @@ spec:
...
-        #  - mountPath: /opt/splunk/appframework/
-        #    name: app-staging
       serviceAccountName: controller-manager
       volumes:
       - configMap:
           name: config
         name: splunk-operator-config
-      #  - name:  app-staging
-      #    persistentVolumeClaim:
-      #      claimName: tmp-app-download
```
sgontla, line 92: same as above

config/manager/manager.yaml
```diff
@@ -72,22 +74,9 @@ spec:
...
-        #resources:
-        #  limits:
-        #    cpu: 1000m
-        #    memory: 1000Mi
-        #  requests:
-        #    cpu: 1000m
-        #    memory: 1000Mi
-        #volumeMounts:
-        #  - mountPath: /opt/splunk/appframework/
-        #    name: app-staging
```
sgontla, line 84: Nit: Having the `volumeMounts:` may be handy for the customer, as we are recommending it.

akondur: But it's commented out? Are we recommending a pvc for app-framework?

sgontla: I think @vivekr-splunk commented on it for now. Ideally, it should have gone to 2.0.0 directly. Even on 2. x, it will be commented, being it is a recommendation in the case of app framework. Also, it needs the customer to create the PVC first, before uncommenting this.

akondur: If its going to be commented out, I believe it doesn't get exported to `splunk-operator-install.yaml`? Maybe it should be a part of the documentation for app-framework such that they can edit the operator deployment as well as create a pvc i.e we could provide a `kubectl patch` command for the deployment as well as a simple yaml for pvc?

vivekbasappa: we do not need this in develop for now. when we merge appframework branch it will be enabled.

sgontla (approved):

kashok-splunk (approved):

smohan-splunk (approved):

vivekr-splunk (approved):

PR #780, opened 2022-05-03
CSPL-1777: Installing UBI security patch docker

vivekr-splunk:
ubi security patch to docker images

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ ENV OPERATOR=/manager \
 
 RUN yum -y install shadow-utils
 RUN useradd -ms /bin/bash nonroot -u 1001
+RUN yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical
 
 LABEL name="splunk" \
       maintainer="support@splunk.com" \


Dockerfile
```diff
@@ -30,6 +30,7 @@ ENV OPERATOR=/manager \
 
 RUN yum -y install shadow-utils
 RUN useradd -ms /bin/bash nonroot -u 1001
+RUN yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical
```
akondur, line 33: Hi @vivekr-splunk , would you mind explaining the importance of update-minimal here or in the description?

vivekr-splunk: I removed minimal as i was testing installing shadow-utils which i was unable to do it , moved to latest that worked. we can move back if needed

akondur (approved):

Dockerfile
vivekr-splunk: to only install security errata use `yum update-minimal`

sgontla (approved):

jryb (approved):

kashok-splunk (approved):

Dockerfile
smohan-splunk, line 33: How can we confirm if the security patch has resolved the vulnerability? Or can this be done only during release time?

Issue #762, opened 2022-04-27
The /opt/splk/etc volume used by initContainer is not defined


mdaniel:
# Summary

In [`pkg/splunk/enterprise/util.go`'s setupInitContainer](https://github.com/splunk/splunk-operator/blob/1.1.0/pkg/splunk/enterprise/util.go#L471) makes reference to a volume named `pvc-etc` that it fails to define

# What happens?

The StatefulSet remains in Pending waiting for the volume to appear

> Warning  FailedCreate  36s (x16 over 3m20s)  statefulset-controller  create Pod splunk-cm-cluster-master-0 in StatefulSet splunk-cm-cluster-master failed error: Pod "splunk-cm-cluster-master-0" is invalid: spec.initContainers[0].volumeMounts[0].name: Not found: "pvc-etc"

# What should happen?

The `setupInitContainer` should inject a `{ name: pvc-etc, emptyDir: {} }` into the `volumes:` of the `PodSpec`

If I had a magic wand, I'd also name that `initc-pvc-etc` since that `/opt/splk/etc` seems to be exclusively used by the `initContainer` and thus `pvc-etc` sounds more generic than in actually is

pogdin:
Thanks @mdaniel.  Two questions. Are you using `ephemeralStorage: true` when observing this and if yes, do you see the same problem without it (using gp2 etc if that is an option).

mdaniel:
I am using `ephemeralStorage:`, yes, but before reporting this I made sure to `git grep splk` to see if my local setup was causing this and the answer appeared to be no, it genuinely was missing

```console
$ git checkout 1.1.0
HEAD is now at be424de4 Merge pull request #732 from splunk/update-release-artifact-upload

$ git grep splk
pkg/splunk/common/paths.go:     OperatorClusterManagerAppsLocal = "/opt/splk/etc/master-apps/splunk-operator/local"
pkg/splunk/common/paths.go:     OperatorClusterManagerAppsLocalIndexesConf = "/opt/splk/etc/master-apps/splunk-operator/local/indexes.conf"
pkg/splunk/common/paths.go:     OperatorClusterManagerAppsLocalServerConf = "/opt/splk/etc/master-apps/splunk-operator/local/server.conf"
[...]

PR #781, opened 2022-05-04
CSPL-1769 - Change the naming of volumeMounts to adopt setup of init container

akondur:
Addressing github [issue](https://github.com/splunk/splunk-operator/issues/762). 

**Context:**
The setup of initContainer for Smartstore fails with error:
```
splunk-cm-example-cluster-master-0 in StatefulSet splunk-cm-example-cluster-master failed error:
Pod "splunk-cm-example-cluster-master-0" is invalid: spec.initContainers[0].volumeMounts[0].name:
Not found: "pvc-etc"
```

when the following example spec is added to use ephemeral volumes:

```
kind: ClusterMaster
metadata:
  name: cm-example
  namespace: splunk-operator
  finalizers:
spec:
  etcVolumeStorageConfig:
    ephemeralStorage: true
  varVolumeStorageConfig:
    ephemeralStorage: true
  smartstore:
    defaults:
        volumeName: pogdin_smartstore
    indexes:
      - name: main
        remotePath: $_index_name
        volumeName: pogdin_smartstore
      - name: cloudwatch
        remotePath: $_index_name
        volumeName: pogdin_smartstore
    volumes:
      - name: pogdin_smartstore
        path: pogdin/smartstore
        secretRef: s3-secret
```

The function `setupInitContainer` uses `pvc-etc` as the volume mount name while the ephemeral volume mount name is different(`mnt-splunk-etc`) & hence the statefulSet controller in K8s cannot find a non-existent volume `pvc-etc`

**Solution**

Change the naming of volume mount to the format `mnt-splunk-<volume_type i.e pvc or ephemeral>-<etc or var volume>`. The setupInitContainer function uses the same nomenclature to name its volume mount such that it matches with the volume created in either `addEphemeralVolumes` or `addPvcVolumes`

With the change the CM pod with above example spec comes up fine:
```
$ kubectl get pods -w
NAME                                                  READY   STATUS    RESTARTS   AGE
splunk-operator-controller-manager-578757f448-h9q69   1/2     Running   0          13s
splunk-operator-controller-manager-578757f448-h9q69   2/2     Running   0          20s
splunk-cm-example-cluster-master-0                    0/1     Pending   0          0s
splunk-cm-example-cluster-master-0                    0/1     Pending   0          0s
splunk-cm-example-cluster-master-0                    0/1     Init:0/1   0          0s
splunk-cm-example-cluster-master-0                    0/1     PodInitializing   0          1s
splunk-cm-example-cluster-master-0                    0/1     Running           0          2s
splunk-cm-example-cluster-master-0                    1/1     Running           0          100s
```

Corrected typo for ephemeral volumes as well.


smohan-splunk (approved):

jryb (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -460,15 +460,24 @@ func ApplySmartstoreConfigMap(ctx context.Context, client splcommon.ControllerCl
...
-func setupInitContainer(podTemplateSpec *corev1.PodTemplateSpec, Image string, imagePullPolicy string, commandOnContainer string) {
+func setupInitContainer(podTemplateSpec *corev1.PodTemplateSpec, Image string, imagePullPolicy string, commandOnContainer string, isEtcVolEph bool) {
+	var volMntName string
+
+	// Populate the volume mount name based on volume type(eph, pvc) and use /opt/splk/etc for init container
+	if isEtcVolEph {
+		volMntName = fmt.Sprintf(splcommon.SplunkMountNamePrefix, splcommon.SplunkMountTypeEph, splcommon.EtcVolumeStorage)
+	} else {
+		volMntName = fmt.Sprintf(splcommon.SplunkMountNamePrefix, splcommon.SplunkMountTypePvc, splcommon.EtcVolumeStorage)
+	}
```
sgontla, line 471: Can we make it a separate function, for long-term ease of maintenance?
Also, the caller can directly pass the return value of that new function, so that, it will be re-used across the places.

akondur: Hi @sgontla , can you suggest the definition of the function? Usually the `SplunkMountNamePrefix` string is created as the volume type is known. Here its a special case where we need to pass it over and hence the logic applies only here.

sgontla: @akondur, my suggestion was to use a function, that derives the `volMntName`  based on the storage type. That way, we don't need to use the `fmt.Sprintf()` across several places. Probably, something like `getEtcMountNameByStorageType()` & `getVarMountNameByStorageType()`, which takes the storage type(ideally, an enum better serves for etc and var), and returns the mount name, that can be used across multiple containers.

akondur: Hi @sgontla , I have created a [task](https://splunk.atlassian.net/browse/CSPL-1793) to enhance the nomenclature of the storage config code in the operator. Will be taking care of it in the next sprint as discussed.

sgontla (approved):

PR #786, opened 2022-05-11
Update the fsGroup context to 1001 on Operator Pod.

sgontla:
Update the fsGroup context to 1001, so that the mounted volume is accessible to the app framework.

### DIFF
--- a/bundle/manifests/splunk-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/splunk-operator.clusterserviceversion.yaml
@@ -488,6 +488,7 @@ spec:
                   allowPrivilegeEscalation: false
               securityContext:
                 runAsUser: 1001
+                fsGroup: 1001
               serviceAccountName: splunk-operator-controller-manager
               terminationGracePeriodSeconds: 10
       permissions:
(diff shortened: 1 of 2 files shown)


bundle/manifests/splunk-operator.clusterserviceversion.yaml
```diff
@@ -488,6 +488,7 @@ spec:
                   allowPrivilegeEscalation: false
               securityContext:
                 runAsUser: 1001
+                fsGroup: 1001
```
akondur, line 491: Hi @sgontla , when does this file come into the picture? Isn't [manager.yml](https://github.com/splunk/splunk-operator/blob/feature-appframework/config/manager/manager.yaml) used to generate the splunk-operator-install.yaml? :)

akondur (approved):

bundle/manifests/splunk-operator.clusterserviceversion.yaml
sgontla: you are right. I think I checked out the wrong file before the commit :-). I couldn't test myself due to the docker issue.

PR #787, opened 2022-05-11
integration workflow changes with git checkin image

vivekr-splunk:
- SHA Digest will be created based on check-in commit id. 
- run all smoke and integration test to above image
- use the same image for `clair` vulnerability testing.

### DIFF
--- a/.github/workflows/build-test-push-workflow.yml
+++ b/.github/workflows/build-test-push-workflow.yml
@@ -151,7 +151,7 @@ jobs:
       S3_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
       ENTERPRISE_LICENSE_LOCATION: ${{ secrets.ENTERPRISE_LICENSE_LOCATION }}
       EKS_SSH_PUBLIC_KEY: ${{ secrets.EKS_SSH_PUBLIC_KEY }}
-      CLUSTER_WIDE: "true"
+      CLUSTER_WIDE: "false"
     steps:
       - name: Set Test Cluster Name
         run: | 
--- a/.github/workflows/build-test-push-workflow.yml
+++ b/.github/workflows/build-test-push-workflow.yml
@@ -234,7 +234,6 @@ jobs:
       - name: Run smoke test
         id: smoketest
         run: |
-          export SPLUNK_OPERATOR_IMAGE_NAME=${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA
           export SPLUNK_OPERATOR_IMAGE=${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA
           make int-test
       - name: Collect Test Logs
--- a/.github/workflows/build-test-push-workflow.yml
+++ b/.github/workflows/build-test-push-workflow.yml
@@ -83,12 +83,11 @@ jobs:
       uses: aws-actions/amazon-ecr-login@v1
     - name: Make Splunk Operator Image
       run: |
-        docker pull registry.access.redhat.com/ubi8/ubi-minimal:latest
-        docker build -t ${{ secrets.ECR_REPOSITORY }}/${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA .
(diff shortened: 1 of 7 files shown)


tpereirasplunk (approved):

.github/workflows/build-test-push-workflow.yml
```diff
@@ -238,6 +234,8 @@ jobs:
       - name: Run smoke test
         id: smoketest
         run: |
+          export SPLUNK_OPERATOR_IMAGE_NAME=${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA
+          export SPLUNK_OPERATOR_IMAGE=${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA
```
jryb, line 238: We have 2 env variables that are set to the same thing?

jryb (approved):

.github/workflows/build-test-push-workflow.yml
```diff
@@ -83,12 +83,11 @@ jobs:
       uses: aws-actions/amazon-ecr-login@v1
     - name: Make Splunk Operator Image
       run: |
-        docker pull registry.access.redhat.com/ubi8/ubi-minimal:latest
-        docker build -t ${{ secrets.ECR_REPOSITORY }}/${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA .
+        make docker-build IMG=${{ secrets.ECR_REPOSITORY }}/${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA
```
sgontla, line 86: Just a thought, Is there a way, we can use the `Install Operator SDK` from the make file? I think that is where some arch-specific values are there, and using it directly from the make file will be a lot of value in terms of maintaining all the supported platforms.

vivekbasappa: we have plans to add that subba. I am still working on it. since testing on pipeline taking sometime It takes time.

.github/workflows/build-test-push-workflow.yml
vivekbasappa: I will remove one of them. export SPLUNK_OPERATOR_IMAGE_NAME=${{ env.SPLUNK_OPERATOR_IMAGE_NAME }}:$GITHUB_SHA will be removed

test/m4/appframework/appframework_test.go
```diff
@@ -258,7 +258,7 @@ var _ = Describe("m4appfw test", func() {
 			uploadedApps = append(uploadedApps, uploadedFiles...)
 
 			// Wait for the poll period for the apps to be downloaded
-			time.Sleep(2 * time.Minute)
+			time.Sleep(4 * time.Minute)
```
tpereirasplunk, line 261: this is a timer between app polling. Not related to the timer for phase. Those timer values are  here (https://github.com/splunk/splunk-operator/blob/master/test/testenv/testenv.go#L64)

tpereirasplunk, line 261: this is a timer between app polling. Not related to the timer for phase. Those timer values are  here (https://github.com/splunk/splunk-operator/blob/master/test/testenv/testenv.go#L64)

.github/workflows/build-test-push-workflow.yml
sgontla: One other thing is, with this change, seems like we are duplicating the UT, as there is a separate UT check item. Probably,  the `test` phony can be decoupled from this path?

vivekr-splunk: this was always there , i changed the way to use make command and COMMIT id (which is in GITHUB_SHA)

test/m4/appframework/appframework_test.go
vivekr-splunk: I have updated the information , this is temporary fix, just to see if this fixes the issue,

vivekr-splunk: I have updated the information , this is temporary fix, just to see if this fixes the issue,

PR #789, opened 2022-05-12
CSPL-1600: Insufficient disk space on operator PV

jambrosiano:
Test added to S1,C3 and M4 SVA of insufficient disk space on operator PV.
By default, Operator PV size is 1GB, so S1 test download 15 100MB-size apps in parallel and verifies all apps are eventually installed. For C3 and M4, it's 15 apps for Indexer Cluster and 15 for Search Head Cluster.

### DIFF
--- a/test/c3/appframework/appframework_test.go
+++ b/test/c3/appframework/appframework_test.go
@@ -2806,7 +2806,95 @@ var _ = Describe("c3appfw test", func() {
 			shcAppSourceInfo.CrAppFileList = testenv.GetAppFileList(appListV2)
 			allAppSourceInfo = []testenv.AppSourceInfo{cmAppSourceInfo, shcAppSourceInfo}
 			testenv.AppFrameWorkVerifications(ctx, deployment, testcaseEnvInst, allAppSourceInfo, splunkPodAge, clusterManagerBundleHash)
+		})
+	})
 
+	Context("Clustered deployment (C3 - clustered indexer, search head cluster)", func() {
+		It("c3, integration, appframeworkc3, appframework: can deploy a C3 SVA and install a bigger volume of apps than the operator PV disk space", func() {
+
+			/* Test Steps
+			   ################## SETUP ####################
+			   * Upload 15 apps of 100MB size each to S3 for Indexer Cluster and Search Head Cluster for cluster scope
+			   * Create app sources for Cluster Manager and Deployer with cluster scope
+			   * Prepare and deploy C3 CRD with app framework and wait for the pods to be ready
+			   ######### INITIAL VERIFICATIONS #############
+			   * Verify Apps are Downloaded in App Deployment Info
+			   * Verify Apps Copied in App Deployment Info
+			   * Verify App Package is deleted from Operator Pod
+			   * Verify Apps Installed in App Deployment Info
+			   * Verify App Package is deleted from Splunk Pod
+			   * Verify bundle push is successful
+			   * Verify  apps are copied, installed on Search Heads and Indexers pods
+			*/
+
+			//################## SETUP ####################
+			// Download 15 apps around 100MB each (Total 1.4GB) to have total volume above Operator PV default size (1GB)
+			appVersion := "V1"
+			appList := testenv.PVTestApps
+			appFileList := testenv.GetAppFileList(appList)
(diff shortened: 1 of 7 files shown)


test/c3/appframework/appframework_test.go
```diff
@@ -2806,7 +2806,95 @@ var _ = Describe("c3appfw test", func() {
...
+			//################## SETUP ####################
+			// Download 15 apps around 100MB each (Total 1.4GB) to have total volume above Operator PV default size (1GB)
+			appVersion := "V1"
+			appList := testenv.PVTestApps
+			appFileList := testenv.GetAppFileList(appList)
+			err := testenv.DownloadFilesFromS3(testDataS3Bucket, s3PVTestApps, downloadDirPVTestApps, appFileList)
+			Expect(err).To(Succeed(), "Unable to download app files")
+
+			// Upload apps to S3 for Indexer Cluster
+			testcaseEnvInst.Log.Info("Upload 15 apps around 100MB each (Total 1.4GB) for Indexer Cluster to be above Operator PV default size (1GB)")
```
sgontla, line 2839: One other alternative(a better alternative ?) is, to fill the local partition on the operator pod, so that we don't need to upload such a large file to S3:
Example:
`dd if=/dev/zero of=./gentoo_root.img bs=4k iflag=fullblock,count_bytes count=1G`

test/c3/appframework/appframework_test.go
```diff
@@ -2806,7 +2806,95 @@ var _ = Describe("c3appfw test", func() {
...
+
+			// Upload apps to S3 for Search Head Cluster
+			testcaseEnvInst.Log.Info("Upload 15 apps around 100MB each (Total 1.4GB) for Search Head Cluster to be above Operator PV default size (1GB)")
+			s3TestDirShc := "c3appfw-shc-" + testenv.RandomDNSName(4)
+			uploadedFiles, err = testenv.UploadFilesToS3(testS3Bucket, s3TestDirShc, appFileList, downloadDirPVTestApps)
+			Expect(err).To(Succeed(), fmt.Sprintf("Unable to upload %s apps to S3 test directory for Search Head Cluster", appVersion))
+			uploadedApps = append(uploadedApps, uploadedFiles...)
+
+			// Maximum apps to be downloaded in parallel
+			maxConcurrentAppDownloads := 30
```
sgontla, line 2853: Similarly, less load on the download part as well.

test/m4/appframework/appframework_test.go
```diff
@@ -2377,4 +2377,90 @@ var _ = Describe("m4appfw test", func() {
...
+			   * Verify Apps Copied in App Deployment Info
+			   * Verify App Package is deleted from Operator Pod
+			   * Verify Apps Installed in App Deployment Info
+			   * Verify App Package is deleted from Splunk Pod
+			   * Verify bundle push is successful
+			   * Verify  apps are copied, installed on Search Heads and Indexers pods
+			*/
+
+			//################## SETUP ####################
+			// Download 15 apps around 100MB each (Total 1.4GB) to have total volume above Operator PV default size (1GB)
```
sgontla, line 2400: same as abovve.

test/c3/appframework/appframework_test.go
smohan-splunk: Or could we test with a smaller PVC for this testcase?

jambrosiano: > Or could we test with a smaller PVC for this testcase?

We actually tried to edit the PV and set a smaller value but it seems 1G is the minimum here, lower values would be ignored.

pdhanoyasplunk (approved):

test/c3/appframework/appframework_test.go
jambrosiano: > One other alternative(a better alternative ?) is, to fill the local partition on the operator pod, so that we don't need to upload such a large file to S3: Example: `dd if=/dev/zero of=./gentoo_root.img bs=4k iflag=fullblock,count_bytes count=1G`

We will use another task to implement this suggestion later and can merge the current version of the test meanwhile.

sgontla (approved):

smohan-splunk (approved):

PR #800, opened 2022-05-23
CSPL-1598: Delete apps from init-apps while app installation is in progress

tpereirasplunk:


test/c3/appframework/appframework_test.go
```diff
@@ -2892,4 +2892,106 @@ var _ = Describe("c3appfw test", func() {
...
+			shReplicas := 3
+			cm, _, shc, err := deployment.DeploySingleSiteClusterWithGivenAppFrameworkSpec(ctx, deployment.GetName(), indexerReplicas, true, appFrameworkSpecIdxc, appFrameworkSpecShc, "", "")
+			Expect(err).To(Succeed(), "Unable to deploy Single Site Indexer Cluster with Search Head Cluster")
+
+			// Verify App Download is completed on Cluster Manager
+			testenv.VerifyAppState(ctx, deployment, testcaseEnvInst, deployment.GetName(), cm.Kind, appSourceNameIdxc, appFileList, enterpriseApi.AppPkgPodCopyComplete, enterpriseApi.AppPkgPodCopyPending)
+
+			//Delete apps from init-apps when app download is complete
+			var opPod string
+			if testcaseEnvInst.IsOperatorInstalledClusterWide() != "true" {
```
pdhanoyasplunk, line 2957: Leaving here for discussion. I think we should update GetOperatorPodName to return correct name base on the clusterwide env variable.

test/testenv/util.go
```diff
@@ -886,3 +886,16 @@ func DeleteOperatorPod(testcaseEnvInst *TestCaseEnv) error {
 	}
 	return nil
 }
+
+// DeleteFilesOnOperatorPod Delete files on Operator Pod
+func DeleteFilesOnOperatorPod(ctx context.Context, deployment *Deployment, podName string, filenames []string) error {
+	for _, filepath := range filenames {
+		cmd := fmt.Sprintf("rm -f %s", filepath)
+		_, err := ExecuteCommandOnOperatorPod(ctx, deployment, podName, cmd)
```
pdhanoyasplunk, line 894: As we discussed offline we might see flakiness if we try to delete files which are not present.

tpereirasplunk: i have used the largest app we have for the very same reason. It takes a while for the app to be copied. I tried running the test case multiple times. Got the timing right each time

test/c3/appframework/appframework_test.go
tpereirasplunk: Should be easy enough. Let me check where all the change would be needed.

tpereirasplunk: I have made the changes and my test passed locally. Let see if it clears the pipeline since the function was used in other places too

test/c3/appframework/appframework_test.go
```diff
@@ -2892,4 +2892,101 @@ var _ = Describe("c3appfw test", func() {
 			testenv.AppFrameWorkVerifications(ctx, deployment, testcaseEnvInst, allAppSourceInfo, splunkPodAge, "")
 		})
 	})
+
+	Context("Single Site Indexer Cluster with Search Head Cluster (C3) and App Framework", func() {
+		It("integration, c3, appframeworkc3, appframework: can deploy a C3 SVA with App Framework enabled and delete apps from init-apps when download is compelte", func() {
```
tpereirasplunk, line 2897: typo: complete

test/m4/appframework/appframework_test.go
```diff
@@ -2457,4 +2457,99 @@ var _ = Describe("m4appfw test", func() {
...
+			   * Verify bundle push is successful
+			   * Verify apps are copied and installed on Monitoring Console and on Search Heads and Indexers pods
+			*/
+
+			//################## SETUP ##################
+			// Download all apps from S3
+			appList := append(testenv.BigSingleApp, testenv.ExtraApps...)
+			appFileList := testenv.GetAppFileList(appList)
+			err := testenv.DownloadFilesFromS3(testDataS3Bucket, s3AppDirV1, downloadDirV1, appFileList)
+			Expect(err).To(Succeed(), "Unable to download big-size app")
```
jambrosiano, line 2484: Nit: Need to edit the comment, as it's not only big-size apps but also extra apps that are downloaded here

test/s1/appframework/appframework_test.go
```diff
@@ -1780,4 +1780,75 @@ var _ = Describe("s1appfw test", func() {
...
+				* When app download is complete, delete apps from init-apps
+				############## VERIFICATIONS ################
+				* Verify App installation is in progress on Standalone
+				* Upload more apps from S3 during bigger app install
+				* Wait for polling interval to pass
+			    * Verify all apps are installed on Standalone
+			*/
+
+			// ################## SETUP FOR STANDALONE ####################
+			// Download all test apps from S3
```
jambrosiano, line 1801: Same comment than for C3, it's only 1 app downloaded here

jambrosiano (approved):

pdhanoyasplunk (approved):

PR #830, opened 2022-06-22
Call fails when there are any Warning message from PodExec Call

vivekr-splunk:
Fix for Call fails when there are any Warning message from PodExec Call and fixed an issue in kustomize file change that used to pick splunk 8.2.6 build (hardcoded value)

### DIFF
--- a/.github/workflows/int-test-workflow.yml
+++ b/.github/workflows/int-test-workflow.yml
@@ -4,7 +4,6 @@ on:
     branches:
       - develop
       - master
-      - cspl-1851
 jobs:
   build-operator-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/int-test-workflow.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - master
+      - cspl-1851
```
gaurav-splunk, line 7: just putting out a comment here so that we don't forget to remove this once all tests pass!

gaurav-splunk (approved):

PR #842, opened 2022-07-13
CSPL-1887 - Add telemetry app for Splunk Operator

akondur:
The PR implements injecting a telemetry app into primary Splunk Instances such as ClusterMaster, Standalone, SHC(Special Handling), ClusterMaster.

When a CR is in the "Ready" State the operator creates a telemetry app and handles the installation in different Splunk Enterprise instances types.

Detailed ERD about implementation can be found [here](https://splunk.atlassian.net/wiki/spaces/PROD/pages/1078154237773/ERD+for+Telemetry+for+Splunk+operator+for+Kubernetes#ERDforTelemetryforSplunkoperatorforKubernetes-1.ProblemStatement).

Telemetry app installation tested on all Splunk Enterprise Instances. Details in the ERD


api/v3/clustermaster_types.go
```diff
@@ -64,6 +64,9 @@ type ClusterMasterStatus struct {
 
 	// App Framework status
 	AppContext AppDeploymentContext `json:"appContext"`
+
```
vivekr-splunk, line 67: Should we use Annotations for Apps, Apps are not standard configurations that remain constant , they may be optional, or they differ based on customer setup

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -89,6 +89,88 @@ func getApplicablePodNameForAppFramework(cr splcommon.MetaObject, ordinalIdx int
 	return fmt.Sprintf("splunk-%s-%s-%d", cr.GetName(), podType, ordinalIdx)
 }
 
+// runCustomCommandOnSplunkPods creates the required directory for the pod/s
```
sgontla, line 92: nit: We can generalize, instead of mentioning it as directory creation, as that is what we are doing in this.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -89,6 +89,88 @@ func getApplicablePodNameForAppFramework(cr splcommon.MetaObject, ordinalIdx int
...
+		scopedLog.Info("arjunnew Creating the telemetry app on Splunk Deployer")
+		// create the app on Splunk pod/s
+		err = runCustomCommandOnSplunkPods(ctx, cr, replicas, command, podExecClient)
+		if err != nil {
+			scopedLog.Error(err, "unable to run command on splunk pod")
+			return err
+		}
+
+		command = fmt.Sprintf(applySHCBundleCmdStr, GetSplunkStatefulsetURL(cr.GetNamespace(), SplunkSearchHead, cr.GetName(), 0, false), "/tmp/status.txt")
+		scopedLog.Info("arjunnew Deployer bundle push")
```
sgontla, line 161: nit: arjunnew

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -89,6 +89,88 @@ func getApplicablePodNameForAppFramework(cr splcommon.MetaObject, ordinalIdx int
...
+	reqLogger := log.FromContext(ctx)
+	scopedLog := reqLogger.WithName("addTelApp").WithValues(
+		"name", cr.GetObjectMeta().GetName(),
+		"namespace", cr.GetObjectMeta().GetNamespace())
+
+	// Create pod exec client
+	podExecClient := splutil.GetPodExecClient(client, cr, "")
+
+	// Handle non SHC scenarios(Standalone, CM, LM)
+	if cr.GetObjectKind().GroupVersionKind().Kind != "SearchHeadCluster" {
```
sgontla, line 132: 1. Any reason, we can't use the bundle push mechanism for the IDXC also, that way, we don't need to call the reload end point for each indexer instance.
2. Don't we need the app on the deployer?
3. Also, considering a future possibility of having a separate CRD for the deployer, i think we can just check for the SHC explicitly.

pkg/splunk/enterprise/names.go
```diff
@@ -124,6 +124,37 @@ const (
...
 	livenessProbePeriodSec       = 30
+
+	// Number of ClusterMasterReplicas
+	numberOfClusterMasterReplicas = 1
+
+	// Number of Deployer replicas
+	numberOfDeployerReplicas = 1
+
+	// Number of Licensemaster replicas
+	numberOfLicenseMasterReplicas = 1
```
sgontla, line 135: Ideally, we should be able to use the CR replica count value? If the reason for this change is invalid value for the replica count, probably, it is high time for validating the same during the spec validation time, with a default and optional values.

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -211,6 +211,16 @@ func ApplySearchHeadCluster(ctx context.Context, client splcommon.ControllerClie
 		cr.Status.AdminPasswordChangedSecrets = make(map[string]bool)
 		cr.Status.NamespaceSecretResourceVersion = namespaceScopedSecret.ObjectMeta.ResourceVersion
 
+		// Add a splunk operator telemetry app
+		if !cr.Status.TelAppInstalled && !cr.Spec.Mock {
```
sgontla, line 215: I may be wrong, but i thought we removed all the references to mock?

sgontla, line 215: For some instances ephemeral storage is common, so, we should cover that scenario as well. Otherwise, we will lose the app across the resets. Indexers are the best examples. I think, if we use the bundle push logic for IDXC, then it is implicitly covered(assuming that the Cluster Master always uses the persistent storage)?

pkg/splunk/enterprise/names.go
```diff
@@ -124,6 +124,37 @@ const (
...
+
+	// Telemetry app.conf string
+	telAppConfString = `[install]
+is_configured = 1
+	
+[ui]
+is_visible = 0
+label = Splunk Operator for K8s
```
sgontla, line 143: Do you think, we can also inject the splunk personality, that way, we can find how many instances of a given kind?

api/v3/clustermaster_types.go
akondur: Hi @vivekr-splunk , in this case the Splunk Operator telemetry app is a static standard configuration. Could you provide an example of how annotations can be added to apps?

docs/Telemetry.md
```diff
@@ -0,0 +1,7 @@
+# Splunk Operator Telemetry
+
+A telemetry solution to collect data from the Splunk Enterprise instances deployed using the Splunk Operator for Kubernetes(SOK) is implemented. The operator installs an app called `splunk_operator_for_k8s_tel` into [primary Splunk Instances](https://docs.splunk.com/Documentation/Splunk/latest/Admin/Shareperformancedata#Which_instance_runs_the_searches_and_sends_data_to_Splunk). The newly installed app's data will be exported as a part of [Splunk Enterprise telemetry solution](https://docs.splunk.com/Documentation/Splunk/latest/Admin/Shareperformancedata) to the telemetry prod servers. The exported telemetry data will help understand number and kind of Splunk deployments using SOK.
```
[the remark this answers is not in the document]
akondur: Hi @sgontla , thanks for the suggestion. Is there any reason for the change in name?

pkg/splunk/enterprise/afwscheduler.go
akondur: Hi @sgontla , as discussed the app will exist only on primary instances i.e CM and not the indexers. The app is on the deployer as well as the SHC members currently. When we have a separate CRD for deployer we can check with kind as "Deployer"

pkg/splunk/enterprise/names.go
akondur: The LM, CM CRs don't have a replica field in the spec. The SHC CR doesn't have a replica count for deployer?

pkg/splunk/enterprise/names.go
akondur: The contents of the app.conf here or any apps in general are not accessible via telemetry. The role of the Splunk instance needs to be determined via other collected fields. One idea is to name the app according to the role?

pkg/splunk/enterprise/searchheadcluster.go
akondur: Understood Subba. I have to introduce logic here to check whether the folder exists if its ephemeral storage.

akondur: The pod exec commands via the operator are avoided for unit test using [MockPodExecClient](https://github.com/splunk/splunk-operator/blob/049d00f5eb64c37db4610922e5890021b0fa42d9/pkg/splunk/test/util.go#L61). But the problem here is that unit tests such as **TestClusterMasterWithReadyState**, **TestStandaloneWithReadyState** are directly calling ApplyClusterManager which either takes either the common controller client or [mock test client](https://github.com/splunk/splunk-operator/blob/049d00f5eb64c37db4610922e5890021b0fa42d9/pkg/splunk/test/controller.go#L224) as an argument but won't take MockPodExecClient as an argument. Should we implement mock test client as PodExecClientImpl?

docs/Telemetry.md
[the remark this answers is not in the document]
akondur: Adopted the new name :)

pkg/splunk/enterprise/names.go
akondur: Changed the app name to include the kind of deployment, thanks for the suggestion @sgontla

pkg/splunk/enterprise/searchheadcluster.go
sgontla: I think @vivekr-splunk is using a new framework, to avoid this?

akondur: That's right @sgontla , but that framework doesn't work here due to the call stack originating from **ApplyClusterManager** which cannot take **MockPodExecClient** as an argument. @vivekr-splunk Any thoughts here?

pkg/splunk/enterprise/names.go
akondur: @sgontla Are you suggesting introducing new `replica` fields in the CM, LM and SHC CR(for deployer)?

akondur: As discussed, we will take care of this at a later time as the controller has logic to initialize default values only when the spec field is at-least an empty struct while creating the CR.

**Details**:

I have two LM's(one with an empty spec struct and the other with no spec struct declared at all):

```
kind: LicenseMaster
metadata:
  name: test-with-spec
  finalizers:
spec: {}
---
kind: LicenseMaster
metadata:
  name: test-no-spec
  finalizers:
```

For the `test-with-spec` LM I see that the replica count is 1(which is set a default value with kubebuilder) while the `test-no-spec` has replicas 0.

pkg/splunk/enterprise/searchheadcluster.go
akondur: Removed the use of mock and added UT's.

sgontla (approved):

vivekr-splunk (approved):
LGTM

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -89,6 +89,108 @@ func getApplicablePodNameForAppFramework(cr splcommon.MetaObject, ordinalIdx int
...
+		return "shc", nil
+	case "ClusterMaster":
+		return "cm", nil
+	default:
+		return "", errors.New("Invalid CR kind for telemetry app")
+	}
+}
+
+// addTelApp adds a telemetry app
+var addTelApp = func(ctx context.Context, podExecClient splutil.PodExecClientImpl, replicas int32, cr splcommon.MetaObject) error {
```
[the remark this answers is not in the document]
akondur: We are using function pointers to opt out of executing the function for UT instead of using the 'mock' flag.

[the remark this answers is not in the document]
gaurav-splunk: Well that is what mockPodExecClient was for in the first place, so that you can still execute the function with mock client and still cover the lines in code coverage.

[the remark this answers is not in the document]
akondur: I have used mockPodExecClient for testing the function **addTelApp**. But when the call comes from UT's calling say **ApplyClusterManager**, **ApplyLicenseManager** wherein the status of the CR reaches a **Ready** state, we are using this method to avoid calling **addTelApp** as we cannot pass **mockPodExecClient** as a client there.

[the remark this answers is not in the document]
gaurav-splunk: Ok got it 👍

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -89,6 +89,108 @@ func getApplicablePodNameForAppFramework(cr splcommon.MetaObject, ordinalIdx int
...
+
+		command = fmt.Sprintf(applySHCBundleCmdStr, GetSplunkStatefulsetURL(cr.GetNamespace(), SplunkSearchHead, cr.GetName(), 0, false), "/tmp/status.txt")
+
+		// Bundle push from deployer
+		err = runCustomCommandOnSplunkPods(ctx, cr, replicas, command, podExecClient)
+		if err != nil {
+			scopedLog.Error(err, "unable to run command on splunk pod")
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
akondur: Good catch @gaurav-splunk , addressed it.

gaurav-splunk (approved):

PR #843, opened 2022-07-13
CSPL-1916 - Update Makefile to ensure correct order in artifact generation

mgaldino-splunk:
The current order of artifact generation can leave the remaining `config/default/kustomization.yaml` set to only watch namespaces after execution, which is not the default. This Task aims at fixing the order of generation in the Makefile to ensure the default value remains after execution.


sgontla (approved):

gaurav-splunk (approved):

vivekr-splunk (approved):
LGTM

PR #847, opened 2022-07-20
CSPL-1711: Add a function to disable apps and upload them to S3

jambrosiano:
This PR adds a function that will download apps from S3, untar them, modify their conf file to disable them, re-tar and upload the disabled versions to S3


test/testenv/s3utils.go
```diff
@@ -226,3 +229,74 @@ func UploadFilesToS3(testS3Bucket string, s3TestDir string, applist []string, do
...
+		wildcardpath := untarredCurrentAppFolder + "/*/./"
+		bytepath, _ := exec.Command("/bin/sh", "-c", "cd "+wildcardpath+"; pwd").Output()
+		untarredAppRootFolder := string(bytepath)
+		untarredAppRootFolder = untarredAppRootFolder[:len(untarredAppRootFolder)-1] //removing \n at the end of folder path
+
+		// - Edit /default/app.conf (add "state = disabled" in [install] stanza)
+		appConfFile := untarredAppRootFolder + "/default/app.conf"
+		input, err := ioutil.ReadFile(appConfFile)
+		if err != nil {
+			log.Fatalln(err)
```
tpereirasplunk, line 272: Ideally if there is a error in the Readfile command we should return the err as there is no point in moving to the next stage

tpereirasplunk:
Do we want to use this function in the test cases that involve disabled apps?

test/testenv/s3utils.go
```diff
@@ -226,3 +229,74 @@ func UploadFilesToS3(testS3Bucket string, s3TestDir string, applist []string, do
 	}
 	return uploadedFiles, nil
 }
+
+// DisableAppsToS3 download apps from S3, untar them, modify their conf file to disable them, re-tar and upload the disabled version to S3
+func DisableAppsToS3(testDataS3Bucket string, s3AppDir string, downloadDir string, appFileList []string, s3TestDir string) ([]string, error) {
+
+	// Download apps tar files
+	DownloadFilesFromS3(testDataS3Bucket, s3AppDir, downloadDir, appFileList)
```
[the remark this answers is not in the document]
jambrosiano: List of apps to be disabled is provided through appFileList parameter, so all the apps in that list would be disabled. That list has to be created correctly in the tests before calling DisableAppsToS3.

Also about the apps files, it would look like this:
<xxappfwxx> folder:
    - disabled apps folder (containing tgz of disabled apps, for ex: 'microsoft-azure-add-on-for-splunk.tgz'
    - original apps tgz files for ex: 'microsoft-azure-add-on-for-splunk.tgz'
    - untarred_apps folder:
         - folders with name of the tgz file minus extension, for ex: microsoft-azure-add-on-for-splunk
              - inside each of these folders, the untarred app root folder, for ex:'TA-MS-AAD'

When uploading to S3, it would be like this:
<xxappfwxx> folder, 
    - 'disabled_apps' folder inside, containing the disabled tgz files

jambrosiano:
> Do we want to use this function in the test cases that involve disabled apps?

Yes, that's the idea for this ticket

test/testenv/s3utils.go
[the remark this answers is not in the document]
tpereirasplunk: For the download would it not be better to just upload to the required s3 location that the appframework spec is pointing to.

Would avoid extra downloads

tpereirasplunk (approved):

test/testenv/s3utils.go
```diff
@@ -226,3 +229,74 @@ func UploadFilesToS3(testS3Bucket string, s3TestDir string, applist []string, do
...
+			return nil, err
+		}
+		lines := strings.Split(string(input), "\n")
+		for i, line := range lines {
+			if strings.Contains(line, "[install]") {
+				lines[i] = "[install]\nstate = disabled"
+			}
+			if strings.Contains(line, "state = enabled") {
+				lines = append(lines[:i], lines[i+1:]...)
+			}
```
gaurav-splunk, line 282: Sorry did not understand the above lines. If we are already adding `disabled` in lines 277-279 then what does lines 280-282 does?

test/testenv/s3utils.go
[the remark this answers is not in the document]
pdhanoyasplunk: Like @tpereirasplunk mentioned the apps should be present on the local directory from the test download ?

test/testenv/s3utils.go
```diff
@@ -226,3 +229,74 @@ func UploadFilesToS3(testS3Bucket string, s3TestDir string, applist []string, do
...
+	for _, key := range appFileList {
+		// Create a specific folder for each app in 'untarred_apps'
+		tarfile := downloadDir + "/" + key
+		lastInd := strings.LastIndex(key, ".")
+		untarredCurrentAppFolder := untarredAppsMainFolder + "/" + key[:lastInd]
+		cmd := exec.Command("mkdir", untarredCurrentAppFolder)
+		cmd.Run()
+
+		// Untar the app
+		cmd = exec.Command("tar", "-xf", tarfile, "-C", untarredCurrentAppFolder)
```
pdhanoyasplunk, line 258: In my opinion it would be better to use go library to untar like here https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07
Depending on exec package installed on OS might raise issues in future.

test/testenv/s3utils.go
jambrosiano: Some apps I used for testing were having "state = enabled" inside the conf file, so it needs to be removed to avoid having both enabled and disabled at the same time in the file.

gaurav-splunk (approved):

test/testenv/s3utils.go
gaurav-splunk: As discussed, please open another task to create util APIs for tar and untar

gaurav-splunk (approved):

test/testenv/s3utils.go
jambrosiano: Will use CSPL-1964 to address this

PR #849, opened 2022-07-22
CSPL-447: Test Indexer Cluster with Standalone SH 

jambrosiano:
Test deploys an Indexer Cluster and a Standalone with the CM ref, then verifies the Standalone is configured correctly


test/c3/appframework/appframework_test.go
```diff
@@ -3049,4 +3049,73 @@ var _ = Describe("c3appfw test", func() {
...
+			   * Deploy Standalone with clusterMasterRef
+			   ############# VERIFICATION #################
+			   * Verify clusterMasterRef is present in Standalone's server.conf file
+			*/
+			//################## SETUP ####################
+			// Deploy C3 CRD
+			indexerReplicas := 3
+			testcaseEnvInst.Log.Info("Deploy Single Site Indexer Cluster")
+			err := deployment.DeploySingleSiteCluster(ctx, deployment.GetName(), indexerReplicas, false, "")
+			Expect(err).To(Succeed(), "Unable to deploy Single Site Indexer Cluste")
```
gaurav-splunk, line 3068: nit:Cluster

test/c3/appframework/appframework_test.go
```diff
@@ -3049,4 +3049,73 @@ var _ = Describe("c3appfw test", func() {
...
+			config := "master_uri"
+			cmPodName := fmt.Sprintf(testenv.ClusterManagerPod, deployment.GetName())
+			lastInd = strings.LastIndex(cmPodName, "-")
+			cmService := cmPodName[:lastInd]
+			value := "master_uri = https://" + cmService + "-service:8089"
+
+			testenv.VerifyConfOnPod(deployment, ns, standalonePodName, confFilePath, config, value)
+
+			// Verify no pods reset by checking the pod age
+			testenv.VerifyNoPodReset(ctx, deployment, testcaseEnvInst, testcaseEnvInst.GetName(), splunkPodAge, nil)
```
gaurav-splunk, line 3118: TBH, I am not a fan of checking the server.conf file to verify if we connected to the CM. We should call a rest endpoint and confirm if we connected to CM. Maybe something like `services/cluster/config`?

test/c3/appframework/appframework_test.go
```diff
@@ -3049,4 +3049,73 @@ var _ = Describe("c3appfw test", func() {
...
+					Volumes: []corev1.Volume{},
+					ClusterMasterRef: corev1.ObjectReference{
+						Name: deployment.GetName(),
+					},
+				},
+			}
+
+			// Deploy Standalone with clusterMasterRef
+			testcaseEnvInst.Log.Info("Deploy Standalone with clusterMasterRef")
+			standalone, err := deployment.DeployStandaloneWithGivenSpec(ctx, deployment.GetName(), spec)
```
pdhanoyasplunk, line 3085: The point of this test was to add standalone as single SH to the Cluster master. I think without adding the extra yaml string Standalone wont be configured as SH on CM. Please modify the test accordingly.

pdhanoyasplunk (approved):

tpereirasplunk (approved):

gaurav-splunk (approved):

PR #851, opened 2022-07-29
Fix pipeline error

tpereirasplunk:
Used PVT apps so that the apps are different from the other tests instead of moving the test case to the end

Consecutive 3 pass runs: https://github.com/splunk/splunk-operator/actions/runs/2777199369 

The failure were seen as we were running the tests in clusterwide mode. Since a single operator serves multiple test scenarios and we were exhausting the memory in the PV scenario, The apps from a different test case were being linked

### DIFF
--- a/.github/workflows/automated-release-workflow.yml
+++ b/.github/workflows/automated-release-workflow.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Generate Release Artifacts
       run: |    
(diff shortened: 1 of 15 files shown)


pdhanoyasplunk:
Does this guarantee the execution of test in the end? Also can you please add explanation on how does disabling apps cause the memory to exhaust?

tpereirasplunk:
> Does this guarantee the execution of test in the end? Also can you please add explanation on how does disabling apps cause the memory to exhaust? Also this would require us to ensure that this test is always executed in the end, are there any alternatives we can use ?


Not sure of any other way. But we can move one of the test case to smoke. It works in that way. Apart from that i wasnt able to repro the issue much locally or in smoke test. So had limited options

gaurav-splunk (approved):

pdhanoyasplunk (approved):

PR #865, opened 2022-08-11
cspl-1896 : Helm documentation for splunk operator 

vivekr-splunk:
https://splunk.atlassian.net/browse/CSPL-1986


docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+2. Using the Helm CLI to set new values
+```
+helm install --set <KEY>=<VALUE> <RELEASE_NAME> splunk/splunk-operator -n <RELEASE_NAMESPACE>
+```
+
+Read more about configuring values [here](https://helm.sh/docs/intro/using_helm/).
+
+## Splunk Enterprise deployments
+
+The Splunk Enterprise chart allows you to install and configure Splunk custom resources.
```
gaurav-splunk, line 62: nit: ... and configure Splunk Enterprise custom resources.

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+
+First build the Splunk Operator chart as a dependency:
+```
+helm dependency build splunk/splunk-enterprise
+```
+
+To install a configured Enterprise deployment:
```
gaurav-splunk, line 69: nit: Splunk Enterprise deployment

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+```
+## Splunk Validated Architecture deployments
+
+The Splunk Enterprise chart has support for three Splunk Validated Architectures:
+
+- [Single Server Deployment (S1)](https://www.splunk.com/pdfs/technical-briefs/splunk-validated-architectures.pdf#page=9)
+- [Distributed Clustered Deployment + SHC - Single Site (C3)](https://www.splunk.com/pdfs/technical-briefs/splunk-validated-architectures.pdf#page=14)
+- [Distributed Clustered Deployment + SHC - Multi-Site (M4)](https://www.splunk.com/pdfs/technical-briefs/splunk-validated-architectures.pdf#page=20)
+
+Install a Single Server Deployment release using the following command:
```
gaurav-splunk, line 93: nit: Install a Standalone deployment....

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+```
+helm upgrade --set splunkOperator.clusterWideAccess=false <RELEASE_NAME> splunk/splunk-operator -n <RELEASE_NAMESPACE>
+```
+
+Here we've upgraded the operator by revoking cluster-wide access.
+
+To rollback the operator to a previous release:
+
+```
+helm rollback <RELEASE_NAME> <REVISION_NUMBER> -n <RELEASE_NAMESPACE>
```
sgontla, line 35: Though this is a very good option at the Helm level, i think it can cause issues, when it comes to the Splunk deployment. I am not sure if this is even supported, particularly with the major releases, as we may not have reverse migration logic in place . Also,  i guess the order of the migration is also important, even if it is some minor releases.

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+
+Read more about configuring values [here](https://helm.sh/docs/intro/using_helm/).
+
+## Splunk Enterprise deployments
+
+The Splunk Enterprise chart allows you to install and configure Splunk Enterprise custom resources.
+
+First build the Splunk Operator chart as a dependency:
+```
+helm dependency build splunk/splunk-enterprise
```
akondur, line 66: Do we need this per this statement - `This chart installs the Splunk Operator as a dependency by default.` below?

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+To install a configured Splunk Enterprise deployment:
+```
+helm install <RELEASE_NAME> splunk/splunk-enterprise -n <RELEASE_NAMESPACE>
```
sgontla, line 71: nit: Any significance for the word RELEASE in RELEASE_NAMESPACE?

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+This chart installs the Splunk Operator as a dependency by default.
+
+If the operator is already installed then you will need to disable the dependency:
+```
+helm install --set splunk-operator.enabled=false <RELEASE_NAME> splunk/splunk-enterprise -n <RELEASE_NAMESPACE>
+```
+
+To see all configurable values contained in the ```values.yaml``` file:
```
akondur, line 81: We are not using the `-f` option for the file with values here for the `splunk/splunk-enterprise` chart, so would this be necessary? Perhaps this command is more useful in the `Splunk Operator Deployments` section.

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+helm uninstall <RELEASE_NAME> -n <RELEASE_NAMESPACE>
+```
+
+### Configuring Splunk Operator deployments
+
+There are a couple ways you can configure your operator deployment
+
+1. Using a ```new_values.yaml``` file to override default values (Recommended)
```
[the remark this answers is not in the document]
sgontla: Also, I think, it will be nice to have, an example deployment with the helm commands with some non-defaults, which will be very good reference for someone new to the helm.

docs/Helm.md
pabboud-splunk: This command is a quick way to disable ```splunk/splunk-operator``` installation when deploying the ```splunk/splunk-enterprise``` chart. Ideally, the user is always using ```-f``` option for either chart to keep track of the override values they have specified at each deployment.

docs/Helm.md
pabboud-splunk: Dependency build packages the ```splunk/splunk-operator``` chart within a sub-directory of the ```splunk/splunk-enterprise``` chart. It does not install the Splunk Operator. If the ```splunk/splunk-operator``` dependency is enabled and the dependencies have not been built, then an error will occur.

docs/Helm.md
pabboud-splunk: When using the any of the commands ```install```, ```upgrade```, ```rollback```, ```uninstall``` for a specific release, the release's namespace must be specified. Including it here is just to be explicit to which namespace we're referring to.

docs/Helm.md
akondur: In that case can we edit the statement above to reflect the dependency i.e `The Splunk Enterprise chart allows you to install and configure Splunk Enterprise custom resources. The splunk/splunk-enterprise chart has the splunk/splunk-operator chart as a dependency by default. To satisfy the dependencies please use the following command: ` ? 

With this the statement `This chart installs the Splunk Operator as a dependency by default.` would no longer be needed

docs/Helm.md
pabboud-splunk: I agree, I've removed the section for now.

docs/Helm.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+helm upgrade -f new_values.yaml <RELEASE_NAME> splunk/splunk-operator -n <RELEASE_NAMESPACE>
+```
+
+Read more about configuring values [here](https://helm.sh/docs/intro/using_helm/).
+
+In the following example, we will install and upgrade the Splunk Operator.
+
+Specify the release name and namespace to install the Operator:
```
akondur, line 48: Nit: If we could specify that we can get the list of releases via the command that would be great:

`helm list`

docs/Helm.md
```diff
@@ -0,0 +1,168 @@
...
+
+indexerCluster:
+  enabled: true
+  name: idxc-test
+
+searchHeadCluster:
+  enabled: true
+  name: shc-test
+```
+The configurations above will override values in ```splunk/splunk-enterprise``` values file.  To see all configurable values contained in the ```values.yaml``` file:
```
akondur, line 110: Nit: If we could somehow mention that by default the `values.yaml `without any override deploys a `standalone` that would be good. The current statement would be fine too if not possible.

akondur (approved):
Great work @pabboud-splunk

docs/Helm.md
```diff
@@ -0,0 +1,106 @@
...
+Install a Standalone deployment using the following command:
+```
+helm install --set s1.enabled=true <RELEASE_NAME> splunk/splunk-enterprise -n <RELEASE_NAMESPACE>
+```
+Visit the Splunk Operator github repository to learn more about the configurable values of [splunk/splunk-operator](https://github.com/splunk/splunk-operator/blob/develop/helm-chart/splunk-operator/values.yaml) and [splunk/splunk-enterprise](https://github.com/splunk/splunk-operator/blob/develop/helm-chart/splunk-enterprise/values.yaml).
+
+
+
+
+
```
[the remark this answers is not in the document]
pabboud-splunk: Yes, that is in the pipeline. In the future, we are hoping to migrate existing deployments into helm releases. This may be possible by annotating all existing resources with the following fields:
`meta.helm.sh/release-name=<RELEASE_NAME>`
`meta.helm.sh/release-namespace=<RELEASE_NAMESPACE>`
`app.kubernetes.io/managed-by=Helm`
However, we need to explore this further. At the moment, the only solution is to create an entirely new release.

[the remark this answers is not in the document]
sgontla: Got it. If not in this PR, let us give that clarity in the documentation, so that we can avoid confusion.

sgontla (commented):
LGTM

PR #869, opened 2022-08-18
CSPL-1874: Helm chart support for all Splunk Instance types

pabboud-splunk:
Helm chart implementation of Splunk Instance as a separate chart from Operator.
Splunk Instance Chart to deploy a supported instance of Splunk managed by an Operator

- Standalone
- Cluster Manager
- Indexer Cluster
- Search Head Cluster
- Monitoring Console
- License Manager

### DIFF
--- a/helm-chart/splunk-enterprise/Chart.yaml
+++ b/helm-chart/splunk-enterprise/Chart.yaml
@@ -1,0 +1,30 @@
+apiVersion: v2
+name: splunk-enterprise
+description: A Helm chart for Splunk Enterprise managed by the Splunk Operator
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 1.0.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "2.0.0"
+
+dependencies:
+- name: splunk-operator
+  version: "1.0.0"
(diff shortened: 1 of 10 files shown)


helm-chart/splunk-enterprise/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "2.0.0"
+
+dependencies:
+- name: splunk-operator
+  version: "1.0.0"
```
vivekr-splunk, line 28: should we change this to 2.0.0 ? , am ok with 1.0.0 also

helm-chart/splunk-enterprise/values.yaml
```diff
@@ -0,0 +1,529 @@
...
+  # - name: 
+  #   value: 
+
+  # Delay in seconds before probing liveness
+  # refernece: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command
+  livenessInitialDelaySeconds: 300
+
+  # Delay in seconds before probing readiness
+  # https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command
+  readinessInitialDelaySeconds: 10
```
sgontla, line 130: These are already defaults. We can remove this? Also, soon these are going to be phased out.
Similarly for other places.

helm-chart/splunk-enterprise/Chart.yaml
vivekr-splunk: I am ok to have separate chart version than splunk operator.

helm-chart/splunk-enterprise/values.yaml
vivekr-splunk: in values.yaml file we need to provide all the variables that are in spec. even if they have default values.

sgontla (approved):

vivekr-splunk (approved):
LGTM

helm-chart/splunk-enterprise/Chart.yaml
pabboud-splunk: Yes, so it can be anything we want. The version specified above is the chart version for the splunk-operator chart. Currrently on the develop branch it is 1.0.0 for the chart version and 2.0.0 for the appVersion.

kashok-splunk (approved):

gaurav-splunk (approved):

PR #872, opened 2022-08-19
CSPL-1996 - Upgrade CRD API version for *Manager CRDs (v4)

mgaldino-splunk:
This PR enables V3 and V4 APIs to share Specs and functions.

The newly created CRDs will be versioned as V4, while the older counterparts `LicenseMaster`, `ClusterMaster` will be versioned as V3. 

**Changes:**
Upgraded `enterpriseApi` to reference V4 APIs. 
Upgraded `LicenseMaster`, `ClusterMaster`  to reference `enterpriseApiV3` across the codebase.


Makefile
```diff
@@ -139,10 +139,10 @@ docker-push: ## Push docker image with the manager.
 ##@ Deployment
 
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	$(KUSTOMIZE) build config/crd | kubectl replace -f -
 
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=true -f -
```
vivekr-splunk, line 145: this is temporary code change  , i will add PR to fix this one later

Makefile
```diff
@@ -139,10 +139,10 @@ docker-push: ## Push docker image with the manager.
 ##@ Deployment
 
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	$(KUSTOMIZE) build config/crd | kubectl replace -f -
```
vivekr-splunk, line 142: we  need to find a solution for this . this is kubectl issue. when we run apply, it copies the whole CR and adds that to annotations as history, but since our CR is too big , its more than 1MB the limitation from etcd database for any row data. so kubectl fails. there is already an issue raised by many on this. replace is not a good idea as this will delete the CRD and replaces that with latest. but `make install` command is used only by developer it should be ok. we need to find a way to fix this for user

PROJECT
```diff
@@ -79,4 +79,36 @@ resources:
...
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: splunk.com
+  group: enterprise
+  kind: Standalone
+  path: github.com/splunk/splunk-operator/api/v4
+  version: v4
+- api:
+    crdVersion: v1
```
vivekr-splunk, line 99: i couldnt find ClusterManager and LicenseManager v4 . is this missing?

Makefile
mgaldino-splunk: Rolled back in the latest commit.

Makefile
mgaldino-splunk: Rolled back in the latest commit.

PROJECT
mgaldino-splunk: Indeed it was missing. I rerun `operator-sdk create api ` for LM & CM to update the PROJECT, then restored the files `api/v4`.

mgaldino-splunk:
@vivekr-splunk , @sgontla , @gaurav-splunk: 

Do you think after these changes the `feature-biaslang` PR needs to through QA again? If so we need to create the stories/allocate resources.

.github/workflows/int-test-workflow.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - master
+      - fixed-version-biaslang
```
gaurav-splunk, line 7: Just keeping a note here so that we don't forget to remove it before merge.

Makefile
```diff
@@ -142,7 +142,7 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found)  -f -
```
gaurav-splunk, line 145: nit: just a difference of a space?

api/v3/indexercluster_types.go
```diff
@@ -123,7 +124,6 @@ type IndexerClusterStatus struct {
 // +kubebuilder:printcolumn:name="Desired",type="integer",JSONPath=".status.replicas",description="Desired number of indexer peers"
 // +kubebuilder:printcolumn:name="Ready",type="integer",JSONPath=".status.readyReplicas",description="Current number of ready indexer peers"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age of indexer cluster"
-// +kubebuilder:storageversion
```
gaurav-splunk, line 126: why this change?

PROJECT
```diff
@@ -22,7 +22,7 @@ resources:
   controller: true
   domain: splunk.com
   group: enterprise
-  kind: ClusterMaster
+  kind: IndexerCluster
```
sgontla, line 25: Why are these 'Kind's changing? same with other places as well.

api/v3/clustermaster_types.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v3
 
 import (
+	enterpriseApi "github.com/splunk/splunk-operator/api/v4"
```
sgontla, line 20: In the past, we use to move the spec to the latest API version. Are we going to retain it going forward? Any specific reason for this, as we are not using it to build the CRDs any more?


Same comment for all the v3 specific files.

pkg/splunk/enterprise/afwscheduler_test.go
```diff
@@ -25,7 +26,7 @@ import (
 	"testing"
 	"time"
 
-	enterpriseApi "github.com/splunk/splunk-operator/api/v3"
+	enterpriseApiV3 "github.com/splunk/splunk-operator/api/v3"
```
sgontla, line 29: Do we need a backlog story item, to cover the clusterManager as well? Starting this release, if my understanding is correct, we are enforcing the customers to use ClusterManager(?), if so, we need to cover all the test scenarios as soon as possible?

@gaurav-splunk , @vivekr-splunk

controllers/common/predicate.go
```diff
@@ -195,7 +196,7 @@ func ClusterManagerChangedPredicate() predicate.Predicate {
 	err := predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			// This update is in fact a Delete event, process it
-			if _, ok := e.ObjectNew.(*enterpriseApi.ClusterMaster); !ok {
+			if _, ok := e.ObjectNew.(*enterpriseApiV3.ClusterMaster); !ok {
```
gaurav-splunk, line 199: @vivekr-splunk @mgaldino-splunk shouldn't this function name be `ClusterMasterChangedPredicate` since it is referring to ClusterMaster?

controllers/suite_test.go
```diff
@@ -90,12 +92,18 @@ var _ = BeforeSuite(func() {
...
 
 	err = enterpriseApi.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	//err = enterpriseApiV3.AddToScheme(scheme.Scheme)
+	//Expect(err).NotTo(HaveOccurred())
+	//
+	//err = enterpriseApiV3.AddToScheme(scheme.Scheme)
+	//Expect(err).NotTo(HaveOccurred())
+
```
gaurav-splunk, line 106: probably needs to be removed?

PROJECT
mgaldino-splunk: The git view became a bit confusing here. If we compare the whole file it's easier to understand the changes. In short:
```
*Master -> Only supported in V3
*Manager -> supported in V4
All others(Indexer,SH,etc) -> supported in both V3 and V4
```
This was caused because we had to run `operator-sdk create api ` again so the order of items changed, but git shows the changes a bit convoluted.

vivekr-splunk:
> @vivekr-splunk , @sgontla , @gaurav-splunk:
> 
> Do you think after these changes the `feature-biaslang` PR needs to through QA again? If so we need to create the stories/allocate resources.

yes this needs to go through one round of QA testing

api/v3/clustermaster_types.go
mgaldino-splunk: This time it's a bit different because we now support both V3 and V4. But we have a couple of shared specs in `common_types.go` that can't be duplicated. 

We linked `enterpriseApi "github.com/splunk/splunk-operator/api/v4"` so we can continue to use it in V3 for example `enterpriseApi.CommonSplunkSpec`

pkg/splunk/enterprise/afwscheduler_test.go
mgaldino-splunk: The tests for *Manager were added in separate files that are similar to their *Master counterparts. For example, we created `test/m4/appframework/manager_appframework_test.go` for Manager, while keeping the `original test/m4/appframework/appframework_test.go` for Master.

controllers/common/predicate.go
vivekr-splunk: I think thats ok, we will reuse the same function for clustermanager also.

api/v3/indexercluster_types.go
vivekr-splunk: we are not going to store v3 instead we will store v4. we only server v3

Makefile
vivekr-splunk: this needs change, that I will be doing it in separate PR later. for now marcus removed the change , i think he left a space.

pkg/splunk/enterprise/afwscheduler_test.go
vivekr-splunk: any new creation in V4 we are asking user to use V4, if they are doing it in V3 we still support. yes in V4 having the 2 API to serve same resource do not make sense.

api/v3/clustermaster_types.go
vivekr-splunk: we are going to move all the resources to V4. except Licensemaster and ClustreMaster which we are going to deprecate in coming releases (not now in near feature)

api/v3/indexercluster_types.go
mgaldino-splunk: Each CRD can be stored once. So for the CRDs that are in both V4 and V3, such as IndexerCluster, we only add the storage option in the latest version(V4).

controllers/suite_test.go
vivekr-splunk: old code , yes need some changes we can take this in separate CR

vivekr-splunk (approved):
LGTM

controllers/suite_test.go
mgaldino-splunk: done

controllers/common/predicate.go
mgaldino-splunk: Because this is attached to IndexerCluster, and we have it in both V3 and V4 I duplicated it in the latest commit. Then it shouldn't be any conflict.

sgontla (approved):

Makefile
mgaldino-splunk: done in the latest commit.

gaurav-splunk (approved):

PR #874, opened 2022-08-20
CSPL-1945: Edit script to create AKS cluster with/without option for managed identity also

jambrosiano:
This PR edits the deploy-azure-cluster.sh script previously introduced by CSPL-1821 by adding the possibility to have a cluster with and without an assigned managed identity.

A new boolean parameter `AZURE_MANAGED_ID_ENABLED` is introduced to switch between using managed identity or not.
If `AZURE_MANAGED_ID_ENABLED` is true, the managed identity  `<CLUSTER-NAME>-agentpool`, which has been created by aks at the same time as when the cluster was created, will be used. Therefore we need to assign read access for the Kubelet user managed identity ( `<CLUSTER-NAME>-agentpool` ) to the storage account, i.e adding the 'Storage Blob Data Reader' role to it. **Note** that this identity is scoped to all the nodes and pods in the same AKS cluster so any pod running in the same AKS cluster will have read permissions to the buckets and app packages.

If AZURE_MANAGED_ID_ENABLED is false, no further action is required after the AKS cluster creation (secret usage is out of this PR scope)


gaurav-splunk (approved):

tpereirasplunk (approved):

kumarajeet (approved):
All looks good, just revised the description little bit :)

PR #884, opened 2022-09-03
CSPL-1823: Add utils to support the use of Azure Blob Storage for App Framework and Smartstore tests

jambrosiano:


kumarajeet (commented):
We may want to centralize  some common code that is used in this util as well @akondur  and I had in the https://github.com/splunk/splunk-operator/pull/884/files.

For example: ComputeHMACSHA256 , buildStringToSign.... methods are in this util (test/testenv/azureutils.go) as well as in the above MR ( under pkg/splunk/test/azureblobclient.go) 

I would say to use this util code as it is for current sprint  and in one of the next sprints, I can work with @jambrosiano  to refactor the code where the common functions/methods can move to central location (e.g. pkg/splunl/util.go)

testenv/azureutils.go                             --> (common methods) --> pkg/splunk/util/util.go

pkg/splunk/client/azureblobclient.go  --> (common methods) -->  pkg/splunk/util/util.go

gaurav-splunk:
Please close this PR

jambrosiano:
Will use https://github.com/splunk/splunk-operator/pull/889 instead to merge into azure feature branch

PR #892, opened 2022-09-09
CSPL-1923: Addition of code and unittests to download the app packages

kumarajeet:
**Adding functionality to download app packages from Azure blob**
Changes related to CSPL-1923:
Extending  `azureblobclient.go` to allow downloading app packages from azure blob containers (buckets) 
Extending the unit tests to tests the download functionality as well as enhance the app lists unit tests

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -16,5 +16,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: splunk/splunk-operator
-  newTag: 1.1.0
+  newName: controller
+  newTag: latest
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -16,5 +16,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: splunk/splunk-operator
+  newTag: 1.1.0
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -17,4 +17,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: splunk/splunk-operator
-  newTag: 1.1.0
+  newTag: 2.0.0
--- a/pkg/splunk/client/azureblobclient_test.go
+++ b/pkg/splunk/client/azureblobclient_test.go
@@ -21,6 +21,8 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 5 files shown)


jryb (approved):

vivekr-splunk (approved):

pkg/splunk/client/azureblobclient_test.go
```diff
@@ -167,230 +298,210 @@ func TestAzureBlobGetAppsList(t *testing.T) {
...
+	wantRequest, _ := http.NewRequest("GET", "https://mystorageaccount.blob.core.windows.net/appscontainer1/adminAppsRepo/app1.tgz", nil)
+	respdata := "This is a test body of an app1.tgz package. In real use it would be a binary file but for test it is just a string data"
 
-	var vol enterpriseApi.VolumeSpec
-	var err error
+	mclient.AddHandler(wantRequest, 200, respdata, nil)
 
+	// Get App source and volume from spec
 	appSource := appFrameworkRef.AppSources[0]
+	vol, _ := GetAppSrcVolume(ctx, appSource, &appFrameworkRef)
```
gaurav-splunk, line 452: why are we not checking the error here?

config/manager/kustomization.yaml
```diff
@@ -16,5 +16,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: splunk/splunk-operator
-  newTag: 1.1.0
\ No newline at end of file
+  newName: controller
```
gaurav-splunk, line 19: are you sure this is correct? CC: @vivekr-splunk ?

kumarajeet: I think I pushed local change, so I reverted that to what we had originally.

config/manager/kustomization.yaml
```diff
@@ -16,5 +16,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: splunk/splunk-operator
+  newTag: 1.1.0
```
gaurav-splunk, line 20: Also, is this tag also correct? it should be something 2.x, isn't it? @vivekr-splunk ?

vivekr-splunk: yes it should be 2.0.0 but creation of image is done dynamically, just for the purpose of having right number we can change this to 2.0.0 . pipeline always uses right tag name when it creates

config/manager/kustomization.yaml
vivekr-splunk: no changes to config directory is required. unless we are doing changes to CRD.

config/manager/kustomization.yaml
kumarajeet: Thanks Vivek and Gaurav. Updated to 2.0.0

gaurav-splunk (approved):

PR #900, opened 2022-09-15
CSPL-1981: Add a lightweight liveness probe

sgontla:
Cases like the Splunk instance decommissioning, or when the Splunk is coming up with large number of indexs, existing liveness may fail, as the mgmt port may be busy.

Changes are made such that the liveness level may be reduced during the maintenance activity, so that the Pod will not be reset due to the liveness probe failure

### DIFF
--- a/tools/k8_probes/livenessProbe.sh
+++ b/tools/k8_probes/livenessProbe.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+# Copyright 2022 Splunk
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


tools/k8_probes/livenessProbe.sh
```diff
@@ -0,0 +1,88 @@
...
+    #If NO_HEALTHCHECK is NOT defined, then we want the healthcheck
+    state="$(< $CONTAINER_ARTIFACT_DIR/splunk-container.state)"
+
+    case "$state" in
+    running|started)
+        curl --max-time 30 --fail --insecure $HTTP_SCHEME://localhost:8089/
+        exit $?
+    ;;
+    *)
+        exit 1
```
vivekr-splunk, line 71: do you think adding logging message will help it to show that error in event log

vivekr-splunk (approved):

gaurav-splunk (approved):

PR #901, opened 2022-09-15
CSPL-2018-cm-implementation-for-probes

pdhanoyasplunk:
Add Config Map to mount probe scripts on pod. 
Modified the liveness and readiness probes to use the mounted scripts


pkg/splunk/enterprise/configuration.go
```diff
@@ -522,9 +528,57 @@ func addStorageVolumes(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunk
 		}
 	}
 
+	// Add Splunk Probe config map
+	probeConfigMap := getProbeConfigMap(ctx, client, cr)
+	if probeConfigMap == nil {
+		scopedLog.Error(nil, "Unable to get probeConfigMap")
```
sgontla, line 534: return on error?

pkg/splunk/enterprise/configuration.go
```diff
@@ -522,9 +528,57 @@ func addStorageVolumes(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunk
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetProbeConfigMapName(cr.GetNamespace()),
+			Namespace: cr.GetNamespace(),
+		},
+	}
+
+	// Add readiness script to config map
+	data, err := ReadFile(ctx, GetReadinessScriptLocation())
+	if err != nil {
+		return nil
```
sgontla, line 552: Can you also propagate the error ?

pkg/splunk/enterprise/configuration.go
```diff
@@ -522,9 +528,57 @@ func addStorageVolumes(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunk
...
+	}
+	configMapData := map[string]string{GetReadinessScriptName(): data}
+
+	// Add liveness script to config mpa
+	livenessScriptLocation, _ := filepath.Abs(GetLivenessScriptLocation())
+	data, err = ReadFile(ctx, livenessScriptLocation)
+	if err != nil {
+		return nil
```
sgontla, line 560: Same as above

pkg/splunk/enterprise/configuration.go
```diff
@@ -522,9 +528,57 @@ func addStorageVolumes(cr splcommon.MetaObject, spec *enterpriseApi.CommonSplunk
...
+			Namespace: cr.GetNamespace(),
+		},
+	}
+
+	// Add readiness script to config map
+	data, err := ReadFile(ctx, GetReadinessScriptLocation())
+	if err != nil {
+		return nil
+	}
+	configMapData := map[string]string{GetReadinessScriptName(): data}
```
sgontla, line 554: nit: you may want directly allocate the `configMap.data` map.

pkg/splunk/enterprise/names.go
```diff
@@ -42,6 +42,27 @@ const (
...
+	readinessScriptName = "readinessProbe.sh"
+
+	// readinessScriptLocation
+	readinessScriptLocation = "tools/k8_probes/" + readinessScriptName
+
+	// livenessScriptLocation
+	livenessScriptLocation = "tools/k8_probes/" + livenessScriptName
+
+	// probeMountDirectory
+	probeMountDirectory = "/mnt/probescripts"
```
sgontla, line 61: nit: can we name it as /mnt/probes/ ?

pkg/splunk/enterprise/configuration.go
vivekr-splunk, line 534: should we return here with error or should we create new configmap

pkg/splunk/enterprise/configuration.go
vivekr-splunk: +1

pkg/splunk/enterprise/names.go
```diff
@@ -250,3 +271,38 @@ func GetSplunkImage(specImage string) string {
 func GetPortName(port string, protocol string) string {
 	return fmt.Sprintf(portNameTemplateStr, protocol, port)
 }
+
+// GetProbeConfigMapName uses a template to name a Kubernetes ConfigMap for a Liveness/Readiness Probe.
+func GetProbeConfigMapName(identifier string) string {
```
akondur, line 276: Can't we have this as a constant like other [places](https://github.com/splunk/splunk-operator/blob/7098178a23c4a51f81f20027ecf0be7b0821b0e1/pkg/splunk/enterprise/names.go#L31) where we can fill it in with a fmt.Sprintf?

pkg/splunk/enterprise/names.go
```diff
@@ -250,3 +271,38 @@ func GetSplunkImage(specImage string) string {
...
+	return fmt.Sprintf(probeConfigMapTemplateStr, identifier)
+}
+
+// GetReadinessScriptLocation return the location of readiness script
+var GetReadinessScriptLocation = func() string {
```
akondur, line 281: Is there any reason why we are using function pointers instead of constants here?

tools/k8_probes/livenessProbe.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# Copyright 2018 Splunk
```
gaurav-splunk, line 3: nit: this should be 2022 Splunk?

pkg/splunk/enterprise/names.go
pdhanoyasplunk: This is a getter for constant `probeConfigMapTemplateStr`

pkg/splunk/enterprise/names.go
pdhanoyasplunk: In unit test cases we needed to ScriptLocation. Defining as var helps us with that.

pkg/splunk/enterprise/clustermaster_test.go
```diff
@@ -44,13 +44,24 @@ import (
 	splutil "github.com/splunk/splunk-operator/pkg/splunk/util"
 )
 
+func init() {
+	// Re-Assigning GetReadinessScriptLocation and GetLivenessScriptLocation to use absolute path for readinessScriptLocation, readinessScriptLocation
+	GetReadinessScriptLocation = func() string {
+		fileLocation, _ := filepath.Abs("../" + "../" + "../" + readinessScriptLocation)
```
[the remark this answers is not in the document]
pdhanoyasplunk: Forgot to clean this up.

pkg/splunk/enterprise/configuration_test.go
```diff
@@ -49,6 +49,8 @@ func configTester(t *testing.T, method string, f func() (interface{}, error), wa
 
 func marshalAndCompare(t *testing.T, compare interface{}, method string, want string) {
 	got, err := json.Marshal(compare)
+	strGot := string(got)
+	fmt.Println(strGot)
```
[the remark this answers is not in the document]
pdhanoyasplunk: Left in error. Was using for debug.

pkg/splunk/enterprise/names.go
akondur: Can't we use a constant in names.go and use it directly? Since there is no `fmt.Sprintf` here we could use `readinessScriptLocation` directly?

pdhanoyasplunk: Then we would have no way of allocating new `readinessScriptLocation` in the unit test which will cause failures due to different relative path. 
This is a simpler approach to the problem. The other solution would be to copy the scripts to relative location before test start and cleaning up after the test finishes. 
Now we can just modify the location like [here](https://github.com/splunk/splunk-operator/blob/CSPL-2018/pkg/splunk/enterprise/clustermaster_test.go#L47)

gaurav-splunk (approved):

sgontla (approved):

PR #916, opened 2022-09-29
CSPL-1979 Add support for startup probe

pdhanoya-splunk:
* Writing Unit Test will update soon. 

Added support for startup probe.


tools/k8_probes/startupProbe.sh
```diff
@@ -0,0 +1,93 @@
...
+    ;;
+    *)
+        liveness_probe_default
+    ;;
+    esac
+
+else
+	#If NO_HEALTHCHECK is defined, ignore the healthcheck
+	exit 0
+fi
```
sgontla, line 93: I think, we will just use the old liveness probe script in its original form

pkg/splunk/enterprise/names.go
```diff
@@ -152,6 +158,11 @@ const (
 	livenessProbeTimeoutSec      = 30
 	livenessProbePeriodSec       = 30
 
+	// Startup probe time values
+	startupProbeDefaultDelaySec = 10
+	startupProbeTimeoutSec      = 5
+	startupProbePeriodSec       = 5
```
sgontla, line 164: These values are good. I think to avoid getting into any cyclic restarts caused by startup probes, particularly  on a pod reset for any lag on the Splunk side(Say, an already existing Splunk instance having large number of buckets, joining the cluster, destructive re-sync, etc...), i think it is safe to also add a large `FailureThreshold` value, just in the case of Startup Probe. Generally, all our pod resets take around 80 seconds(Cluster Manager is exception, but i guess should be OK to respond on mgmt port), so, considering the Period of 5 seconds, we can introduce the default `FailureThreshold` to 16 for the Startup Probe alone. We leave the default 3 for other probes. Thoughts?

CC: @vivekr-splunk

pkg/splunk/enterprise/configuration.go
```diff
@@ -998,6 +1025,15 @@ func getReadinessProbe(ctx context.Context, cr splcommon.MetaObject, instanceTyp
 	return readinessProbe
 }
 
+// getStartupProbe the probe for checking the first start of splunk on the Pod
+func getStartupProbe(ctx context.Context, cr splcommon.MetaObject, instanceType InstanceType, spec *enterpriseApi.CommonSplunkSpec) *corev1.Probe {
+	reqLogger := log.FromContext(ctx)
+	scopedLog := reqLogger.WithName("getStartupProbe").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+	startupProbe := getProbeWithConfigUpdates(&defaultStartupProbe, spec.StartupProbe, 0)
+	scopedLog.Info("ReadinessProbe", "Configured", startupProbe)
```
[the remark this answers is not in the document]
gaurav-splunk: +1

pkg/splunk/enterprise/configuration.go
```diff
@@ -67,6 +67,20 @@ var defaultReadinessProbe corev1.Probe = corev1.Probe{
...
 
+var defaultStartupProbe corev1.Probe = corev1.Probe{
+	InitialDelaySeconds: startupProbeDefaultDelaySec,
+	TimeoutSeconds:      startupProbeTimeoutSec,
+	PeriodSeconds:       startupProbePeriodSec,
+	FailureThreshold:    startupProbeFailureThreshold,
+	ProbeHandler: corev1.ProbeHandler{
+		Exec: &corev1.ExecAction{
+			Command: []string{
+				GetProbeMountDirectory() + "/" + GetStartupScriptName(),
```
gaurav-splunk, line 78: so this does not require an initial shell command like `/bin/bash` or something like that? Just providing the name of the script would work?

pkg/splunk/enterprise/configuration.go
```diff
@@ -1834,3 +1873,34 @@ func validateReadinessProbe(ctx context.Context, cr splcommon.MetaObject, readin
 
 	return err
 }
+
+// validateStartupProbe validates the liveness startup config
```
gaurav-splunk, line 1877: nit: validates startup probe config?

pkg/splunk/enterprise/standalone_test.go
```diff
@@ -50,6 +50,10 @@ func init() {
 		fileLocation, _ := filepath.Abs("../../../" + readinessScriptLocation)
```
gaurav-splunk, line 50: livenessScriptLocation?

pkg/splunk/enterprise/configuration.go
pdhanoya-splunk: Yes this is exactly same way we execute liveness and readiness probe.

sgontla (approved):

gaurav-splunk (approved):

vivekr-splunk (approved):

PR #932, opened 2022-10-10
Update Test to check cluster platform before running test

pdhanoya-splunk:
- Update test to check platform before running test setup
- Add feature branch to int test

### DIFF
--- a/test/appframework/azure/m4/appframework_azure_test.go
+++ b/test/appframework/azure/m4/appframework_azure_test.go
@@ -88,7 +88,7 @@ var _ = Describe("m4appfw test", func() {
 	})
 
 	Context("Multisite Indexer Cluster with Search Head Cluster (m4) with App Framework", func() {
-		It("smoke, m4azure, masterappframeworkm4azure, appframeworkazure: can deploy a M4 SVA with App Framework enabled, install apps and upgrade them", func() {
+		It("smoke, m4azure, masterappframeworkazurem4, appframeworkazure: can deploy a M4 SVA with App Framework enabled, install apps and upgrade them", func() {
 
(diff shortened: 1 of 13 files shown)


gaurav-splunk:
I have a basic question(and maybe I did not understand this at all). Why are we running aws test suite when `testenv.ClusterProvider = azure` or vice-versa? That would save us all the if/else checks in the code and if we can provide this check in a single place where it would run aws/azure test suite based on testenv.ClusterProvider.

tpereirasplunk (approved):

test/appframework/azure/m4/appframework_azure_test.go
```diff
@@ -839,7 +839,7 @@ var _ = Describe("m4appfw test", func() {
 	})
 
 	Context("Multi Site Indexer Cluster with Search Head Cluster (m4) with App Framework", func() {
-		It("integration, m4azure, masterappframeworkm4azure, appframeworkazure: can deploy a M4 SVA with App Framework enabled for manual poll", func() {
+		It("integration, m4azure, masterappframeworkazurem4azure, appframeworkazure: can deploy a M4 SVA with App Framework enabled for manual poll", func() {
```
tpereirasplunk, line 842: NIT: think azure got added twice to the label

pdhanoya-splunk:
@gaurav-splunk This the default behavior of Ginkgo. It collects all test cases and execute beforeSuite and then decides which test to execute based on TEST_FOCUS. I can create a common setup method which can filter based on the CLUSTER_PROVIDER. Just trying to get a successful run before trying that.

gaurav-splunk (approved):

PR #935, opened 2022-10-12
Address prodsec inputs from CSPL-2099

kumarajeet:
Updated the appframework azure blob doc based on prod sec inputs regarding adding recommendations for the right authorization method for the azure blob. 

Address a few minor changes as per the doc review.


kumarajeet:
Closing this PR and I will open another one to direct to develop

PR #958, opened 2022-10-31
Add CRD spec for ES app installation epic

akondur:
Adding CRD spec for ES app installation epic. There is a newly introduced scope `premiumApps` will be used for the `EnterpriseSecurity` app and any such premiumApps in the future(ITSI etc..). There is a field named `premiumAppsProps` which will give us further details on the premiumApps. For the ES app we have a field `esDefaults` which is used to provide further information such as `sslEnablement` flag for installation etc.


api/v4/common_types.go
```diff
@@ -348,9 +361,38 @@ type AppSourceDefaultSpec struct {
...
+// PremiumAppsProps represents properties for premium apps such as ES
+type PremiumAppsProps struct {
+	// Type: enterpriseSecurity for now, can accomodate itsi etc.. later
+	// +optional
+	Type string `json:"type,omitempty"`
+
+	// Enterpreise Security App defaults
+	// +optional
+	EsDefaults EsDefaults `json:"esDefaults,omitempty"`
+}
```
sgontla, line 382: Considering the future needs, can we define a new type called `PremiumAppsType`, then use it in the ES specific data structure?

akondur: The `type` here represents the type of premium app. It is under `PremiumAppsProps` which represents properties for premiumApps. The `type` field can have `enterpriseSecurity`, `itsi` or other premiumApps as values. 

With the suggestion, wouldn't we be incorrectly putting the type under the purview of the ES app(while it can mean itsi also)? Could you provide a psuedo struct as an example(to understand better)?

sgontla (approved):

api/v4/common_types.go
```diff
@@ -348,9 +361,39 @@ type AppSourceDefaultSpec struct {
...
+	//     strict: Ensure that SSL is enabled
+	//             in the web.conf configuration file to use
+	//             this mode. Otherwise, the installer exists
+	//	     	   with an error.
+	//     auto: Enables SSL in the etc/system/local/web.conf
+	//           configuration file. This is the DEFAULT mode
+	//           used by the operator if left empty.
+	//     ignore: Ignores whether SSL is enabled or disabled.
+	// +optional
+	SslEnablement string `json:"sslEnablement,omitempty"`
```
gaurav-splunk, line 396: How about something like this?? - 
```
type GenericDefaults interface{}

type PremiumAppsDefaults interface {
	GetDefaults() GenericDefaults
}

// PremiumAppsProps represents properties for premium apps such as ES
type PremiumAppsProps struct {
	// Type: enterpriseSecurity for now, can accomodate itsi etc.. later
	// +optional
	Type string `json:"type,omitempty"`

	PremiumAppsDefaults PremiumAppsDefaults `json:"premiumAppsDefaults,omitempty"`
}

type EsDefaultsHandler struct {
	// Enterpreise Security App defaults
	// +optional
	EsDefaults EsDefaults `json:"esDefaults,omitempty"`
}

var _ GenericDefaults = &EsDefaults{}

// EsDefaults captures defaults for the Enterprise Security App
type EsDefaults struct {
	// Sets the sslEnablement value for ES app installation
	//     strict: Ensure that SSL is enabled
	//             in the web.conf configuration file to use
	//             this mode. Otherwise, the installer exists
	//	     	   with an error.
	//     auto: Enables SSL in the etc/system/local/web.conf
	//           configuration file. This is the DEFAULT mode
	//           used by the operator if left empty.
	//     ignore: Ignores whether SSL is enabled or disabled.
	// +optional
	SslEnablement string `json:"sslEnablement,omitempty"`
}

func (EsDefaultsHandler *EsDefaultsHandler) GetDefaults() GenericDefaults {

	return EsDefaultsHandler.EsDefaults
}
```
This would make it more generic and in the future, implementation of other future premium apps such as ITSI can be easily integrated.

gaurav-splunk: I guess the use of interfaces is not supported while generating CRD spec, hence parking this approach for now.

gaurav-splunk (approved):

api/v4/common_types.go
akondur: Per our discussion, controller-gen throws the following error when using interface as a type:

```
/Users/akondur/Desktop/operator/test/splunk-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
github.com/splunk/splunk-operator/api/v4:-: invalid type: interface{}
github.com/splunk/splunk-operator/api/v4:-: invalid type: interface{GetDefaults() github.com/splunk/splunk-operator/api/v4.GenericDefaults}
Error: not all generators ran successfully
run `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -h` for usage
make: *** [generate] Error 1
I 
```

PR #965, opened 2022-11-04
fix flaky test on standalone

tpereirasplunk:


test/appframework_aws/s1/appframework_aws_test.go
```diff
@@ -1723,19 +1724,9 @@ var _ = Describe("s1appfw test", func() {
...
 			appVersion = "V1"
-			appFileList = testenv.GetAppFileList(appListV1)
-			standalonePod := []string{fmt.Sprintf(testenv.StandalonePod, deployment.GetName(), 0)}
-			standaloneAppSourceInfo := testenv.AppSourceInfo{CrKind: standalone.Kind, CrName: standalone.Name, CrAppSourceName: appSourceName, CrPod: standalonePod, CrAppVersion: appVersion, CrAppScope: enterpriseApi.ScopeLocal, CrAppList: appListV1, CrAppFileList: appFileList}
-			allAppSourceInfo := []testenv.AppSourceInfo{standaloneAppSourceInfo}
-			testenv.AppFrameWorkVerifications(ctx, deployment, testcaseEnvInst, allAppSourceInfo, splunkPodAge, "")
+			testenv.VerifyAppInstalled(ctx, deployment, testcaseEnvInst, testcaseEnvInst.GetName(), []string{fmt.Sprintf(testenv.StandalonePod, deployment.GetName(), 0)}, appListV1, false, "enabled", false, false)
 
 			// Check for changes in App phase to determine if next poll has been triggered
 			appFileList = testenv.GetAppFileList(appListV2)
```
jambrosiano, line 1732: we got this 3 times on lines 1721,1732,1743 but once is enough in this case, isn't it?

test/appframework_aws/s1/appframework_aws_test.go
```diff
@@ -1683,6 +1683,7 @@ var _ = Describe("s1appfw test", func() {
 			// ################## SETUP FOR STANDALONE ####################
 			// Download all test apps from S3
 			appVersion := "V1"
+			appListV1 := []string{appListV1[0]}
```
jambrosiano, line 1686: So, for V1 we take 1 app only, but for V2 later on we take the normal appListV2?
Is that what we want to do?

jambrosiano:
We would need this in appframework_az folder also btw

test/appframework_aws/s1/appframework_aws_test.go
tpereirasplunk: The test case involves updating the app as soon as we download it. If there are too many apps by the time we download the updated apps, the apps installed get updated and we are unable to check that previous install took place.

This added flakiness to the test. Changing the number of apps in this test case to counteract this issue

jambrosiano (approved):

gaurav-splunk (approved):

PR #998, opened 2022-12-06
CSPL-2178: es post install failure should not trigger update of app install

kumarajeet:
**Problem:**
As part of ES install in SHC, the operator executes the following steps:

1. Download ES
2. Install the ES app (`splunk install app <download_dir_on_pod>/splunk-enterprise-security_xxx.spl`)
3. Run ES post install commands `essinstall`
4. Cleanup the app archive

What we saw is that if step 3 fails, then the Operator Install worker thread tries run step 2 again that is -  install the ES app again with command `splunk install app <download_dir_on_pod>/splunk-enterprise-security_xxx.spl`

In this case, the App install command comes back with an error: `\nApp \"SplunkEnterpriseSecuritySuite\" already exists; use the \"update\" argument to install anyway\n", "err": "command terminated with exit code 22"`

This error comes back because the install worker thread does not know about the ES post-install failure and reruns step 2.

**Solution:**
In this fix, at step 2 above, the logic is enhanced so the install worker thread first checks if the app is already installed so it **_skips_** the install step.

Updated steps:
1. Download ES
2. Check if ES app is already installed and enabled (through splunk list app ...command)
3. If step 2 returns true, skip the app install ((`splunk install app <download_dir_on_pod>/splunk-enterprise-security_xxx.spl`)
5. Run ES post install commands `essinstall`
6. Cleanup the app archive

### DIFF
--- a/pkg/splunk/enterprise/afwscheduler.go
+++ b/pkg/splunk/enterprise/afwscheduler.go
@@ -760,11 +760,18 @@ func isAppAlreadyInstalled(ctx context.Context, cr splcommon.MetaObject, podExec
 
 	scopedLog.Info("check app's installation state")
 
-	command := fmt.Sprintf("/opt/splunk/bin/splunk list app %s| grep ENABLED; echo -n $?", installedAppName)
+	command := fmt.Sprintf("/opt/splunk/bin/splunk list app %s -auth admin:`cat /mnt/splunk-secrets/password`| grep ENABLED; echo -n $?", installedAppName)
 
 	streamOptions := splutil.NewStreamOptionsObject(command)
 
 	stdOut, stdErr, err := podExecClient.RunPodExecCommand(ctx, streamOptions, []string{"/bin/sh"})
+
+	if strings.Contains(stdErr, "Could not find object") {
+		// when app is not installed you will see something like on StdErr:
+		// "Could not find object id=<app_name>"
+		// which mean app is not installed (no need to check enabled at this time)
+		return false, nil
+	}
 
 	if stdErr != "" || err != nil {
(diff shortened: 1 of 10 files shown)


akondur (approved):
Looks good!

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -700,11 +700,42 @@ func installApp(rctx context.Context, localCtx *localScopePlaybookContext, cr sp
 		return fmt.Errorf("app pkg missing on Pod. app pkg path: %s", appPkgPathOnPod)
 	}
 
+	// gather installed app name
+	// this is the name of top folder when we untar the app archive package
+	if worker.appDeployInfo.InstalledAppName == "" {
```
gaurav-splunk, line 705: Since we are about to install the app, this name `InstalledAppName` is a misnomer. It should be something like `actualAppName` or something along those lines.

kumarajeet: Good point Gaurav. I have changed the `InstalledAppName` to `AppPackageTopFolder` to better match where this name is coming from. I have also left some comments for the existing AppName that this is really an app archive name.

gaurav-splunk (approved):

akondur:
> @akondur , what sets the `isUpdate` flag?

If the downloaded app's etag is different from the one we have stored we set the flag to true. Ref [here](https://github.com/splunk/splunk-operator/blob/main/pkg/splunk/enterprise/util.go#L1180).

akondur:
Hi @chipzzz , the app's name is unique to an app package within an app source. So if the name of the app changes, it will be treated as a separate app package and installed fresh. Please open a GitHub issue detailing the specifics of the exact problem seen along with supporting configs to help us understand the issue better. Be sure to include the version of the operator and Splunk Enterprise docker image used as well.

PR #1006, opened 2022-12-14
CSPL-2187 - Make ssl_enablement ignore as default, disable strict for shc

akondur:
Make ssl_enablement ignore as default, disable strict for shc

### DIFF
--- a/api/v4/common_types.go
+++ b/api/v4/common_types.go
@@ -387,11 +387,11 @@ type EsDefaults struct {
 	//     strict: Ensure that SSL is enabled
 	//             in the web.conf configuration file to use
 	//             this mode. Otherwise, the installer exists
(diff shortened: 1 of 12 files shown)


api/v4/common_types.go
```diff
@@ -387,11 +387,11 @@ type EsDefaults struct {
 	//     strict: Ensure that SSL is enabled
```
kumarajeet, line 387: Should we call out that :
For Search Head Cluster only `Ignore` option is supported by the operator.

kumarajeet (approved):
left a comment but otherwise it is good.

gaurav-splunk (approved):

api/v4/common_types.go
akondur: Hi @kumarajeet , let's take care of that in the documentation.

PR #1025, opened 2023-01-11
Improve error message

bdelvecchio:
If the credentials secret does not contain an expected key, then display the name of the missing key in the error message! This is important because I have not found the required key names documented anywhere else.


bdelvecchio:
closed for retargeting to develop

PR #1033, opened 2023-01-19
Fix ES failures in pipeline

tpereirasplunk:
The ES test failed sometimes in the pipeline. It was observed that the tests would pass when run locally or isolated on the pipeline. 

This could be because the ES app is almost 1GB in size which is close to the PV disk space on the operator

Fix: Moving the ES test to the end of the testfile so it runs alone

### DIFF
--- a/.github/workflows/int-test-workflow.yml
+++ b/.github/workflows/int-test-workflow.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - main
+      - CSPL-2219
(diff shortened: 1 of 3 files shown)


gaurav-splunk (approved):

akondur (approved):

gaurav-splunk:
Closing this PR since https://github.com/splunk/splunk-operator/pull/1036 is taking care of this fix.

PR #1047, opened 2023-01-26
Merge main to develop

gaurav-splunk:


akondur (approved):

.github/workflows/helm-test-workflow.yml
```diff
@@ -88,6 +88,7 @@ jobs:
         if: github.ref == 'ref/head/main'
```
vivekr-splunk, line 88: are we planning to change head to heads later ?

vivekr-splunk (approved):

.github/workflows/helm-test-workflow.yml
gaurav-splunk: will be taken up as part of tech debt item

Issue #1014, opened 2022-12-30
App Framework: Changes on cluster manager not get pulled from config map


bendermar:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
When changing the cluster manager yaml and push out to the environment only the config map gets updated. The changes are not visible inside of the pod himself. Seems same behaviour as in the CSPL-1529 bug. 

**Expected behavior**
Changes should get pulled from config map.

**Splunk setup on K8S**
1.24.6

**Reproduction/Testing steps**
Create a cluster manager. Then add a new index for example.

Operator version 2.1.1

akondur:
Hi @bendermar , could you:

1.  Please share the exact steps to reproduce the bug? i.e the exact clusterManager yamls before and after.
2.  Please share what is the expectation i.e what changes do we expect to see on the pod?
3. Are you using the latest way of configuring [appFramework](https://github.com/splunk/splunk-operator/blob/main/docs/AppFramework.md)?

Looking forward to hearing from you.

bendermar:
Hello Arjun,

we configured for testing purposes some indexes over the clustermanger yaml file. When we will add an index after first push the not will appear in the manage-apps folder only on the config map.
We also implemented APP Framework. Does it mean adding indexes over yaml is not supported? The documentation about all those things is not really good. We also not want to have unencrypted passwords which is the case in both variants.
Over App Framework the password is in cleartext on the S3 .

Here example of a new index that was added after installation.

***@***.***

Mit freundlichen Grüßen

Markus Bender
Monitoring Solutions

Schwarz IT GmbH & Co. KG
Stiftsbergstraße 1 · D-74172 Neckarsulm

Büro: D-74189 Weinsberg · Haller Straße 60

Telefon: +49 (0) 7132 30 455132
***@***.******@***.***>

Kommanditgesellschaft · Sitz: Neckarsulm
Registergericht Stuttgart · HRA 730995

akondur:
Hey @bendermar , while SmartStore can be used to configure indexes on the Indexer Cluster we recommend using AppFramework as the primary mechanism for configuration management with the Splunk Operator as mentioned in the smartstore guide [here](https://github.com/splunk/splunk-operator/blob/main/docs/SmartStore.md#smartstore-resource-guide).

[...]

Issue #1009, opened 2022-12-16
CrashLoopBackOff of Splunk Operator after upgrade to 2.1.1


markusspitzli:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
The Splunk Operator has been upgraded from 2.1.0 to 2.1.1. Since then the Operator keeps crashing. The Pod has the State CrashLoopBackOff. 

In the logfiles of the operator I see that the crash must somehow be related to updateing the CR STatus

```
2022-12-16T07:44:03.632038149Z INFO updateCRStatus Cache is reflecting the latest CR {"controller": "standalone", "controllerGroup": "enterprise.splunk.com", "controllerKind": "Standalone", "Standalone": {"name":"mycompany-prod","namespace":"mycompany-prod"}, "namespace": "mycompany-prod", "name": "mycompany-prod", "reconcileID": "aa4de5fa-607f-46a5-8f3a-d5917956da72", "original cr version": "186476862", "updated CR version": "186476862"}
34
2022-12-16T07:44:03.63207698Z INFO Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference {"controller": "standalone", "controllerGroup": "enterprise.splunk.com", "controllerKind": "Standalone", "Standalone": {"name":"mycompany-prod","namespace":"mycompany-prod"}, "namespace": "mycompany-prod", "name": "mycompany-prod", "reconcileID": "aa4de5fa-607f-46a5-8f3a-d5917956da72"}
33
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
32
panic: runtime error: invalid memory address or nil pointer dereference
31
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16f1c61]
30
29
goroutine 801 [running]:
28
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
27
/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:118 +0x1f4
26
panic({0x1927500, 0x2c5e6a0})
25
/usr/local/go/src/runtime/panic.go:884 +0x212
24
github.com/splunk/splunk-operator/pkg/splunk/enterprise.(*AppInstallPipeline).addWorkersToPipelinePhase(0xc000920540, {0x1f51998?, 0xc0000264b0?}, {0x0, 0x0}, {0xc000c6e838, 0x1, 0xa?})
23
[...]

PR #1053, opened 2023-02-06
CSPL-2255, 2256 - Fix smartstore symbolic links, add validation check for phase info

akondur:
**CSPL_2255** - [Github issue](https://github.com/splunk/splunk-operator/issues/1014)

**Problem Statement**

After `9.0.x`, the symbolic links setup by SmartStore on a ClusterManager are wiped out after bundle push by Splunk after the main container `splunk` has started. In operator version `1.1.0`, docker splunk `8.2.6` the symbolic links are retained even after container start/bundle push.

On a change in smartstore spec, this causes the bundle push triggered by the operator ineffective as the mounted config map changes doesn't show up in `/opt/splunk/etc/manager-apps`. 

**Fix**:

Re-add the symbolic links if smartstore is configured.

**CSPL-2256** - [Github issue](https://github.com/splunk/splunk-operator/issues/1009)

**Problem Statement**

The operator crashes as the `phaseInfo` for a couple of apps are wiped out with deletion of apps in S3. Issue is not reproducible.

**Preventive fix:**

If phaseInfo is not valid, log it out and prevent the crash.

### DIFF
--- a/pkg/splunk/enterprise/afwscheduler.go
+++ b/pkg/splunk/enterprise/afwscheduler.go
@@ -35,6 +35,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var appPhaseInfoStatuses = map[enterpriseApi.AppPhaseStatusType]bool{
+	enterpriseApi.AppPkgDownloadPending:     true,
+	enterpriseApi.AppPkgDownloadInProgress:  true,
+	enterpriseApi.AppPkgDownloadComplete:    true,
+	enterpriseApi.AppPkgDownloadError:       true,
+	enterpriseApi.AppPkgPodCopyPending:      true,
+	enterpriseApi.AppPkgPodCopyInProgress:   true,
+	enterpriseApi.AppPkgPodCopyComplete:     true,
+	enterpriseApi.AppPkgMissingFromOperator: true,
+	enterpriseApi.AppPkgPodCopyError:        true,
+	enterpriseApi.AppPkgInstallPending:      true,
+	enterpriseApi.AppPkgInstallInProgress:   true,
+	enterpriseApi.AppPkgInstallComplete:     true,
+	enterpriseApi.AppPkgMissingOnPodError:   true,
+	enterpriseApi.AppPkgInstallError:        true,
+}
+
 // isFanOutApplicableToCR confirms if a given CR needs fanOut support
 func isFanOutApplicableToCR(cr splcommon.MetaObject) bool {
 	switch cr.GetObjectKind().GroupVersionKind().Kind {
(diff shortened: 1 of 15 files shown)


pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1338,6 +1338,47 @@ func isPhaseStatusComplete(phaseInfo *enterpriseApi.PhaseInfo) bool {
...
+		enterpriseApi.AppPkgPodCopyInProgress:   true,
+		enterpriseApi.AppPkgPodCopyComplete:     true,
+		enterpriseApi.AppPkgMissingFromOperator: true,
+		enterpriseApi.AppPkgPodCopyError:        true,
+		enterpriseApi.AppPkgInstallPending:      true,
+		enterpriseApi.AppPkgInstallInProgress:   true,
+		enterpriseApi.AppPkgInstallComplete:     true,
+		enterpriseApi.AppPkgMissingOnPodError:   true,
+		enterpriseApi.AppPkgInstallError:        true,
+	}
```
sgontla, line 1373: Instead of filing it on every iteration, probably, we can fill it in a global map, and use that, wherever necessary.

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1338,6 +1338,47 @@ func isPhaseStatusComplete(phaseInfo *enterpriseApi.PhaseInfo) bool {
...
+	scopedLog := reqLogger.WithName("validatePhaseInfo").WithValues("phaseInfo", phaseInfo)
+
+	// Check for phase in phaseInfo
+	phases := string(
+		enterpriseApi.PhaseDownload +
+			enterpriseApi.PhasePodCopy +
+			enterpriseApi.PhaseInstall)
+
+	if !strings.Contains(phases, string(phaseInfo.Phase)) {
+		scopedLog.Info("Invalid phase in PhaseInfo")
```
sgontla, line 1353: nit: Error()?

pkg/splunk/enterprise/afwscheduler.go
```diff
@@ -1338,6 +1338,47 @@ func isPhaseStatusComplete(phaseInfo *enterpriseApi.PhaseInfo) bool {
...
+		enterpriseApi.AppPkgPodCopyError:        true,
+		enterpriseApi.AppPkgInstallPending:      true,
+		enterpriseApi.AppPkgInstallInProgress:   true,
+		enterpriseApi.AppPkgInstallComplete:     true,
+		enterpriseApi.AppPkgMissingOnPodError:   true,
+		enterpriseApi.AppPkgInstallError:        true,
+	}
+
+	if ok := statuses[phaseInfo.Status]; !ok {
+		scopedLog.Info("Invalid status in PhaseInfo")
```
sgontla, line 1376: nit: same here, Error()?

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -213,6 +223,14 @@ func ApplyClusterManager(ctx context.Context, client splcommon.ControllerClient,
...
 		}
 
+		// Reset symbolic links for clusterManager
+		if cr.Status.SmartStoreConfigured {
+			err = resetSymbolicLinks(ctx, client, cr, 1, podExecClient)
+			if err != nil {
+				return result, err
+			}
+		}
+
```
sgontla, line 233: I think, if you add this logic at [here](https://github.com/splunk/splunk-operator/pull/1053/files#diff-4fab20ab91ce601b65a8f1961f3809d7429faffe27c2c88c6c08956dc23a0a74R370), we can avoid the CR changes.

akondur: @sgontla The link seems to be broken. Also the symbolic links are broken every time a pod is restarted for any reason and hence I was thinking we need this logic here. If we modify only where smartstore is modified then the symbolic links might be broken in between?

sgontla: Seems like some issue with the link, anyhow, i was referring the line number 

> 	podExecClient := splutil.GetPodExecClient(c, cr, cmPodName)
	err := CheckIfsmartstoreConfigMapUpdatedToPod(ctx, c, cr, podExecClient)
	if err != nil {
		return err
	}

   <<<<<<<<<<<<< Moving the above code to hear, will avoid the CR changes, and that way no need to track through the SmartStoreConfigured flag >>>>>>>>>>>>>

	err = PushManagerAppsBundle(ctx, c, cr)
	if err == nil {
		scopedLog.Info("Bundle push success")
		cr.Status.BundlePushTracker.NeedToPushManagerApps = false
	}

sgontla: Just thinking through the main issue of softlinks, as the softlinks are no more reliable, wondering if we can just copy the files?

akondur: Per our discussion, removed the CR status flag and moved the function call to bundle push API

pdhanoya-splunk (approved):

sgontla (approved):

PR #1068, opened 2023-02-14
Cspl 2291: helm test case to test smartstore with standalone

vivekr-splunk:


kuttl/tests/helm/s1-with-smartstore/01-update-app-config.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
akondur, line 1: Is this yaml invoked somewhere? I see that `s1-config.yaml` is invoked but don't see `update-app-config.yaml`, `scaleup-s1.yaml` being invoked. I might be missing something here.

vivekr-splunk: removed scaleup values yaml file as its not required.

Issue #1052, opened 2023-02-01
Splunk Operator: PersistentVolumeClaim splunk-operator-app-download cannot be bound with volumeMode: Filesystem


vitalyrychkov:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
When I deploy the splunk-operator Helm chart in our bare-metal K8s cluster, the PersistentVolumeClaim splunk-operator-app-download cannot be bound with volumeMode: Filesystem and stays in the "Pending" state. 
Could you please add an option to use a storageClass instead?
If possible, also allow for defining the volume size.

**Expected behavior**
Splunk Operator pod to start successfully.

**Splunk setup on K8S**
Fresh install

**Reproduction/Testing steps**
kubectl create ns splunk-operator
 helm install -f values-operator.yaml splunk-operator splunk/splunk-operator -n splunk-operator

**K8s environment**
Bare-metal K8S v1.24.5

**Proposed changes(optional)**
Could you please add an option in the values file to use a storageClass instead of "volumeMode: Filesystem" in the https://github.com/splunk/splunk-operator/blob/main/helm-chart/splunk-operator/templates/app_download.yaml template?
If possible, also allow for defining the volume size.

Rendered spec example would be:
```
  volumes:
  - name: app-staging
    persistentVolumeClaim:
      claimName: splunk-operator-app-download
  resources:
    requests:
      storage: 5Gi
  storageClassName: sc-nfs
```

**K8s collector data(optional)**
N/A

**Additional context(optional)**

iaroslav-nakonechnikov:
@vitalyrychkov  have you already tried to redefine to whatever you need with help of https://kubernetes.io/docs/concepts/storage/volumes/#local (or any other volume type) ?

and which version do you install? in all previous versions pvc was forced in any case, but now it seems to work correct with custom definitions.

vivekr-splunk:
@vitalyrychkov I understand your issue now. currently we create pvc with default storageclass. and we ignore what has been provided in values.yaml file. we will fix this in next release. For now you can modify the template file if needed before applying helm charts.

vitalyrychkov:
Thank you @vivekr-splunk , will do this for the time being.

vitalyrychkov:
> @vitalyrychkov have you already tried to redefine to whatever you need with help of https://kubernetes.io/docs/concepts/storage/volumes/#local (or any other volume type) ?
> 
[...]

PR #1071, opened 2023-02-14
Cspl 2294: custom volume creation in splunk -operator

vivekr-splunk:
this is for the issue https://github.com/splunk/splunk-operator/issues/1052. now user can provide size and storage class while creating custom persistent storage volumes


helm-chart/splunk-operator/templates/deployment.yaml
```diff
@@ -108,7 +108,13 @@ spec:
 {{- toYaml . | nindent 8 }}
 {{- end }}
       terminationGracePeriodSeconds: {{ .Values.splunkOperator.terminationGracePeriodSeconds }}
-{{- with .Values.splunkOperator.volumes }}
+{{- if .Values.splunkOperator.volumes }}
       volumes:
-{{- toYaml . | nindent 8 }}
-{{- end }}
\ No newline at end of file
+{{- range $volume := .Values.splunkOperator.volumes }}
```
akondur, line 113: Do we support this without helm?

vivekr-splunk: yes we do support. customer can use any storage class type and size.

akondur (approved):
Looks good. Nit: Need to remove the branch from helm test workflow before merge

pdhanoya-splunk (approved):

Issue #394, opened 2021-06-28
ClusterMaster CRD Creates record conflict


ryanfaircloth:
Using the following cr leads to a new svc named `splunk-s1-indexer-service` the correct name for this record is `splunk-s1-cluster-master-service` this leads to a conflict with an indexercluster with the same name
```
apiVersion: enterprise.splunk.com/v1
kind: ClusterMaster
metadata:
  name: s1
  finalizers:
    - enterprise.splunk.com/delete-pvc
```

akondur:
The issue has been fixed on develop, will close the issue with next release.

akondur:
Closing with [2.2.1 release](https://github.com/splunk/splunk-operator/releases/tag/2.2.1)

PR #1073, opened 2023-02-14
Service name conflict for ClusterManager/IDXC names

akondur:
[Github issue 394](https://github.com/splunk/splunk-operator/issues/394) - ClusterManager service name record conflict

Previously the cluster master/manager CRs were using the component(instanceType) as indexer for its services. Fix the same by removing the extra service created unnecessarily.


vivekr-splunk (approved):
LGTM

pdhanoya-splunk (approved):

sgontla:
Just for my understanding, why is it specific to clustermanager, but not clustermaster?

akondur:
> Just for my understanding, why is it specific to clustermanager, but not clustermaster?

The problem exists in clusterMaster too. But since we are supporting backward compatibility only for it, we can keep the current working the same for it.

Moving forward for `clusterManager` we can remove it as it not needed.

sgontla (approved):

PR #1096, opened 2023-02-28
Added symlink int test and fix for Cluster Master CRD

pdhanoya-splunk:
- Ported Arjun's fix for Cluster Manager to Cluster Master [PR 1053](https://github.com/splunk/splunk-operator/pull/1053) 
- Added new test to check new index through smartstore for Cluster master and cluster manager CR's


test/smartstore/manager_smartstore_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("Smartstore test", func() {
 		}
 	})
 
-	Context("Standalone Deployment (S1)", func() {
+	XContext("Standalone Deployment (S1)", func() {
```
akondur, line 44: Did we want to disable this test?

test/smartstore/manager_smartstore_test.go
```diff
@@ -92,7 +92,7 @@ var _ = Describe("Smartstore test", func() {
 		})
 	})
 
-	Context("Standalone Deployment (S1)", func() {
+	XContext("Standalone Deployment (S1)", func() {
```
akondur, line 95: Same comment as above

pkg/splunk/enterprise/util.go
```diff
@@ -650,7 +650,7 @@ var resetSymbolicLinks = func(ctx context.Context, client splcommon.ControllerCl
 
 	// Create command for symbolic link creation
 	var command string
-	if crKind == "ClusterManager" {
+	if crKind == "ClusterManager" || crKind == "ClusterMaster" {
 		command = setSymbolicLinkCmanager
```
akondur, line 654: I believe for clusterMaster the location is etc/master-apps while the string points to [setSymbolicLinkCmanager](https://github.com/splunk/splunk-operator/blob/1deadcb2641c9c6d945222963acfcf26e351f436/pkg/splunk/enterprise/names.go#L132). Has the soft link been tested? Maybe I am missing something here.

pdhanoya-splunk: The location is same for Cluster Master as well in any 9.x.x splunk.

akondur (approved):

pkg/splunk/enterprise/clustermaster.go
```diff
@@ -195,9 +200,11 @@ func ApplyClusterMaster(ctx context.Context, client splcommon.ControllerClient,
 			}
 		}
 
+		// Create podExecClient
+		podExecClient := splutil.GetPodExecClient(client, cr, "")
+
```
gaurav-splunk, line 205: any specific reason for moving it out?

pdhanoya-splunk: This is direct port of fix from Arjun here : https://github.com/splunk/splunk-operator/pull/1053

akondur: @gaurav-splunk @pdhanoya-splunk This change is not required. It was a part of an earlier logic workflow which was changed later due to a review comment. IMO to maintain symmetry we can keep it as is for now.

vivekr-splunk (approved):

gaurav-splunk (approved):

PR #1099, opened 2023-03-01
Build issues

vivekr-splunk:
fixed build issue for develop branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,13 +73,13 @@ else
 	SCANNER_FILE = clair-scanner_windows_amd64.exe
 endif
 
-SED := sed -i 
+SED := sed -i
 ifeq ($(shell uname), Linux)
-	SED = sed -i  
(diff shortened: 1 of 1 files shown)


gaurav-splunk (approved):

Makefile
```diff
@@ -73,13 +73,13 @@ else
 	SCANNER_FILE = clair-scanner_windows_amd64.exe
 endif
 
-SED := sed -i
```
akondur, line 76: To understand this better, what was the issue we are solving here?

vivekr-splunk: this is existing code. the one I have fixed is in makefile build creation. I changed test to use ginkgo instead of using simple go test just because we are using it in integration and in unit test case. but some how the build creation fails as ginkgo is not installed. I added `setup/ginkgo` to required in before build creation

akondur (approved):

Issue #997, opened 2022-12-05
App Framework: Deployment breaks if Istio is used


markusspitzli:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
The app deployment is broken when istio is used by the splunk operator and the splunk enterprise instances. If Istio is disabled the app deployment works fine. 

The splunk operator throws the following error. 
`2022-12-05T11:42:28.555443825Z ERROR afwSchedulerEntry unable to create directory on splunk pod {"controller": "standalone", "controllerGroup": "enterprise.splunk.com", "controllerKind": "Standalone", "Standalone": {"name":"mycompany-int","namespace":"mycompany-int"}, "namespace": "mycompany-int", "name": "mycompany-int", "reconcileID": "032c228b-b03f-4e5c-b3f0-44cf0a7e9b85", "name": "mycompany-int", "namespace": "mycompany-int", "error": "unable to create directory on Pod at path=/operator-staging/appframework/local. stdout: , stdErr: , err: a container name must be specified for pod splunk-mycompany-int-standalone-0, choose one of: [istio-init splunk istio-proxy]"}`

**Expected behavior**
app deployment takes istio into account and delivers the app in the appropriate container. 

**Splunk setup on K8S**
The Splunk Operator (2.1.0) and Standalone Splunk Enterprise (9.0.2) Instance are installed in the same namespace. 

**Reproduction/Testing steps**
1. Install Istio
2. create a namespace with istio enabled: `istio-injection: enabled`
3. deploy splunk operator in this namespace
4. deploy a standalone instance in the same namespace
5. try to deploy a app by using the app framework. 

**K8s environment**
K8s Version 1.22.12

**Proposed changes(optional)**
treat pods as multi container and specify the splunk container you want to deploy apps.

markusspitzli:
@sgontla here you go

```
Name:         splunk-mycompany-int-standalone-0
Namespace:    mycompany-int
Priority:     0
Node:         32fb9f24-0118-49ab-bce7-96416aacbdaa/11.0.78.9
Start Time:   Mon, 05 Dec 2022 12:39:17 +0100
Labels:       app.kubernetes.io/component=standalone
              app.kubernetes.io/instance=splunk-mycompany-int-standalone
              app.kubernetes.io/managed-by=splunk-operator
              app.kubernetes.io/name=standalone
              app.kubernetes.io/part-of=splunk-mycompany-int-standalone
[...]

PR #1100, opened 2023-03-02
Cspl 2318: adding container name to pod exec command

vivekr-splunk:
addressing issue here https://github.com/splunk/splunk-operator/issues/997.


sgontla (approved):

pkg/splunk/util/util.go
```diff
@@ -228,6 +230,7 @@ func GetPodExecClient(client splcommon.ControllerClient, cr splcommon.MetaObject
 	return &PodExecClient{
 		client:        client,
 		cr:            cr,
+		name: 		   "splunk",
```
akondur, line 233: Nit: Just curious, is this the only podExecClient used? Is there any other place where we need to update this?

akondur (approved):

gaurav-splunk (approved):

vivekr-splunk:
i do not think we are doing anything other than pod exec call specifically for splunk container. so just this change should work fine.

PR #1109, opened 2023-03-08
Cspl 2322: FOSSA Scan fix

vivekr-splunk:


.github/workflows/prodsec-workflow.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
     - main
     - develop
+    - cspl-2322
```
pdhanoya-splunk, line 7: Just a reminder to remove CSPL-2322 before merge.

pdhanoya-splunk (approved):

.github/workflows/prodsec-workflow.yml
vivekr-splunk: removed

PR #1118, opened 2023-03-10
fixed image version in helm test

vivekr-splunk:
some of the test are using public image i am changing that to use image created in pipeline


.github/workflows/helm-test-workflow.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - main
+      - helm-test-image-fix
```
gaurav-splunk, line 7: need to remove this

gaurav-splunk (approved):

PR #1144, opened 2023-04-14
[Draft] CSPL-2304 - Added smartstore documentation

akondur:
Splunk Operator via the smartstore feature, sets the `repFactor` to `auto` in the `[default]` stanza of the `splunk-operator` app in order to help with the requirements [here](https://docs.splunk.com/Documentation/Splunk/9.0.4/Indexer/ConfigureSmartStore). This setting was applied to all indexes except for certain internal indexes such as `metrics`, `introspection`, `telemetry`, `_metrics_rollup`, `_configtracker` which have specific stanzas in the `_cluster` app with `repFactor` as `0`.

Modifying the documentation to:
- Inform the Splunk Admin of the default settings by the Splunk Operator and potential overrides of configuration
- README.md has smartstore as a required feature incorrectly. AppFramework is the way to go starting 1.0.2.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -99,7 +99,11 @@ information.
 ### What Storage Type To Use?
 
 The Kubernetes infrastructure must have access to storage that meets or exceeds the recommendations provided in the Splunk Enterprise storage type recommendations at [Reference Hardware documentation - what storage type to use for a given role?](https://docs.splunk.com/Documentation/Splunk/latest/Capacity/Referencehardware#What_storage_type_should_I_use_for_a_role.3F) In summary, Indexers with SmartStore need NVMe or SSD storage to provide the necessary IOPs for a successful Splunk Enterprise environment.
- 
+
+### Splunk SmartStore Required
+
+For production environments, we are requiring the use of Splunk SmartStore. As a Splunk Enterprise deployment's data volume increases, demand for storage typically outpaces demand for compute resources. [Splunk's SmartStore Feature](https://docs.splunk.com/Documentation/Splunk/latest/Indexer/AboutSmartStore) allows you to manage your indexer storage and compute resources in a ___cost-effective___ manner by scaling those resources separately. SmartStore utilizes a fast storage cache on each indexer node to keep recent data locally available for search and keep other data in a remote object store. Look into the [SmartStore Resource Guide](SmartStore.md) document for configuring and using SmartStore through operator.
+
 ## Installing the Splunk Operator
 
 A Kubernetes cluster administrator can install and start the Splunk Operator for specific namespace by running:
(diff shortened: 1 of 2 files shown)


sgontla:
I believe, we just need to highlight the situation surrounding the _introspection, _metrics, _metrics_rollup, and _configtracker indexes, covering:
1. Reason for those indexes being forced to `repFactor = 0` 
2. Workaround: Use the latest app framework to modify the repFactor for those indexes.
     Alternatively, do a onetime configuration by explicitly modifying it directly on the cluster manager. This may not go well, if the customer's configuration is coming from their internal repo. 
 
Can we also check, if there is a way, this config can be set through the Ansible environment variables.?

akondur:
> I believe, we just need to highlight the situation surrounding the _introspection, _metrics, _metrics_rollup, and _configtracker indexes, covering:
> 
> 1. Reason for those indexes being forced to `repFactor = 0`
> 2. Workaround: Use the latest app framework to modify the repFactor for those indexes.
>    Alternatively, do a onetime configuration by explicitly modifying it directly on the cluster manager. This may not go well, if the customer's configuration is coming from their internal repo.
> 
> Can we also check, if there is a way, this config can be set through the Ansible environment variables.?

@sgontla Although the issue surfaced points out the `repFactor` setting,  I believe the issue is generic to other settings like `maxDataSize` in the `splunk-operator` app. The `repFactor` being set to `0` by default is done on the Splunkd side and I am not sure if it's worth mentioning here(similar theory behind other settings in splunkd vs `splunk-operator` app). Based on the findings, I was aiming to correct the documentation for a couple of things:

1. README.md marks SmartStore as required - which is corrected with this change.
2. The configurations in `splunk-operator` app are trying to be helpful but not actually taking effect due to multiple factors. Clearly calling this out in the documentation will alert the customer that there might be other configurations which can override it.

In my research Ansible currently does not have a setting for `repFactor` in particular.

docs/README.md
```diff
@@ -99,10 +99,6 @@ information.
...
-
-
-### Splunk SmartStore Required
-For production environments, we are requiring the use of Splunk SmartStore. As a Splunk Enterprise deployment's data volume increases, demand for storage typically outpaces demand for compute resources. [Splunk's SmartStore Feature](https://docs.splunk.com/Documentation/Splunk/latest/Indexer/AboutSmartStore) allows you to manage your indexer storage and compute resources in a ___cost-effective___ manner by scaling those resources separately. SmartStore utilizes a fast storage cache on each indexer node to keep recent data locally available for search and keep other data in a remote object store. Look into the [SmartStore Resource Guide](SmartStore.md) document for configuring and using SmartStore through operator.
```
akondur, line 105: Needs to be corrected

Issue #1151, opened 2023-04-26
Splunk Operator: indexers statefulset deleted after image upgrade when using ClusterManager 


tmaltaverne:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
When doing a minor Splunk image version upgrade (for example, 9.0.4 to 9.0.4.1) in the indexercluster custom resource, the statefulset of the indexers is deleted, resulting in all the indexers being terminated at the same time.
There was a [fix that was addressing this issue](https://github.com/splunk/splunk-operator/pull/1041). However, the code change was done only in ApplyIndexerCluster function, which is [called ](https://github.com/splunk/splunk-operator/blob/f910f6cc9228b66844dcc80921c8c2ed55fdaedc/controllers/indexercluster_controller.go#L119) when the ClusterMasterRef is used. When the deployment is migrated to ClusterManagerRef, the function called is ApplyIndexerClusterManager. The fix wasn't done it this one, which keeps the [old behaviour](https://github.com/splunk/splunk-operator/blob/f910f6cc9228b66844dcc80921c8c2ed55fdaedc/pkg/splunk/enterprise/indexercluster.go#L188) of deleting the statefulset when there is an image change.

**Expected behavior**
Rolling restart of the indexers when docker image version is updated in the custom resource.

**Splunk setup on K8S**
Multisite indexer cluster, migrated to the ClusterManager custom resource.

**Reproduction/Testing steps**
Upgrade the _image_ of an IndexerCluster. The statefulset managed by this IndexerCluster is deleted and recreated.

**K8s environment**
Kubernetes 1.23

**Proposed changes(optional)**
Do the same fix in the ApplyIndexerClusterManager function.
I have tested that locally and validated the rolling restart in a development cluster. I can raise a [pull request](https://github.com/splunk/splunk-operator/compare/develop...tmaltaverne:splunk-operator:bugfix/indexer_deletion_upgrade) if it's fine.

akondur:
Hi @tmaltaverne , thank you for opening the issue as well as identifying the root cause. Please open the suggested pull request.

michal-tatusko-splunk:
This issue was fixed in the 2.5.0 SOK release.

PR #1153, opened 2023-04-27
Fix indexer deletion in upgrade scenario

tmaltaverne:
This PR addresses issue #1151.
There are 2 changes:

- Make sure that the fix of PR #1041 is applied in the cluster-manager usecase
- Fix the check on the image version. The format of the image field being like splunk/splunk-operator:8.2.0, it is needed to get the version that is after the colon. The previous check was only checking if the image was starting with 8 or 9, which is always false.


Philbow:
Discovered the same problem when triggering a helm upgrade. Unfortunately the operator is not logging out the deletion of the STS.

Any plans to fix the pull request in order to merge it for the upcoming release? This is currently blocking the seamless upgrade for our Splunk customers.

Issue #1160, opened 2023-05-11
Splunk Operator: RBAC/Auth Proxy Service Label Selector Mismatch


Astroknut:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
Request to update the Selector labels used in https://github.com/splunk/splunk-operator/blob/main/helm-chart/splunk-operator/templates/rbac/auth_proxy_service.yaml#L15 to use `splunk-operator.selectorLabels`. The current use of `splunk-operator.labels` includes multiple labels not present on the deployment template(https://github.com/splunk/splunk-operator/blob/main/helm-chart/splunk-operator/templates/deployment.yaml#L26). The auth proxy service is unable to create a valid Endpoints resource due to the mismatch in the specified selector labels.

**Expected behavior**
Auth proxy service is able create a valid Endpoints object

**Splunk setup on K8S**
- Helm Chart: splunk.github.io/splunk-operator
- Minimal value overrides, service port mapping for `metric` port enabled to match https://github.com/splunk/splunk-operator/blob/main/helm-chart/splunk-operator/values.yaml#L96-L99

**Reproduction/Testing steps**
- Steps to reproduce the bug. For an enhancement or feature request, please provide steps to test.

**K8s environment**
- v1.24 on AKS

**Proposed changes(optional)**
- P[roposed change, if any.](https://github.com/splunk/splunk-operator/blob/main/helm-chart/splunk-operator/templates/rbac/auth_proxy_service.yaml#L15) < Replace `splunk-operator.labels` with `splunk-operator.selectorLabels` value.

**K8s collector data(optional)**
- Please provide data collected from the [K8s collectors](https://github.com/splunk/splunk-operator/blob/develop/docs/K8SCollectors.md), if any.

**Additional context(optional)**
- Add any other context about the problem here.

patrykw-splunk:
Hi, Based on the provided code, it looks like the issue has been already fixed.

PR #1164, opened 2023-05-17
fix: Fixes auth proxy service selector labels

Astroknut:
Addresses https://github.com/splunk/splunk-operator/issues/1160

Updates auth proxy service selector labels value to `selectorLabels` to ensure service can attach to splunk-operator pod and create an Endpoints resource.


vivekr-splunk (commented):
Thanks for sending this fix. I checked the current branch against `develop`, and there is no remaining diff to merge here.

It looks like this change already landed through `#1233` (`89a39732`), so I think this PR is effectively a duplicate / already-resolved branch at this point.

PR #1147, opened 2023-04-20
CSPL-2341 - Enable private access links as endpoints for AWS client sessions

akondur:
- Enable support for private access links as endpoints for AWS client sessions
- Fix Splunk Operator upgrade documentation

PR #1165, opened 2023-05-17
CSPL-2388 - Fix helm int tests

akondur:
Fix helm int tests [failure](https://github.com/splunk/splunk-operator/actions/runs/4997225731/jobs/8951485896) caused by the [PR](https://github.com/splunk/splunk-operator/pull/1147).

- Extended the timeout for the helm test suite
- Passed the `TEST_VPC_ENDPOINT_URL` string as an env variable to be picked up by `c3_with_apps_private_link` helm test case.
- Fixed the `m4-with-operator` and `m4` helm test cases to include the right regions(`us-west-2c` was not annotated on the nodes, switched to `us-west-2d` instead).
- Modified the regular expression string to extract region for appFramework
- Added a timeout for aws, azure, minio clients for appFramework
- Upgraded eksctl version


.github/workflows/helm-test-workflow.yml
```diff
@@ -69,6 +70,7 @@ jobs:
       DEPLOYMENT_TYPE: "helm"
       HELM_REPO_PATH: "../../../../helm-chart"
       INSTALL_OPERATOR: "true"
+      TEST_VPC_ENDPOINT_URL: ${{ secrets.TEST_VPC_ENDPOINT_URL }}
```
gaurav-splunk, line 73: Can we instead use env variable as part of `.env` file in our project? Just don't want to fall into the trap again where we have to update the secrets again and we dont know why the tests failed when someone changed the secrets or something like that.

akondur: @vivekr-splunk and I discussed it. I proposed using this method in order to not have a VPC endpoint on an open sourced repo. In the future we can store these in a separate repo(gitlab?).

akondur: If the env method is preferred I can switch to that as well :)

gaurav-splunk (approved):

.github/workflows/helm-test-workflow.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - main
+      - cspl_2388
```
gaurav-splunk, line 7: Lets remove this if helm tests are passing now.

akondur:
Successful test runs:

[Helm tests](https://github.com/splunk/splunk-operator/actions/runs/5132948107/jobs/9234907754?pr=1165)
[Int tests](https://github.com/splunk/splunk-operator/actions/runs/5132948112/jobs/9234907745?pr=1165)
[Smoke tests](https://github.com/splunk/splunk-operator/actions/runs/5132948111/jobs/9235464867?pr=1165)

vivekr-splunk (approved):

Issue #1167, opened 2023-05-25
Splunk Operator: pods are being stucked in Terminating state.


iaroslav-nakonechnikov:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
Hello, 
after upgrading EKS from 1.25 to 1.26 we started to see, that some pods are not being terminated on-time. 
for a long time it wasn't clear what is happening and we asked for the help of AWS to get into the problem, as we thought it is connected to kubernetes, not to splunk. 

But recently we received a very nice update from support, which says about possible issues, and all of them are in splunk-operator. 

**Expected behavior**
all works as expected

**Splunk setup on K8S**
EKS with 1.26
splunk-operator controller with `edge` label

**Reproduction/Testing steps**
Install on AWS EKS, try to use latest versions. 
get Terminating stuck....

**Additional context(optional)**
- Add any other context about the problem here.

answer from AWS:
```
I was able to check some of the API server audit logs and observed a lot of 'patch' calls to the pods are failing with a 500 response code resulting in pods being stuck in 'terminating' state. The captured logs are showing the following event information:

        {
            "kind": "Event",
            "apiVersion": "audit.k8s.io/v1",
            "requestURI": "/api/v1/namespaces/splunk-operator/pods/splunk-site2-dev-indexer-1/status?fieldManager=PodGC&force=true",
            "verb": "patch",
            "user": {
                "username": "system:serviceaccount:kube-system:pod-garbage-collector",
                "uid": "b4406e61-f9c2-43df-89b1-a69fdb542b9f",
                "groups": [
                    "system:serviceaccounts",
                    "system:serviceaccounts:kube-system",
                    "system:authenticated"
                ]
            },
            "userAgent": "kube-controller-manager/v1.26.4 (linux/amd64) kubernetes/4a34796/system:serviceaccount:kube-system:pod-garbage-collector",
            "responseStatus": {
                "metadata": {},
                "status": "Failure",
                "message": "failed to create manager for existing fields: failed to convert new object (splunk-operator/splunk-site2-dev-indexer-1; /v1, Kind=Pod) to smd typed: .spec.containers[name=\"splunk\"].env: duplicate entries for key [name=\"SPLUNK_LICENSE_MASTER_URL\"]",
                "code": 500
            },
            "requestObject": {
                "kind": "Pod",
                "apiVersion": "v1",
                "metadata": {
                    "name": "splunk-site2-dev-indexer-1",
                    "namespace": "splunk-operator"
                },
[...]

PR #1176, opened 2023-06-13
cspl-2399: Splunk Operator: pods are being stuck in Terminating state

vivekr-splunk:
https://github.com/splunk/splunk-operator/issues/1167


pkg/splunk/enterprise/configuration.go
```diff
@@ -24,6 +24,7 @@ import (
 	"reflect"
 	"strconv"
 
+	"github.com/wk8/go-ordered-map/v2"
```
gaurav-splunk, line 27: @vivekr-splunk Probably we need to add `go.sum` too.

pkg/splunk/enterprise/configuration.go
```diff
@@ -999,6 +1000,24 @@ func updateSplunkPodTemplateWithConfig(ctx context.Context, client splcommon.Con
 	// append any extra variables
 	env = append(env, extraEnv...)
 
+	// check if there are any duplicate entries
+	// we use orderedmap so the test case can pass as json marshal
+	// expects order
+	if len(env) > 0 {
```
akondur, line 1006: Can we optimize to a single loop? Pseudo code below. We can add to ordered map?
```
func removeDuplicate[T string | int](sliceList []T) []T {
    allKeys := make(map[T]bool)
    list := []T{}
    for _, item := range sliceList {
        if _, value := allKeys[item]; !value {
            allKeys[item] = true
            list = append(list, item)
        }
    }
    return list
}
```

vivekr-splunk: We have not used generics anywhere in the code, I am not sure if this is the place to start. or may be its ok. will try adding this

akondur (approved):

gaurav-splunk (approved):

PR #1181, opened 2023-06-28
CSPL-2094, 2342: Upgrade Strategy for LM and CM

tgarg-splunk:
Implemented the design to enable Level 2 Support for the Splunk Operator. Specifically:
- Added `changeAnnotation` to trigger the reconcile loop of instances once the CR of the higher precedence is upgraded.
- Added `isReadyforUpgrade` and `getCurrentImage` to only allow upgrade of the Cluster Manager once the License Manager upgrade in complete and get the current (deployed) image of a given instance
- Added kuttl test cases for the upgrade
- Added and fixed unit tests


pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	if licenseManager.Status.Phase == enterpriseApi.PhaseReady {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// getClusterManagerCurrentImage gets the image of the pods of the clusterManager before any upgrade takes place,
+// returns the image, and error if something goes wring
+func getClusterManagerCurrentImage(ctx context.Context, c splcommon.ControllerClient, cr *enterpriseApi.ClusterManager) (string, error) {
```
gaurav-splunk, line 514: Is it possible to write a generic API for all this instead of individual APIs such as getClusterManagerCurrentImage and getLicenseManagerCurrentImage. I see almost everything is common in both the APIs, with the only exception of instanceType(i.e. SplunkClusterManager and SplunkLicenseManager) and `cr`.

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+
+	// check conditions for upgrade
+	if cr.Spec.Image != cmImage && lmImage == cr.Spec.Image && licenseManager.Status.Phase == enterpriseApi.PhaseReady {
+		return true, nil
+	}
+
+	// Temporary workaround to keep the clusterManager method working only when the LM is ready
+	if licenseManager.Status.Phase == enterpriseApi.PhaseReady {
+		return true, nil
+	}
```
gaurav-splunk, line 507: Sorry did not understand this part, why this workaround?

gaurav-splunk, line 507: Can these 2 be converted to a single condition like this - 
```
if ((cr.Spec.Image != cmImage && lmImage == cr.Spec.Image) || licenseManager.Status.Phase == enterpriseApi.PhaseReady) {
		return true, nil
}	
```

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
 	}
 	return extraEnv, err
 }
+
+// upgradeScenario checks if it is suitable to update the clusterManager based on the Status of the licenseManager, returns bool, err accordingly
+func upgradeScenario(ctx context.Context, c splcommon.ControllerClient, cr *enterpriseApi.ClusterManager) (bool, error) {
```
gaurav-splunk, line 456: nit:maybe change the name of the function to be something like `isReadyForUpgrade`?

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+				return image, nil
+			}
+
+		}
+	}
+
+	return "", nil
+}
+
+// changeClusterManagerAnnotations updates the checkUpdateImage field of the CLuster Manager Annotations to trigger the reconcile loop
```
gaurav-splunk, line 566: super nit: ....of the ClusterManager annotations ...

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	// create/update the checkUpdateImage annotation field
+	annotations["checkUpdateImage"] = image
+
+	clusterManagerInstance.SetAnnotations(annotations)
+	err = c.Update(ctx, clusterManagerInstance)
+	if err != nil {
+		eventPublisher.Warning(ctx, "changeClusterManagerAnnotations", fmt.Sprintf("Could not update annotations. Reason %v", err))
+		scopedLog.Error(err, "ClusterManager types update after changing annotations failed with", "error", err)
+		return err
+	}
```
gaurav-splunk, line 606: Maybe this whole piece of code can be a common code since it is same in `changeMonitoringConsoleAnnotations` as well? So all these lines can be a part of a generic API which can be called from both `changeMonitoringConsoleAnnotations` and  `changeClusterManagerAnnotations` ?

Makefile
```diff
@@ -137,7 +137,7 @@ build: setup/ginkgo manifests generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build: test ## Build docker image with the manager.
+docker-build: #test ## Build docker image with the manager.
```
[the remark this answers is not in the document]
tgarg-splunk: I had done this in the beginning when I wanted to test the code locally and the unit tests were failing. Will revert it back.

pkg/splunk/enterprise/clustermanager.go
tgarg-splunk: According to the initial condition, any update in the CM will only happen if there was a change in the image. To avoid this and for the situation where the LM has an image tag greater than the CM (still valid), this workaround was added.

tgarg-splunk: The second condition was a workaround so wasn't sure if it should be included in the main condition?

pkg/splunk/enterprise/clustermanager.go
vivekr-splunk: example here 
```
// getClusterManagerCurrentImage gets the image of the pods of the clusterManager before any upgrade takes place,
// returns the image, and error if something goes wring
func getCRCurrentImage (ctx context.Context, c splcommon.ControllerClient, objectMeta metav1.ObjectMeta ) (string, error) {

	reqLogger := log.FromContext(ctx)
	scopedLog := reqLogger.WithName("getClusterManagerCurrentImage").WithValues("name", objectMeta.GetName(), "namespace", objectMeta.GetNamespace())

	namespacedName := types.NamespacedName{
		Namespace: objectMeta.GetNamespace(),
		Name:      GetSplunkStatefulsetName(SplunkClusterManager, objectMeta.GetName()),
	}
	statefulSet := &appsv1.StatefulSet{}
	err := c.Get(ctx, namespacedName, statefulSet)
	if err != nil {
		scopedLog.Error(err, "StatefulSet types not found in namespace", "namsespace", objectMeta.GetNamespace())
		return "", err
	}
	labelSelector, err := metav1.LabelSelectorAsSelector(statefulSet.Spec.Selector)
	if err != nil {
		scopedLog.Error(err, "Unable to get labels")
		return "", err
	}

	// get a list of all pods in the namespace with matching labels as the statefulset
	statefulsetPods := &corev1.PodList{}
	opts := []rclient.ListOption{
		rclient.InNamespace(objectMeta.GetNamespace()),
		rclient.MatchingLabelsSelector{Selector: labelSelector},
	}

	err = c.List(ctx, statefulsetPods, opts...)
	if err != nil {
		scopedLog.Error(err, "Pods types not found in namespace", "namsespace", objectMeta.GetNamespace())
		return "", err
	}

	// find the container with the phrase 'splunk' in it
	for _, v := range statefulsetPods.Items {
		for _, container := range v.Status.ContainerStatuses {
			if strings.Contains(container.Name, "splunk") {
				image := container.Image
				return image, nil
			}

		}
	}

	return "", nil
}
```

and then call 
```cmImage, err := getCRCurrentImage(ctx, c, &cr.ObjectMeta)```

File: pkg/splunk/enterprise/monitoringconsole.go [hunk 7]
```diff
@@ -355,3 +355,40 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+	if err != nil && k8serrors.IsNotFound(err) {
+		return nil
+	}
+	image, _ := getClusterManagerCurrentImage(ctx, client, cr)
+	annotations := monitoringConsoleInstance.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	if _, ok := annotations["checkUpdateImage"]; ok {
+		if annotations["checkUpdateImage"] == image {
```
vivekr-splunk, line 378: change this to lower case `splunk/image-tag` something like this

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	// create new object
+	licenseManager := &enterpriseApi.LicenseManager{}
+
+	// get the license manager referred in cluster manager
+	err = c.Get(ctx, namespacedName, licenseManager)
+	if err != nil {
+		return true, nil
+	}
+
+	lmImage, err := getLicenseManagerCurrentImage(ctx, c, licenseManager)
```
akondur, line 486: Can we pass multiple arguments here and return multiple images?

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	}
+	cmImage, err := getClusterManagerCurrentImage(ctx, c, cr)
+	if err != nil {
+		eventPublisher.Warning(ctx, "upgradeScenario", fmt.Sprintf("Could not get the Cluster Manager Image. Reason %v", err))
+		scopedLog.Error(err, "Unable to clusterManager current image")
+		return false, err
+	}
+
+	// check conditions for upgrade
+	if cr.Spec.Image != cmImage && lmImage == cr.Spec.Image && licenseManager.Status.Phase == enterpriseApi.PhaseReady {
```
akondur, line 500: Nit: What happens when there are updates to CM and LM other than the container image? Does level 2 upgrade include only images and not other spec?

akondur: Flip the conditions here and return false here?

For example:
```
if cr.Spec.Image == cmImage && lmImage != cr.Spec.Image && licenseManager.Status.Phase != enterpriseApi.PhaseReady {
    return <upgradeInProgressBool>, nil
}
```

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +451,160 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+
+	reqLogger := log.FromContext(ctx)
+	scopedLog := reqLogger.WithName("changeClusterManagerAnnotations").WithValues("name", cr.GetName(), "namespace", cr.GetNamespace())
+	eventPublisher, _ := newK8EventPublisher(c, cr)
+
+	namespacedName := types.NamespacedName{
+		Namespace: cr.GetNamespace(),
+		Name:      cr.Spec.ClusterManagerRef.Name,
+	}
+	clusterManagerInstance := &enterpriseApi.ClusterManager{}
```
akondur, line 578: Do we always configure a `ClusterManagerRef` on LM? If not, we can list all CM's in the namespace and find the one with the correct LM Ref

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +442,95 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+		return false, err
+	}
+
+	// check conditions for upgrade
+	annotations := cr.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	if _, ok := annotations["splunk/image-tag"]; ok {
+		if (cr.Spec.Image != cmImage) && (licenseManager.Status.Phase != enterpriseApi.PhaseReady || licenseManager.Spec.Image != annotations["splunk/image-tag"]) {
```
akondur, line 496: nit: Can we add a comment here. Something like:

Determining if there is an upgrade is going on & whether clusterManager is ready for the upgrade

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +442,95 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+			return false, nil
+		}
+	} else {
+		eventPublisher.Warning(ctx, "isClusterManagerReadyForUpgrade", fmt.Sprintf("Could not find the annotations. Reason %v", err))
+	}
+
+	return true, nil
+}
+
+// changeClusterManagerAnnotations updates the checkUpdateImage field of the clusterManager annotations to trigger the reconcile loop
```
akondur, line 506: Nit: annotation is "splunk/image-tag"?

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +442,95 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	clusterManagerInstance := &enterpriseApi.ClusterManager{}
+	err := c.Get(ctx, namespacedName, clusterManagerInstance)
+	if err != nil && k8serrors.IsNotFound(err) {
+		return nil
+	}
+
+	image, _ := getCurrentImage(ctx, c, cr, SplunkLicenseManager)
+
+	err = changeAnnotations(ctx, c, image, clusterManagerInstance)
+
```
akondur, line 527: Nit: Don't need this extra line here?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +355,34 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+		Namespace: cr.GetNamespace(),
+		Name:      cr.Spec.MonitoringConsoleRef.Name,
+	}
+	monitoringConsoleInstance := &enterpriseApi.MonitoringConsole{}
+	err := client.Get(ctx, namespacedName, monitoringConsoleInstance)
+	if err != nil && k8serrors.IsNotFound(err) {
+		return nil
+	}
+	image, _ := getCurrentImage(ctx, client, cr, SplunkClusterManager)
+
```
akondur, line 377: Same as above.

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,50 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
...
+		Namespace: cr.GetNamespace(),
+		Name:      GetSplunkStatefulsetName(instanceType, cr.GetName()),
+	}
+	statefulSet := &appsv1.StatefulSet{}
+	err := c.Get(ctx, namespacedName, statefulSet)
+	if err != nil {
+		return "", err
+	}
+
+	image := statefulSet.Spec.Template.Spec.Containers[0].Image
```
akondur, line 2290: Can we make sure that the slice `statefulSet.Spec.Template.Spec.Containers` is not empty?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +355,34 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
 		}
 	}
 }
+
+// changeMonitoringConsoleAnnotations updates the checkUpdateImage field of the Monitoring Console Annotations to trigger the reconcile loop
+// on update, and returns error if something is wrong.
+func changeMonitoringConsoleAnnotations(ctx context.Context, client splcommon.ControllerClient, cr *enterpriseApi.ClusterManager) error {
```
akondur, line 361: Nit: Is there a way where we can make this API generic? I know there are some places where the API has to be CR specific.
cc: @vivekr-splunk

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,50 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
 	}
 	return fmt.Sprintf("splunk-%s-%s-%d", cr.GetName(), podType, ordinalIdx)
 }
+
+// getClusterManagerCurrentImage gets the image of the pods of the clusterManager before any upgrade takes place,
+// returns the image, and error if something goes wrong
+func getCurrentImage(ctx context.Context, c splcommon.ControllerClient, cr splcommon.MetaObject, instanceType InstanceType) (string, error) {
+
```
akondur, line 2279: Nit: do we need this line here?

pkg/splunk/enterprise/monitoringconsole.go
akondur: Also, we can have an API like `changeAnnotations` as a generic API and depending on the type of CR take a specific action i.e the dependencies might be different based on the CRs.

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,50 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
...
+	image := statefulSet.Spec.Template.Spec.Containers[0].Image
+
+	return image, nil
+
+}
+
+// changeAnnotations updates the checkUpdateImage field of the CLuster Manager Annotations to trigger the reconcile loop
+// on update, and returns error if something is wrong
+func changeAnnotations(ctx context.Context, c splcommon.ControllerClient, image string, cr splcommon.MetaObject) error {
+
```
akondur, line 2299: nit: can we remove this empty line?

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,50 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
...
+	annotations := cr.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	if _, ok := annotations["splunk/image-tag"]; ok {
```
akondur, line 2304: The two return values of the annotations map lookup are value, bool(whether there is an entry in the map). The second lookup is unnecessary? We can have something like:

```
if val, ok := annotations["splunk/image-tag"]; ok && (val == img) {
    return nil
}
```

akondur (commented):
Added a few more comments

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,49 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
...
+		Namespace: cr.GetNamespace(),
+		Name:      GetSplunkStatefulsetName(instanceType, cr.GetName()),
+	}
+	statefulSet := &appsv1.StatefulSet{}
+	err := c.Get(ctx, namespacedName, statefulSet)
+	if err != nil {
+		return "", err
+	}
+
+	if statefulSet.Spec.Template.Spec.Containers == nil {
```
akondur, line 2288: Can we make this something like:

```
if len(statefulSet.Spec.Template.Spec.Containers) > 0 {
      return statefulSet.Spec.Template.Spec.Containers[0].Image
}

return "", <error stating that we couldn't find an image>
```
We need to catch the error condition where the image is not present in the spec to avoid a silent ignore of it.

pkg/splunk/enterprise/util.go
akondur: What I meant here is that we can avoid the nested if and a second evaluation of `annotations["splunk/image-tag"]`. I would keep it as stated above to keep the code cleaner and still achieve the same functionality.

akondur (approved):
Looks good. Great work @tgarg-splunk 

cc: @gaurav-splunk @vivekr-splunk

pkg/splunk/enterprise/clustermanager.go
gaurav-splunk: not necessarily, we may/may not configure CM with LM

pkg/splunk/enterprise/monitoringconsole.go
tgarg-splunk: There is already a `changeAnnotations` API which does the common actions for all the instanceTypes. To have a fully generic API, the number of arguments which will be required will increase because we want to get the instance from the reference along with the image of the caller instance which I think can become a bit cluttered.
Do you reckon the common API will be a better way to go forward?

pkg/splunk/enterprise/util.go
tgarg-splunk: I think throwing an error could cause an issue during the initial deployment. Is it possible that the stateful set is initialized but the container is still in creation? If yes, this would stop the instance reconciliation/set the phase of the instance to `Error` till the container is fully up. The reconcile will not get stuck though because of the multiple attempts that take place.

tgarg-splunk: I think throwing an error could cause an issue during the initial deployment. Is it possible that the stateful set is initialized but the container is still in creation? If yes, this would stop the instance reconciliation/set the phase of the instance to `Error` till the container is fully up. The reconcile will not get stuck though because of the multiple attempts that take place.

akondur: We are reading the statefulSet spec from etcd and not checking if the container is actually UP so we should be fine.

pkg/splunk/enterprise/monitoringconsole.go
akondur: Makes sense to keep this as-is.

pkg/splunk/enterprise/clustermanager.go
akondur: In that case, we should list all CM's and find the one folding LMRef here?

pkg/splunk/enterprise/util.go
```diff
@@ -2272,3 +2272,49 @@ func getApplicablePodNameForK8Probes(cr splcommon.MetaObject, ordinalIdx int32)
...
+	statefulSet := &appsv1.StatefulSet{}
+	err := c.Get(ctx, namespacedName, statefulSet)
+	if err != nil {
+		return "", err
+	}
+
+	if len(statefulSet.Spec.Template.Spec.Containers) > 0 {
+		return statefulSet.Spec.Template.Spec.Containers[0].Image, nil
+	}
+	err = fmt.Errorf("Unable to get image from statefulset of type %s.", instanceType.ToString())
```
akondur, line 2291: Nit: We can make this as follows:

```
return "", fmt.Errorf("Unable to get image from statefulset of type %s.", instanceType.ToString())
```

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +443,115 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+		}
+
+		if len(clusterManagerInstance.GetName()) == 0 {
+			return nil
+		}
+	}
+
+	image, _ := getCurrentImage(ctx, c, cr, SplunkLicenseManager)
+	err := changeAnnotations(ctx, c, image, clusterManagerInstance)
+
```
akondur, line 549: Nit: The empty line here can be removed.

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +356,58 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+		opts := []rclient.ListOption{
+			rclient.InNamespace(cr.GetNamespace()),
+		}
+		objectList := enterpriseApi.MonitoringConsoleList{}
+		err := client.List(ctx, &objectList, opts...)
+		if err != nil && k8serrors.IsNotFound(err) {
+			return nil
+		}
+
+		// check with instance has the required ClusterManagerRef
```
akondur, line 389: *check if

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +356,58 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+		}
+
+		if len(monitoringConsoleInstance.GetName()) == 0 {
+			return nil
+		}
+	}
+
+	image, _ := getCurrentImage(ctx, client, cr, SplunkClusterManager)
+	err := changeAnnotations(ctx, client, image, monitoringConsoleInstance)
+
```
akondur, line 404: Nit: Can we remove the empty line here?

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +443,115 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+				break
+			}
+		}
+
+		if len(clusterManagerInstance.GetName()) == 0 {
+			return nil
+		}
+	}
+
+	image, _ := getCurrentImage(ctx, c, cr, SplunkLicenseManager)
```
akondur, line 547: Are we ignoring the error here?

tgarg-splunk: Made the change to handle the error if any

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +443,125 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	if err != nil && k8serrors.IsNotFound(err) {
+		return true, nil
+	}
+
+	namespacedName = types.NamespacedName{Namespace: cr.GetNamespace(), Name: licenseManagerRef.Name}
+	licenseManager := &enterpriseApi.LicenseManager{}
+
+	// get the license manager referred in cluster manager
+	err = c.Get(ctx, namespacedName, licenseManager)
+	if err != nil {
```
vivekr-splunk, line 477: only when 
```
import (k8serrors "k8s.io/apimachinery/pkg/api/errors")
if k8serrors.IsNotFound(err) {
   return true, nil
} 
```
any other err return false

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +443,125 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+	clusterManagerInstance := &enterpriseApi.ClusterManager{}
+	if len(cr.Spec.ClusterManagerRef.Name) > 0 {
+		// if the LicenseManager holds the ClusterManagerRef
+		namespacedName := types.NamespacedName{
+			Namespace: cr.GetNamespace(),
+			Name:      cr.Spec.ClusterManagerRef.Name,
+		}
+		err := c.Get(ctx, namespacedName, clusterManagerInstance)
+		if err != nil {
+			if err.Error() == "NotFound" || k8serrors.IsNotFound(err) {
```
vivekr-splunk, line 521: err.Error() == "NotFound" is not needed I think

pkg/splunk/enterprise/clustermanager.go
```diff
@@ -434,3 +443,125 @@ func VerifyCMisMultisite(ctx context.Context, cr *enterpriseApi.ClusterManager,
...
+		}
+	} else {
+		// List out all the ClusterManager instances in the namespace
+		opts := []rclient.ListOption{
+			rclient.InNamespace(cr.GetNamespace()),
+		}
+		objectList := enterpriseApi.ClusterManagerList{}
+		err := c.List(ctx, &objectList, opts...)
+		if err != nil {
+			if err.Error() == "NotFound" || k8serrors.IsNotFound(err) {
```
vivekr-splunk, line 534: err.Error() == "NotFound" is not needed I think

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +356,67 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+	if len(cr.Spec.MonitoringConsoleRef.Name) > 0 {
+		// if the ClusterManager holds the MonitoringConsoleRef
+		namespacedName := types.NamespacedName{
+			Namespace: cr.GetNamespace(),
+			Name:      cr.Spec.MonitoringConsoleRef.Name,
+		}
+		err := client.Get(ctx, namespacedName, monitoringConsoleInstance)
+		if err != nil {
+			if err.Error() == "NotFound" || k8serrors.IsNotFound(err) {
+				return nil
```
vivekr-splunk, line 377: err.Error() == "NotFound" is not needed I think

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -355,3 +356,67 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+		}
+	} else {
+		// List out all the MonitoringConsole instances in the namespace
+		opts := []rclient.ListOption{
+			rclient.InNamespace(cr.GetNamespace()),
+		}
+		objectList := enterpriseApi.MonitoringConsoleList{}
+		err := client.List(ctx, &objectList, opts...)
+		if err != nil {
+			if err.Error() == "NotFound" || k8serrors.IsNotFound(err) {
```
vivekr-splunk, line 389: I think `List` will never throw IsNotFound

akondur (approved):

pkg/splunk/enterprise/clustermanager.go
tgarg-splunk: The ApplyLicenseManager test case fails without the check

gaurav-splunk (approved):

PR #1194, opened 2023-07-17
CSPL-2343: Upgrade Strategy for MC

tgarg-splunk:
Implemented the design for the MC upgrade strategy for the Splunk Operator Level II support. Specifically:

- Added `changeMonitoringConsoleAnnotation` to trigger the reconcile loop of instances once the CR of the higher precedence is upgraded.
- Added `isMonitoringConsoleReadyforUpgrade` to check if CM is ready before a version upgrade takes place.
- Added and fixed unit tests


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -357,6 +363,64 @@ func DeleteURLsConfigMap(revised *corev1.ConfigMap, crName string, newURLs []cor
...
+
+	mcImage, err := getCurrentImage(ctx, c, cr, SplunkMonitoringConsole)
+	if err != nil {
+		eventPublisher.Warning(ctx, "isMonitoringConsolerReadyForUpgrade", fmt.Sprintf("Could not get the Monitoring Console Image. Reason %v", err))
+		scopedLog.Error(err, "Unable to get monitoring console current image")
+		return false, err
+	}
+
+	// check if an image upgrade is happening and whether the ClusterManager is ready for the upgrade
+	if (cr.Spec.Image != mcImage) && (clusterManager.Status.Phase != enterpriseApi.PhaseReady || cmImage != cr.Spec.Image) {
```
akondur, line 417: Can we add a comment here:

// Upgrade in progress(CM CR not upgraded fully yet), return false to stop further reconcile operations on MC until CM is ready

Similarly [here](https://github.com/splunk/splunk-operator/blob/b83b304f710dd18f3444401b5108a4e85f4ce0b9/pkg/splunk/enterprise/clustermanager.go#L501) as well.

akondur (approved):

vivekr-splunk (approved):

PR #1197, opened 2023-07-19
CSPL-2344 Upgrade Strategy for SHC

tgarg-splunk:
Implemented the design for the SHC upgrade strategy for the Splunk Operator Level II support. Specifically:

- Added `isSearchHeadReadyforUpgrade` to check if MC is ready before a version upgrade takes place.
- Added and fixed unit tests


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -149,6 +149,14 @@ func ApplyMonitoringConsole(ctx context.Context, client splcommon.ControllerClie
 	if cr.Status.Phase == enterpriseApi.PhaseReady {
 		finalResult := handleAppFrameworkActivity(ctx, client, cr, &cr.Status.AppContext, &cr.Spec.AppFrameworkConfig)
 		result = *finalResult
+
+		// TODO: Fix the Change Annotation logic/ find an alternative (eg. state machine); right now the search head deployer
```
akondur, line 153: Is there any reason why the replicas don't come up properly?

tgarg-splunk: While running the tests the deployer was going down and some times (not every time) the replica-2 went down and went into a restart loop. This got fixed after rerunning the tests though. There is still one test case that fails in the 1st commit of the PR which can pass eventually (it checks if the LM is configured properly and because replica - 2 goes down this test fails but will pass if we wait for sometime).

akondur: Is there a reason why the pods are going down? Ideally we should be taking care of those in this MR or create a task to take care of it later.

akondur (approved):

pkg/splunk/enterprise/monitoringconsole.go
tgarg-splunk: From what I gathered from the test cases it is happening because of the change annotation which is why I have removed the call for that in the final commit ( the pods are not going down now). We had earlier thought that the pods should not restart with a change in the annotations but it is happening for the search head

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -667,3 +673,64 @@ func getSearchHeadClusterList(ctx context.Context, c splcommon.ControllerClient,
...
+
+	namespacedName := types.NamespacedName{
+		Namespace: cr.GetNamespace(),
+		Name:      GetSplunkStatefulsetName(SplunkSearchHead, cr.GetName()),
+	}
+
+	// check if the stateful set is created at this instance
+	statefulSet := &appsv1.StatefulSet{}
+	err := c.Get(ctx, namespacedName, statefulSet)
+	if err != nil && k8serrors.IsNotFound(err) {
```
vivekr-splunk, line 698: if we just want to return with any err then we do not have to check for IsNotFound

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -667,3 +673,64 @@ func getSearchHeadClusterList(ctx context.Context, c splcommon.ControllerClient,
...
+		return true, nil
+	}
+
+	namespacedName = types.NamespacedName{Namespace: cr.GetNamespace(), Name: monitoringConsoleRef.Name}
+	monitoringConsole := &enterpriseApi.MonitoringConsole{}
+
+	// get the monitoring console referred in search head cluster
+	err = c.Get(ctx, namespacedName, monitoringConsole)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
tgarg-splunk: isNotFound is done here to check if the monitoring console exists or not. We return true if it doesn't but if it does then we return false if an error is there

pkg/splunk/enterprise/searchheadcluster.go
tgarg-splunk: Should we instead return a false here if an error occurs and it is not the notFound error?

pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -149,6 +149,14 @@ func ApplyMonitoringConsole(ctx context.Context, client splcommon.ControllerClie
...
+		// starts terminating with this and few replicas do not come up properly after that
+
+		// err = changeSearchHeadAnnotations(ctx, client, cr)
```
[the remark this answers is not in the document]
tgarg-splunk: The changeAnnotation code for the SHC is already removed. Only the comment is there

vivekr-splunk (approved):

pkg/splunk/enterprise/monitoringconsole.go
akondur: @vivekr-splunk I believe we are appending the annotations from CR to the statefulSet [here](https://github.com/splunk/splunk-operator/blob/6f2e605134adf7a7d07310e1f1651797d1091121/pkg/splunk/enterprise/configuration.go#L709). Since we identify the differences [here](https://github.com/splunk/splunk-operator/blob/6f2e605134adf7a7d07310e1f1651797d1091121/pkg/splunk/controller/util.go#L61) we end up doing a restart of all the pods of the SS. The time taken to restart the deployer/Search heads seems to be the issue here - it is failing the tests.

gaurav-splunk:
Closing this since it will be covered as part of https://github.com/splunk/splunk-operator/pull/1216

PR #1203, opened 2023-07-25
removed duplicate dep chart 2.3 updated 2.3.0

vivekr-splunk:
refreshed 2.3.0 images, as they had entries for 2.3


docs/index.yaml
```diff
@@ -21,11 +21,11 @@ entries:
     name: splunk-enterprise
     type: application
     urls:
-    - https://splunk.github.io/splunk-operator/splunk-enterprise-2.3.0.tgz
+    - https://vivekr-splunk.github.io/splunk-operator/splunk-enterprise-2.3.0.tgz
```
gaurav-splunk, line 24: this is not correct, please correct it to point to splunk.github.io  and same for other places as well

vivekr-splunk: fixed

gaurav-splunk (approved):

Issue #1217, opened 2023-08-16
Splunk Operator: Cannot retrieve apps from S3 when using IRSA


tnycum:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
- On versions of splunk operator newer than 2.2.0 when using AWS [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), the operator controller Pod logs an error about not being able to retrieve apps stored in S3
- Sample error message:

```
ERROR   GetAppListFromRemoteBucket      Unable to get apps list
{{"controller"}, "namespace": "splunk", "name": "sh", "reconcileID": "1e31ce12-83a0-4862-b83b-cb3bc9c9353c", "name": "sh", "namespace": "splunk", "appSource": "esApps", "error": ""WebldentityErr: failed to retrieve credentials#n
caused by: SerializationError: failed to unmarshal error message\n\t
status code: 405, request id: \n
caused by: UnmarshalError: failed to unmarshal error message\n\t
00000000 3c 3f 78 6d 6c 20 76 65 72 73 69 6f 6e 3d 22 31 |<?xml version=\"1|\n
00000010 2e 30 22 20 65 6e 63 6f 64 69 6e 67 3d 22 55 54 |.0\" encoding=\"UT|\n
00000020 46 2d 38 22 3f 3e 0a 3c 45 72 72 6f 72 3e 3c 43 |F-8\"?>.<Error> <C|\n
00000030 6f 64 65 3e 4d 65 74 68 6f 64 4e 6f 74 41 6c 6c lode>MethodNotAll|\Hn
00000040 6f 77 65 64 3c 2f 43 6f 64 65 3e 3c 4d 65 73 73 lowed</Code><Mess|\n
00000050 61 67 65 3e 54 68 65 20 73 70 65 63 69 66 69 65 |age> The specifie|\n
00000060 64 20 6d 65 74 68 6f 64 20 69 73 20 6e 6f 74 20 Id method is not |\n
00000070 61 6c 6c 6f 77 65 64 20 61 67 61 69 6e 73 74 20 |allowed against |\n
00000080 74 68 69 73 20 72 65 73 6f 75 72 63 65 2e 3c 2f |this resource. </|\n
00000090 4d 65 73 73 61 67 65 3e 3c 4d 65 74 68 6f 64 3e |Message> <Method>I\n
[...]

PR #1218, opened 2023-08-16
Fix issue with retrieving apps from s3 when using IRSA

tnycum:
Seems to fix https://github.com/splunk/splunk-operator/issues/1217


vivekr-splunk (approved):

vivekr-splunk:
closing this as this change is already merged

PR #1223, opened 2023-08-22
Fix/deployment pvc

mvleandro:
This PR fixes a malformed deployment artifact that is being created when we specify a storageClassName for the persistencVolumeClaim on the Splunk operator.


vivekr-splunk (approved):

patrykw-splunk:
Hey, according to the current state of splunk-operator, changes from these PR are already there. 
In case otherwise we can reopen it.

PR #1240, opened 2023-10-05
Cspl-2489: GITHUG-1153 - Fix indexer deletion in upgrade scenario

vivekr-splunk:
This PR test GITHUT -1153 issue

addresses issue https://github.com/splunk/splunk-operator/issues/1151.
There are 2 changes:

Make sure that the fix of PR https://github.com/splunk/splunk-operator/pull/1041 is applied in the cluster-manager usecase
Fix the check on the image version. The format of the image field being like splunk/splunk-operator:8.2.0, it is needed to get the version that is after the colon. The previous check was only checking if the image was starting with 8 or 9, which is always false.

### DIFF
--- a/pkg/splunk/enterprise/indexercluster.go
+++ b/pkg/splunk/enterprise/indexercluster.go
@@ -185,7 +185,7 @@ func ApplyIndexerClusterManager(ctx context.Context, client splcommon.Controller
 		for _, owner := range v.GetOwnerReferences() {
 			if owner.UID == statefulSet.UID {
 				// get the pod image name
-				if v.Spec.Containers[0].Image != cr.Spec.Image {
+				if imageUpdatedTo9(v.Spec.Containers[0].Image, cr.Spec.Image) {
 					// image do not match that means its image upgrade
 					versionUpgrade = true
 					break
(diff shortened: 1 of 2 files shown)


gaurav-splunk (approved):

pkg/splunk/enterprise/indexercluster.go
```diff
@@ -1069,3 +1066,14 @@ func RetrieveCMSpec(ctx context.Context, client splcommon.ControllerClient, cr *
...
 	return "", nil
 }
+
+// Tells if there is an image migration from 8.x.x to 9.x.x
+func imageUpdatedTo9(previousImage string, currentImage string) bool {
+	// If there is no colon, version can't be detected
+	if !strings.Contains(previousImage, ":") || !strings.Contains(currentImage, ":") {
+		return false
+	}
+	previousVersion := strings.Split(previousImage, ":")[1]
```
akondur, line 1076: Can the string end with a `:` ? In that case this will seg fault? Maybe better to check length also unless the validation has been taken care of elsewhere.

vivekr-splunk: test case already covers this scenario

akondur (approved):

PR #1247, opened 2023-10-13
changes to support azure docker login 

vivekr-splunk:
Few changes were done to Azure pipeline, current credentials were not working for docker image pull/push
changed the workflow and github secrets


akondur (commented):
Nit: we need to remove comments here. Will remove it after the pipelines pass

gaurav-splunk (approved):

PR #1249, opened 2023-10-24
CSPL-2438 - Changes to add identifier for SOK deploymens

akondur:
Splunk Operator for Kubernetes(SOK) will add an environment variable `SPLUNK_RESOURCE_LIMITS_SOURCE` to all of its Splunk deployments in order to indicate the source for resource limits calculations

Note: If a customer introduces this environment variable unknowingly we have other checks as well in place to _hopefully_ catch it in Splunk. On a failure of the same, we can inform the customer of such a setting in resolve the issue.


vivekr-splunk (approved):
why is there change in library version. otherwise everything looks fine

akondur:
> why is there change in library version. otherwise everything looks fine

It was auto-generated with a build - maybe there are newer libraries for the golang version?

akondur:
Fixed in splunkd.

PR #1256, opened 2023-12-01
CSPL-2497: run nightly build only once a week

vivekr-splunk:


rlieberman-splunk:
This is done on internal gitlab pipelines regularly.

PR #1341, opened 2024-06-14
CSPL-2710: replace clair with trivy for security scanning

vivekr-splunk:
Replace clair scanner with trivy. Clair scanner is outdated.


akondur (approved):
Looks good!!

PR #1372, opened 2024-08-23
prevent masking of error messages in searchheadcluster.go

s1113950:
Throwing an error and then returning fmt.Errorf("custom err msg") is masking the underlying err and making it harder to debug why "custom err msg" happened. This change keeps the "custom err msg" but passes through the original error that happened!


akondur:
Implemented in [PR#1385](https://github.com/splunk/splunk-operator/pull/1385)

PR #1386, opened 2024-10-15
Update Examples.md

gjanders:
Adding additional context around the use of the external license manager settings as the previous section advises "you need to add licenseManagerRef only to the ClusterManager spec as follows:"

Which is not the case for external cluster managers


rlieberman-splunk:
This change is included in https://github.com/splunk/splunk-operator/pull/1575.

PR #1389, opened 2024-10-17
Splunk STS bug workaround

niraj-desai-sf:
Splunkd is not able to refresh STS token after 36hr and that is bring the pod in a wierd state.. Where it is reporting as down to CM but the Splunkd process is not down..

We tried restarting Splunkd on a cadence via the app here but that doesnt seem to be working.

Adding this liveliness probe update which will check for the same error and if the error exists it will just kill the pod ( something that we are doing today to manually fix this issue.


akondur:
@niraj-desai-sf Closing this as the issue is fixed in Splunk Enterprise 9.3.2.

Issue #1390, opened 2024-10-18
Splunk Operator: indexers don't start if search-heads still starting


yaroslav-nakonechnikov:
### Please select the type of request

Bug

### Tell us more

**Describe the request**
```
[yn@ip-100-65-8-59 /]$ kubectl get pods -n splunk-operator
NAME                                                  READY   STATUS    RESTARTS   AGE
splunk-43105-cluster-manager-0                        1/1     Running   0          19m
splunk-43105-license-manager-0                        1/1     Running   0          30m
splunk-c-43105-standalone-0                           1/1     Running   0          30m
splunk-e-43105-deployer-0                             0/1     Running   0          8m15s
splunk-e-43105-search-head-0                          0/1     Running   0          8m15s
splunk-e-43105-search-head-1                          0/1     Running   0          8m15s
splunk-e-43105-search-head-2                          /1     Running   0          8m15s
splunk-operator-controller-manager-58b545f67c-8rrhx   2/2     Running   0          31m
```

and then:
```
NAME                                                  READY   STATUS    RESTARTS   AGE
splunk-43105-cluster-manager-0                        1/1     Running   0          21m
splunk-43105-license-manager-0                        1/1     Running   0          32m
splunk-c-43105-standalone-0                           1/1     Running   0          32m
splunk-e-43105-deployer-0                             0/1     Running   0          11m
splunk-e-43105-search-head-0                          1/1     Running   0          11m
splunk-e-43105-search-head-1                          1/1     Running   0          11m
splunk-e-43105-search-head-2                          1/1     Running   0          11m
splunk-operator-controller-manager-58b545f67c-8rrhx   2/2     Running   0          34m
splunk-site3-43105-indexer-0                          0/1     Running   0          2m17s
splunk-site3-43105-indexer-1                          0/1     Running   0          2m17s
splunk-site3-43105-indexer-2                          0/1     Running   0          2m17s
```

this is unbelievable, and extremely strange that still, in 2.6.1 there is a dependency check between splunk search-heads and indexers!!!!

**Expected behavior**
Indexers should start without dependency of search-heads!

yaroslav-nakonechnikov:
it was already reported before #1260, and then there were 2 calls, where i described why logic with dependency is broken for kubernetes deployment. 

and now can test 2.6.1 and we still see, that part of platform can't be started just because of problematic logic.

vivekr-splunk:
@yaroslav-nakonechnikov we will get back to you with regard to this issue.

[...]

PR #1399, opened 2024-11-15
CSPL-3060 - Avoid upgrade path during deployment creation

akondur:
Addressing github issue #1390 

Splunk Operator applies the upgrade path business logic regardless of whether deployments are getting created or upgraded. While we have to maintain the order in which we upgrade the Splunk Enterprise deployments, we can avoid it during creation.


vivekr-splunk (approved):

akondur:
[Rest](https://github.com/splunk/splunk-operator/actions/runs/11858553307) of the integration tests.

PR #1405, opened 2024-11-21
CSPL-3041 - Fix run-tests.sh to pull enterprise image appropriately

akondur:
Fix run-tests.sh to pull enterprise image appropriately


patrykw-splunk:
stale PR, we can close it. In case of mistake we can reopen it

PR #1419, opened 2024-12-18
CSPL-3256 - Support to configure deployer spec in SHC CRD

akondur:
Currently, the SHC CRD deploys both the SHC as well as the deployer.

However, the deployer requires significantly less resources/different node affinity. The MR here adds support to explicitly add **resource spec** and **node affinity** config capability(for deployer) within the SHC CRD. The common spec config will apply by default and is overridden by the deployer spec for the deployer statefulset if configured.

Splunk Hardware spec [docs](https://docs.splunk.com/Documentation/Splunk/9.4.0/Capacity/Referencehardware#Reference_host_specification_for_single-instance_deployments).


akondur:
Integration test [run](https://github.com/splunk/splunk-operator/actions/runs/12404199533/job/34629038235?pr=1419) which tests the deployer spec resources here. PFA the log with the TC from the PR.

[1_int-tests (managercrcrud).txt](https://github.com/user-attachments/files/18199289/1_int-tests.managercrcrud.txt)

.github/workflows/int-test-workflow.yml
```diff
@@ -5,6 +5,7 @@ on:
       - develop
       - main
       - feature**
+      - CSPL_3256
```
rlieberman-splunk, line 8: Looks good! Make sure to clean this file up before merging.

rlieberman-splunk (approved):

pkg/splunk/enterprise/searchheadcluster.go
```diff
@@ -651,13 +652,41 @@ func getSearchHeadStatefulSet(ctx context.Context, client splcommon.ControllerCl
...
+// CSPL-3652 Configure deployer resources if configured
+// Use default otherwise
+// Make sure to set the resources ONLY for the deployer
+func setDeployerResources(cr *enterpriseApi.SearchHeadCluster, podTemplate *corev1.PodTemplateSpec) error {
+	// Break out if this is not a deployer
+	if !strings.Contains("deployer", podTemplate.Labels["app.kubernetes.io/name"]) {
+		return errors.New("not a deployer, skipping setting resources")
+	}
+	depRes := cr.Spec.DeployerResourceSpec
+	for i := range podTemplate.Spec.Containers {
```
vivekr-splunk, line 664: Ensure the deployer has default resource values, regardless of whether the customer provides them. We want to avoid using search head resource values for the deployer since it does not require as many resources.

test/testenv/verificationutils.go
```diff
@@ -641,6 +646,48 @@ func VerifyCPULimits(deployment *Deployment, ns string, podName string, expected
 	}, deployment.GetTimeout(), PollInterval).Should(gomega.Equal(true))
 }
 
+// VerifyResourceConstraints verifies value of CPU limits is as expected
```
vivekr-splunk, line 649: I am not saying we have to do this with this CR. We have to come up with different approach to manage non splunk configurations  Since our CR has configurations that are very diverse, we should consider splitting the CR into multiple, smaller CRDs or using sub-resources that reference each other. For example:

**Primary CR for Core Deployment:** Contains high-level configuration.
**Supplementary CRs:** Separate CRDs for storage, networking, or Splunk-specific configurations that reference the primary CR by labels or owner references.

**Example:**
`SearchHeadCluster` for core deployment.
`SplunkStorageConfig`, `SplunkNetworkingConfig`, etc., as separate CRDs linked to the main resource.
**Benefits:**

- **Separation of Concerns:** Each CR handles a specific part of configuration.
- **Easier Extensibility:** Add new configuration types without modifying the main CR.

vivekr-splunk (approved):

Igor-splunk (approved):

PR #1420, opened 2024-12-18
[CSPL-3298] Remove change splunk operator name step in integration test workflow

rlieberman-splunk:
This step is no longer needed, and to be consistent with the smoke test workflow, it can be removed.


akondur (approved):

patrykw-splunk (approved):
LGTM

PR #1434, opened 2025-02-04
Move pushing latest image from build test workflow to automated releasese workflow

rlieberman-splunk:
This change moves pushing the latest dockerhub tag from the build-test-push workflow to the automated-release workflow. The latest tag and the release tag should be the same image, so pushing both together is the best option instead of pushing two separate images.

### DIFF
--- a/.github/workflows/automated-release-workflow.yml
+++ b/.github/workflows/automated-release-workflow.yml
@@ -87,10 +87,12 @@ jobs:
     - name: Promote RC Image to Release
       run: |
         docker tag splunk/splunk-operator-rc:${{ github.event.inputs.release_version }}-RC splunk/splunk-operator:${{ github.event.inputs.operator_image_tag }}
+        docker tag splunk/splunk-operator:${{ github.event.inputs.operator_image_tag }} splunk/splunk-operator:latest
 
     - name: Push Release Image
       run: |
         docker push splunk/splunk-operator:${{ github.event.inputs.operator_image_tag }}
+        docker push splunk/splunk-operator:latest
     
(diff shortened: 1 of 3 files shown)


.github/workflows/automated-release-workflow.yml
```diff
@@ -87,10 +87,12 @@ jobs:
     - name: Promote RC Image to Release
       run: |
         docker tag splunk/splunk-operator-rc:${{ github.event.inputs.release_version }}-RC splunk/splunk-operator:${{ github.event.inputs.operator_image_tag }}
+        docker tag splunk/splunk-operator:${{ github.event.inputs.operator_image_tag }} splunk/splunk-operator:latest
```
vivekr-splunk, line 90: we should be good with this. but we should also check if we can use `docker buildx` so we can push all supported architecture same time. we do that in graviton workflow

vivekr-splunk (approved):

.github/workflows/automated-release-workflow.yml
rlieberman-splunk: The image will have already been built during the merge to main workflow https://github.com/splunk/splunk-operator/blob/main/.github/workflows/merge-develop-to-main-workflow.yml#L89, so this step is just retagging the existing image. I think this could be an enhancement for future releases where we build and push all architectures, but we currently don't provide multiple architectures on Docker Hub and I think that will be a larger effort that we should create a task for.

rlieberman-splunk: https://splunk.atlassian.net/browse/CSPL-3429

PR #1442, opened 2025-02-12
CSPL-3435 Fix bias_lang_migration.sh logic for monitoringConsole

patrykw-splunk:
This branch fixes bias_lang_migration.sh logic for handling monitoringConsole resource type


patrykw-splunk:
> Do we want to merge this straight into main? I think it should go to develop and if a customer needs it we can make a special build for them, but I'm not totally sure what the process is

my bad it should be merged into develop

PR #1452, opened 2025-02-24
CSPL-3426: Support tar.gz extension for apps

Igor-splunk:


patrykw-splunk (approved):
LGTM

rlieberman-splunk (approved):

rlieberman-splunk:
Can we add an integration test that uses a splunkbase app with the .tar.gz extension to actually test the download and install of it? Or did you test it manually?

Igor-splunk:
> Can we add an integration test that uses a splunkbase app with the .tar.gz extension to actually test the download and install of it? Or did you test it manually?

Yeah it works tgz and tar.gz are the exact same thing. tgz is just archaic because old dos floppy disks could only have 3 letters for extensions.

kasiakoziol (approved):

PR #1468, opened 2025-03-14
CSPL-3499: Fix Segmentation Fault in Splunk Operator Due to App Volume Name Change

rlieberman-splunk:
### Description

This pull request checks for an error returned from the getRemoteDataClientMgr() function to avoid a segmentation fault in an error case. A customer changed the name of an app source, and the status in the CR was not correctly updated, leading to this situation.

### Key Changes

* `remoteDataClientMgr, _ := getRemoteDataClientMgr()` -> `remoteDataClientMgr, err := getRemoteDataClientMgr()`: Handle the situation where the operator is not able to get the remote data client manager due to the app source spec and status being out of sync. In this situation, we will gracefully log and error statement and continue with the reconciliation instead of seg faulting.

* Move `downloadWorker.waiter.Add(1)`: A waiter should only be incremented when we actually call a go routine. Moving this incrementation will make sure that the operator does not stall in edge cases where errors are caught and the go routine is not started.

* Add a note in the AppFramework.md doc to alert users to this scenario.

### Testing and Verification

* `func TestScheduleDownloadsFailRemoteDataClientMgr(t *testing.T)`: Unit test added to test this scenario. It is based off of the existing TestScheduleDownloads() function, but with the spec and status out of sync to trigger this error case. This test was also ran with the develop code (not including code changes in this PR), and it did result in a segmentation fault that the customer saw, verifying that these changes fix the error handling.

### Related Issues

https://splunk.atlassian.net/browse/CSPL-3499

### PR Checklist

- [x] All tests pass locally.

### DIFF
--- a/pkg/splunk/enterprise/afwscheduler_test.go
+++ b/pkg/splunk/enterprise/afwscheduler_test.go
@@ -1828,8 +1828,6 @@ func TestScheduleDownloads(t *testing.T) {
 	downloadPhaseWaiter.Wait()
 }
 
-// CSPL-3490: Fix Segmentation Fault in Splunk Operator Due to App Volume Name Change
-// A customer changed the name of an app in the app source, which caused the operator to crash.
 // This test case is to verify that the operator does not crash when the app name is changed.
 func TestScheduleDownloadsFailRemoteDataClientMgr(t *testing.T) {
 	ctx := context.TODO()
--- a/pkg/splunk/enterprise/afwscheduler_test.go
+++ b/pkg/splunk/enterprise/afwscheduler_test.go
@@ -1828,6 +1828,126 @@ func TestScheduleDownloads(t *testing.T) {
 	downloadPhaseWaiter.Wait()
 }
 
+// CSPL-3490: Fix Segmentation Fault in Splunk Operator Due to App Volume Name Change
+// A customer changed the name of an app in the app source, which caused the operator to crash.
+// This test case is to verify that the operator does not crash when the app name is changed.
+func TestScheduleDownloadsFailRemoteDataClientMgr(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/splunk/enterprise/afwscheduler_test.go
```diff
@@ -1828,6 +1828,126 @@ func TestScheduleDownloads(t *testing.T) {
 	downloadPhaseWaiter.Wait()
 }
 
+// CSPL-3490: Fix Segmentation Fault in Splunk Operator Due to App Volume Name Change
```
patrykw-splunk, line 1831: I think that I would just leave the last line of comment.

rlieberman-splunk: @vivekr-splunk suggested adding the CSPL number in case we want to search for it in the code in the future. I could add it to the code instead of the test if you think that makes more sense?

patrykw-splunk: Hmmm I think that added CSPL in commit/branch for this particular change should be enough. In the end CSPL tags are only helpful for us. 
But all good, just leave it and we can dicuss someday a scenario like that

aalokmehta:
In util.go, I am wondering if `appSrcSpc` should be changed to `appSourceSpec` as part of this PR, so that troubleshooting is easier.

rlieberman-splunk:
> In pkg/splunk/enterprise/util.go#L538, I am wondering if `appSrcSpc` should be changed to `appSourceSpec` as part of this PR, so that troubleshooting is easier.

I updated that line in the newest commit.

vivekr-splunk (approved):

patrykw-splunk (approved):

PR #1510, opened 2025-05-21
Feature/secrets with complexity

araman-m:
### Description

this PR make sure secrets are generated with crypto rand function for optimal randomness and that they at least include a lower, upper case + decimal + special characters , which is usually a requirement for regulated context like PCI (and a good practise in general)

### Key Changes

* replace math/rand with crypto/rand 
* Add Generate Secrets With complexity requirements 

files modified pkg/splunk/common/names.go -> add more const variables for lower case, upper case, decimal and special characters
pkg/splunk/util/util.go pkg/splunk/util/util_test.go , pkg/splunk/util/secrets.go ,pkg/splunk/util/secrets_test.go 

add GenerateSecretPartWithComplexity function , which choose m times a character for the SecretBytes passed as argument at a random position in the byte array (but checking it is not yet used position)

add GenerateSecretWithComplexity(n int, minlower int, minupper int, mindecimal int, minspecial int) fucntion
which for a n size secrets will :
- check the added constraints are possible 
- call GenerateSecretPartWithComplexity for all the 4 cases
- complete the remaining with a random characters from the extended allowed characters

GenerateSecret function is still there but only used for HEC
it is now using the crypto rand function also (you cant have different rand in the same file and this is more secure anyway to use the crypto version)

updated the code to call the GenerateSecretWithComplexity with 24 lentgh (as before) and 1 complexity for each type

### Testing and Verification

updated tests

added a test that verify we produce a error if ask more complexity than secrets length

manual tests
- run the code as standalone function then once good within operator
- check secrets are now complex (by deleting older secrets) -> OK

### Related Issues

Yes, customer issue discussed internally

### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/pkg/splunk/common/names.go
+++ b/pkg/splunk/common/names.go
@@ -32,6 +32,15 @@ const (
 	// SecretBytes used to generate Splunk secrets
 	SecretBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
 
+	// SecretBytes with complexity used to generate Splunk secrets with complexity
+	SecretBytesLower   = "abcdefghijklmnopqrstuvwxyz"
+	SecretBytesUpper   = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	SecretBytesDecimal = "0123456789"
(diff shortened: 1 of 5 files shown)


rlieberman-splunk:
Closing as these changes are being handled in https://github.com/splunk/splunk-operator/pull/1511

PR #1549, opened 2025-07-25
[CSPL-3912] Allow Custom Probe Scripts

rlieberman-splunk:
### Description

This PR allows edits to the splunk-<namespace>-probe-configmap configmap so that users can use custom probe scripts for CRs. Previously, the operator would overwrite the data in the configmap on every reconciliation, so users were not able to edit probe scripts.

### Key Changes

- docs/Examples.md: Added documentation on how to create or edit probe scripts.
- pkg/splunk/enterprise/configuration.go: Updated the logic to check if the configmap exists. If it does, use the existing one. If it does not, create one with the default probe scripts. This will be a transparent change for users who continue to use the default scripts.
- *_test.go files: Updated existing unit test to work with the logic changes.

### Testing and Verification

* Tested on an eks cluster. 
    1. Configmap created before deploying a c3 instance: existing configmap data correctly mounted on the pods in the namespace.
    2. Configmap editing after it was created: new configmap data correctly mounted to the pods in the namespace.
    3. Configmap not created and not edited: default probe data correctly mounted to the pods in the namespace.

### Related Issues

- https://splunk.atlassian.net/browse/CSPL-3912

### PR Checklist

- [X] All tests pass locally.


patrykw-splunk (approved):
LGTM!

PR #1568, opened 2025-08-26
CSPL-3974: improve region parsing and enforce latest version downloads

vivekr-splunk:
This PR fixes region parsing in the AWS S3 client and ensures downloads always fetch the **latest version** of objects from S3. It also improves input validation, error handling, and cleanup behavior in the download workflow.

---

### Key Changes

* **`awss3client.go`**

  * Updated `GetRegion` to correctly parse full S3 URLs and broadened regex for both path-style and virtual-hosted endpoints.
  * Fixed `NewAWSS3Client` to properly set `Endpoint` and handle region derivation.
  * Removed `IfMatch` in `DownloadApp` to avoid stale ETag errors.
  * Added optional `HeadObject` preflight check to log differences when provided ETag does not match current.
  * Added input validation for empty remote/local file paths.
  * Fixed cleanup logic to remove partially written **local files only**.

* **Tests**

  * Strengthened `TestNewAWSS3Client` to cover valid, backward-compatibility, and invalid scenarios.
  * Updated `TestAWSDownloadAppShouldNotFail`, `TestAWSDownloadAppShouldFail`, and `TestAWSDownloadAppIgnoresProvidedETagAndGetsLatest` to validate latest-version behavior and error handling.

---

### Testing and Verification

* Ran unit tests locally:

  * ✅ `TestNewAWSS3Client` passes for valid endpoints, fails for invalid ones.
  * ✅ `TestAWSDownloadAppShouldNotFail` confirms successful downloads.
  * ✅ `TestAWSDownloadAppShouldFail` removes partial files on failure.
  * ✅ `TestAWSDownloadAppIgnoresProvidedETagAndGetsLatest` validates latest object is always downloaded.
* Verified log messages for mismatched ETags and error conditions.
```
Validation results (S3 versioned bucket):

Before:
  description = MyTestApp v1

After pushing new content to same key and triggering App Framework:
  description = MyTestApp v2
  previous version preserved at default.old.<timestamp>

No 412/PreconditionFailed observed in operator logs.
Operator consistently fetches latest object version for the key.
```
---

### Related Issues

* Fix region parsing and enforce latest version download for S3 apps
* GitHub: Internal bug report for failing `TestAWSDownloadAppShouldNotFail`

---

### PR Checklist

* [x] Code changes adhere to the project's coding standards.
* [x] Relevant unit and integration tests are included.
* [x] Documentation has been updated accordingly.
* [x] All tests pass locally.
* [x] The PR description follows the project's guidelines.

### DIFF
--- a/pkg/splunk/client/awss3client.go
+++ b/pkg/splunk/client/awss3client.go
@@ -68,30 +68,30 @@ var regionRegex = `(?i)(^|\.)(s3)[\.-]([a-z0-9-]+)\.amazonaws\.com$`
 
 // GetRegion extracts the region from the endpoint field
 func GetRegion(ctx context.Context, endpoint string, region *string) error {
-    var host string
+	var host string
 
-    // If endpoint looks like a URL, extract the hostname; otherwise use raw string.
-    if u, err := url.Parse(endpoint); err == nil && u.Hostname() != "" {
-        host = u.Hostname()
-    } else {
-        // tolerate raw host (with or without scheme)
-        host = endpoint
-        // strip a possible leading scheme manually if present
-        if strings.HasPrefix(host, "http://") || strings.HasPrefix(host, "https://") {
-            if u2, err2 := url.Parse(host); err2 == nil && u2.Hostname() != "" {
-                host = u2.Hostname()
-            }
-        }
-    }
+	// If endpoint looks like a URL, extract the hostname; otherwise use raw string.
+	if u, err := url.Parse(endpoint); err == nil && u.Hostname() != "" {
+		host = u.Hostname()
+	} else {
+		// tolerate raw host (with or without scheme)
+		host = endpoint
+		// strip a possible leading scheme manually if present
+		if strings.HasPrefix(host, "http://") || strings.HasPrefix(host, "https://") {
+			if u2, err2 := url.Parse(host); err2 == nil && u2.Hostname() != "" {
+				host = u2.Hostname()
+			}
+		}
+	}
 
-    pattern := regexp.MustCompile(regionRegex)
-    m := pattern.FindStringSubmatch(host)
-    if len(m) >= 4 {
-        // capture group 3 is the region
-        *region = m[3]
-        return nil
-    }
-    return fmt.Errorf("unable to extract region from the endpoint: %q (host: %q)", endpoint, host)
+	pattern := regexp.MustCompile(regionRegex)
+	m := pattern.FindStringSubmatch(host)
+	if len(m) >= 4 {
+		// capture group 3 is the region
+		*region = m[3]
+		return nil
+	}
+	return fmt.Errorf("unable to extract region from the endpoint: %q (host: %q)", endpoint, host)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/splunk/client/awss3client.go
```diff
@@ -62,18 +64,34 @@ type AWSS3Client struct {
...
+
+    // If endpoint looks like a URL, extract the hostname; otherwise use raw string.
+    if u, err := url.Parse(endpoint); err == nil && u.Hostname() != "" {
+        host = u.Hostname()
+    } else {
+        // tolerate raw host (with or without scheme)
+        host = endpoint
+        // strip a possible leading scheme manually if present
+        if strings.HasPrefix(host, "http://") || strings.HasPrefix(host, "https://") {
+            if u2, err2 := url.Parse(host); err2 == nil && u2.Hostname() != "" {
```
rlieberman-splunk, line 81: I'm confused how this line is different from line 74. If line 74 is false, then we set `host = endpoint`. It looks like line 81 runs the same command with `host` instead of `endpoint`, but they are set to the same thing.

pkg/splunk/client/awss3client.go
```diff
@@ -62,18 +64,34 @@ type AWSS3Client struct {
...
+                host = u2.Hostname()
+            }
+        }
+    }
+
+    pattern := regexp.MustCompile(regionRegex)
+    m := pattern.FindStringSubmatch(host)
+    if len(m) >= 4 {
+        // capture group 3 is the region
+        *region = m[3]
```
rlieberman-splunk, line 91: Why is this getting index 3 instead of index 1 like in the original?

vivekr-splunk: Depending on how many capturing groups we’ve got, `FindStringSubmatch` returns:

m[0]: full match
m[1]: first capture group
m[2]: second capture group
m[3]: third capture group
If our regex has multiple groups (like one for prefix, one for separators, one for the region), the region won’t always be in index 1 anymore. That’s why this version uses m[3] ,  because the third group in the pattern corresponds to the actual region.

pkg/splunk/client/awss3client.go
vivekr-splunk: **First branch (line 74):**
If endpoint is a valid URL (https://s3.us-west-2.amazonaws.com), it grabs the hostname (s3.us-west-2.amazonaws.com) directly.
**Else branch (line 81):**
This is for when endpoint is not parseable as a full URL. Example: maybe the input is just "s3.us-west-2.amazonaws.com" or "http://s3.us-west-2.amazonaws.com" but without being parsed cleanly the first time.

First it assigns host = endpoint (raw string).

Then it does a quick scheme check (http:// or https://) and tries parsing again, in case the string was close to a URL but missing structure.

So while it looks like the same parsing, the fallback is trying to recover from malformed or odd inputs where the first url.Parse didn’t yield a hostname.

rlieberman-splunk (approved):

patrykw-splunk (approved):
LGTM!

Issue #1180, opened 2023-06-26
App Framework: Add scope:cluster to Standalone CRD to deploy to etc/deployment-apps


ddance:
### Please select the type of request

Feature Request

### Tell us more

**Describe the request**
Currently the Standalone CRD only supports the scope of local.  Please can it also support the scope of cluster and have the target set to $SPLUNK_HOME/etc/deployment-apps

Having the ability to output to deployment-apps from the app framework would unlock the Deployment Server usecase for this CRD.

**Expected behavior**
The scope of cluster in the Standalone CRD to function and output to the deployment-apps location, with apps deposited here we can make use of this CRD as a deployment server.

**Proposed changes(optional)**
- Addition of scope:cluster to the Standalone CRD

PrymalInstynct:
I 100% agree, and do not understand how this has not been added as a capability yet.

PR #1589, opened 2025-09-29
CSPL-3532: Allowing scope: cluster for standalone CRDs as per GHI request

gabrielm-splunk:
### Description

This PR allows scope: cluster on standalone CRDs, and thus allowing to be installed in the /etc/deployment-apps dir

### Key Changes

- Modifying standalone CRDs to allow scope: cluster (as well as scope: local, where behavior is unchanged) 
    - This will now allow us to install to /etc/deployment-apps 

### Testing and Verification

Ran operator and automation locally, however still in the process of manually verifying the changes. Just creating this PR to start getting review on work.

### Related Issues

Jira: https://splunk.atlassian.net/browse/CSPL-3532
GHI: https://github.com/splunk/splunk-operator/issues/1180

### PR Checklist

- [x] All tests pass locally.

### DIFF
--- a/.biased_lang_exclude
+++ b/.biased_lang_exclude
@@ -75,6 +75,7 @@ api/v4/zz_generated.deepcopy.go
 api/v4/common_types.go
 #Below should be reviewed once CRDs have no bias lang
 tools/bias_lang_migration.sh
+go.mod
 go.sum
 .github/
 .circleci/


kasiakoziol:
> It would be good to add a test case to the appframeworksS1 test suite that includes an app deployed to the cluster scope, or possibly even add to one of the existing test cases with the cluster scope for a new app.
> 
> There are also some existing unit tests in standalone_test.go that are related to appframework that we should add to as well to test both the local and cluster scoped apps.

I agree. We need to add new test cases. Also, I believe that automated tests are not enough and some portion of manual tests is required for both local and cluster scopes.

hampusstrom:
I suggest including documentation similar to #1603 for how to handle serverclasses and the new clusterscope for etc/deployment-apps together.

PR #1644, opened 2025-12-15
Amadeus patches: Advanced Statefulset scaling + Support for updating VolumeClaimTemplates

ductrung-nguyen:
There are several different changes in this PR, in some dedicated commits.
As they have some overlapped parts, it is difficult to split them into different PRs.

**Notes**: Why do we mainly use annotations to control some behavior of the operator instead of updating the CR fileds? Quick answer, to maintain the patches of the CRDs are nightmare for us. Moreover, the CRD deployment is out of our control, and we often need to wait for them to be deployed. This process takes time and slow down our development process.

---
# 1.0. Prefix-based metadata filtering 
### Description

This PR implements prefix-based metadata filtering to give users control over how labels and annotations propagate from the Custom Resource (CR) to the underlying StatefulSet and Pods.

Previously, almost all metadata was copied everywhere. Now you can use specific prefixes to target where you want your labels to go:

*   **`sts-only.*`**: Keys starting with this are added *only* to the StatefulSet metadata (and the prefix is stripped!). They won't appear on the Pod Template, so adding them won't trigger a pod rollout.
*   **`pod-only.*`**: Keys are added *only* to the Pod Template (and excluded from the StatefulSet metadata).
*   **System prefixes**: We now automatically exclude `kubectl.kubernetes.io/*` and `operator.splunk.com/*` prefixes to prevent clutter and conflicts.

Regular labels without these prefixes work exactly as before—propagatating to both.

### Key Changes

*   **`pkg/splunk/common/util.go`**: Added the core filtering logic. It checks for `sts-only.` and `pod-only.` prefixes.
*   **`pkg/splunk/common/statefulset_util.go`**: Implemented `SyncParentMetaToStatefulSet` which handles the transformation (stripping `sts-only.`) and syncing logic.
*   **`pkg/splunk/enterprise/configuration.go`**: Updated the main reconciliation loop to use these new syncing functions instead of the old blind copy.
*   **`pkg/splunk/splkcontroller/statefulset.go`**: Added logic to detect if *only* StatefulSet metadata changed (so we can update the STS object without touching the Pod specs).

### Testing and Verification

Many unit tests are implemented.

For E2E validation:
1.  Update the existing Splunk CR.
2.  Add a label `sts-only.my-tag: true`.
    *   **Verifed**: The StatefulSet has `my-tag: true`.
    *   **Verifed**: The Pods do *not* have `my-tag` (and thus no rollout if added later).
3.  Add a label `pod-only.scrapper: true`.
    *   **Verified**: The Pods have `pod-only.scrapper: true`.
    *   **Verified**: The StatefulSet does *not* have this label.
4.  Add a generic label `environment: prod`.
    *   **Verified**: Both STS and Pods have `environment: prod`.

### Related Issues
https://github.com/splunk/splunk-operator/issues/1652

### PR Checklist

- [x] All tests pass locally.

# 2.0. Flexible scaling
### Description

Makes StatefulSet scaling more flexible by fixing two annoying behaviors:

1. **Scale-down no longer blocks on unhealthy pods** - Previously the operator waited for all pods to be ready before scaling down, which doesn't make sense since removing pods doesn't need healthy pods. Now it just checks if we have more replicas than desired and proceeds.

2. **Scale-up has configurable timeout** - Instead of waiting forever for pods to be ready, users can now set a timeout via annotation. After the timeout, scale-up proceeds anyway (with a warning logged). Default is still "wait forever" to maintain backward compat.

Also includes some defensive improvements to scale-down:
- Better bounds checking in `PrepareScaleDown()` to handle edge cases like manually deleted pods
- Fallback mechanism to query Cluster Manager directly when CR status is stale
- Empty peer status now treated as "decommission complete" instead of blocking

### Key Changes
**New annotations:**
- `operator.splunk.com/scale-up-ready-wait-timeout`: time of waiting for all pods to be ready before scaling up
- `operator.splunk.com/scale-up-wait-started`: **internal** tracking the starting time of scale up wait

**Behavior at a glance** of the new annotation `operator.splunk.com/scale-up-ready-wait-timeout`:

| Annotation value | Result |
| --- | --- |
| (missing) | proceed immediately |
| `"invalid"` | proceed immediately |
| `"0"` or `"0s"` | proceed immediately |
| `"5m"` | wait up to 5 minutes |
| `"-1"` or any negative | wait forever |

### Testing and Verification

**Unit tests added:**
Please check the code for more information.

**Manual E2E testing:**
Was done by scaling up and down cluster often and quickly.

### Related Issues

https://github.com/splunk/splunk-operator/issues/1646

### PR Checklist

- [x] All tests pass locally.

---

# 3.0. Support updating CPU or changing VCT with CPU-awareness option
### Description

This PR introduces 3 major enhancements for better scaling and updates when changing the CPU or Volume Claim Template:

1.  **CPU-aware scaling**: Keeps your total cluster CPU capacity constant when you change Pod sizes. It automatically adjusts replica counts up or down to match the new Pod specs.
2.  **Parallel pod updates**: Speeds up rolling restarts by updating multiple pods at once (e.g., 25%).
3.  **Automatic VolumeClaimTemplate Updates**: Safely handles changes to `VolumeClaimTemplates` (which are normally immutable in K8s).
    *   **Storage Expansion**: Automatically expands PVCs.
    *   **Immutable Changes**: Safely recreates the StatefulSet (orphaning pods) or manages migration for changes like StorageClass or access modes.

We also handles the safe ordering of scaling operations (scale-up first, then scale-down, or vice versa) to ensure capacity is never lost.

This change is fully backward compatible.

### Key Changes

**New Annotations:**
- `operator.splunk.com/preserve-total-cpu` - enables CPU-preserving replica adjustment
- `operator.splunk.com/parallel-pod-updates` - controls concurrent pod updates (< 1.0 = percentage, >= 1.0 = absolute count)

**Behavior:**
- Scale-up: happens immediately when CPU decreases per pod (more pods needed)
- Scale-down: happens gradually with safety checks (waits for new-spec pods to be ready, respects CPU bounds)
- Parallel updates: defaults to 1 pod at a time, configurable via annotation

### Testing and Verification

- Added unit tests for all new helper functions
- Added integration-style tests for CPU-aware scaling scenarios
- Added tests for percentage and absolute parallel update modes
- Existing tests updated to reflect new scale-down prioritization behavior
...

### Related Issues
https://github.com/splunk/splunk-operator/issues/1645
https://github.com/splunk/splunk-operator/issues/1647

### PR Checklist

- [x] All tests pass locally.

---
# 4.0 Supporting improving startup performance on large persistent volumes by configuring `fsGroupChangePolicy`

### Description

This PR adds support for configuring `fsGroupChangePolicy` on Splunk pod security contexts to improve startup performance on large persistent volumes.

When pods with large PVs start up, Kubernetes can spend a long time recursively changing ownership/permissions on all files. By setting the policy to `OnRootMismatch`, we skip this expensive operation if the root dir already has the correct fsGroup - which is the case for most restarts.

### Key Changes

- **api/v4/common_types.go**: Added `FSGroupChangePolicy` field to the common Spec struct. Users can now set this in their CR spec with validation for `Always` or `OnRootMismatch` values.

- **pkg/splunk/enterprise/configuration.go**: Added `getFSGroupChangePolicy()` helper that implements the precedence logic (annotation > spec > default). The default is now `OnRootMismatch` for better out-of-the-box perf.

- **pkg/splunk/splkcontroller/statefulset.go**: Added the `FSGroupChangePolicyAnnotation` constant for the `operator.splunk.com/fs-group-change-policy` annotation.

- **docs/CustomResources.md**: Added docs explaining the feature, configuration options (spec field vs annotation), and the precedence rules.

- **CRD yaml files**: Updated all CRD definitions to include the new field (generated from `make generate` and command `make manifest`

### Testing and Verification

- Added unit tests for `getFSGroupChangePolicy()` covering all precedence scenarios
- Tested with both spec field and annotation configurations
- Verified invalid annotation values fall back gracefully with warning log
- Manual testing on cluster with large PVs showed significant startup time improvement

### Related Issues

https://github.com/splunk/splunk-operator/issues/1648

### PR Checklist

- [x] All tests pass locally.

### DIFF
--- a/api/v4/common_types.go
+++ b/api/v4/common_types.go
@@ -112,6 +112,14 @@ type Spec struct {
 
 	// TopologySpreadConstraint https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// FSGroupChangePolicy defines the policy for changing ownership and permission of the volume
+	// before being exposed inside the Pod. Valid values are "Always" and "OnRootMismatch".
+	// Default is "OnRootMismatch" for improved performance.
+	// Can be overridden by the operator.splunk.com/fs-group-change-policy annotation.
+	// +kubebuilder:validation:Enum=Always;OnRootMismatch
+	// +optional
+	FSGroupChangePolicy *corev1.PodFSGroupChangePolicy `json:"fsGroupChangePolicy,omitempty"`
 }
 
 // Phase is used to represent the current phase of a custom resource
--- a/docs/CustomResources.md
+++ b/docs/CustomResources.md
@@ -15,6 +15,7 @@ you can use to manage Splunk Enterprise deployments in your Kubernetes cluster.
   - [Metadata Parameters](#metadata-parameters)
   - [Common Spec Parameters for All Resources](#common-spec-parameters-for-all-resources)
   - [Common Spec Parameters for Splunk Enterprise Resources](#common-spec-parameters-for-splunk-enterprise-resources)
+    - [FSGroup Change Policy](#fsgroup-change-policy)
   - [LicenseManager Resource Spec Parameters](#licensemanager-resource-spec-parameters)
   - [Standalone Resource Spec Parameters](#standalone-resource-spec-parameters)
   - [SearchHeadCluster Resource Spec Parameters](#searchheadcluster-resource-spec-parameters)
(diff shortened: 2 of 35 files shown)


vivekr-splunk (changes requested):
Thanks for putting the four pieces of this stack in the PR description. I went through this from the operator/product point of view, and I do not think we should merge this combined PR as-is.

The main concern is that this PR establishes a fairly large StatefulSet-management contract before we have aligned the base design. In particular, it introduces several new operator-managed annotations for rollout and transition behavior (`operator.splunk.com/scale-up-ready-wait-timeout`, `operator.splunk.com/preserve-total-cpu`, `operator.splunk.com/parallel-pod-updates`, `operator.splunk.com/fs-group-change-policy`, and the unified transition annotations), while we are currently using #1710 to spike how we want to manage StatefulSet lifecycle, restart/reload ordering, and rolling behavior. That spike is shaping the user-facing contract through CR spec in `rollingUpdateConfig` rather than adding more annotations.

Because of that overlap, I think it would be safer to align on the `#1710` direction first and then layer smaller pieces from here on top where they still add value. Would that direction solve the underlying rollout/update problem you are trying to address here?

A second concern is that this PR also changes the default `fsGroupChangePolicy` to `OnRootMismatch` for all workloads. That may be a useful optimization, but it is still a product-level behavior change and seems safer to review independently than to bundle inside a large StatefulSet-management redesign.

My suggestion would be:
1. align first on the base StatefulSet behavior in `#1710`
2. then bring over the parts that remain needed as smaller follow-up PRs

From what I can see today, the metadata-filtering portion overlaps the direction already being discussed in `#1720`, and the `fsGroupChangePolicy` support also looks like it could be evaluated on its own once the lifecycle/update contract is settled.

PR #1688, opened 2026-02-05
Cluster controller first iteration

DmytroPI-dev:
### Description

Cluster controller for CNPG first iteration

### Key Changes

Drafted API and controller for Cluster controller

### Testing and Verification

Manual tests only

### Related Issues

JIRA: CPI-1883

### PR Checklist

- [ ] All tests pass locally.


api/v4/cluster_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// This field is IMMUTABLE after creation.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="class is immutable"
+	Class string `json:"class"`
+
+	// Storage overrides the storage size from ClusterClass.
+	// Example: "5Gi"
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == null || oldSelf == null || quantity(self).compareTo(quantity(oldSelf)) >= 0",message="storage size can only be increased"
```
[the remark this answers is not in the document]
DmytroPI-dev: Storage is not downgradable, but good point, will check, did I understand you correctly, @limak9182 , that you are describing a situation when user created a default cluster, and then updated spec?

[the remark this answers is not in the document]
mploski: I dont this is the way it works. You put || between rules so it means that either of this rule succeed the whole validation pass i.e:

  - self refers to the current/new value being validated
  - oldSelf refers to the previous value (when updating an existing resource)
  - self == null || oldSelf == null - If either value is null, the validation passes (this allows creation and 
  deletion)
  - quantity(self).compareTo(quantity(oldSelf)) >= 0 - If both values exist, convert them to Kubernetes 
  quantities and compare them. The new value must be greater than or equal to the old value.

internal/controller/cluster_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+		BeforeEach(func() {
+			By("creating the custom resource for the Kind Cluster")
+			err := k8sClient.Get(ctx, typeNamespacedName, cluster)
+			if err != nil && errors.IsNotFound(err) {
+				resource := &enterprisev4.Cluster{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      resourceName,
+						Namespace: "default",
+					},
+					// TODO(user): Specify other spec details if needed.
```
mploski: do you plan to fills this comments?

internal/controller/database_cluster_controller.go
mploski: is it still Database? Shouldnt be Cluster?

api/v4/cluster_types.go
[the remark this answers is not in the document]
DmytroPI-dev: Will update the logic, you are correct.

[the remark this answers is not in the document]
DmytroPI-dev: And will update for PG version as well

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,430 @@
...
+		resultConfig.PostgresVersion = clusterClassDefaulfConfig.PostgresVersion
+	}
+	if resultConfig.Resources == nil {
+		resultConfig.Resources = clusterClassDefaulfConfig.Resources
+	}
+	if resultConfig.Storage == nil {
+		resultConfig.Storage = clusterClassDefaulfConfig.Storage
+	}
+	if len(resultConfig.PostgreSQLConfig) == 0 {
+		resultConfig.PostgreSQLConfig = clusterClassDefaulfConfig.PostgreSQLConfig
```
mploski: nil: We should consider initializing clusterClassDefaultConfig.PostgreSQLConfig when it is nil. While nil is currently accepted by CNPG, this becomes risky if we later decide to modify the value directly in the controller, for example:

cnpgCluster.Spec.PostgresConfiguration.Parameters["max_connections"] = "100"

This would cause a panic, and the attribute would also be omitted from kubectl describe output of our CR. Initializing it to an empty map ({}) would be safer.

Alternatively, we can set a default at the class level so the parameter is always initialized to an empty map:

// +kubebuilder:default={}
same goes for pgHBA

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,430 @@
...
+
+// defineNewCNPGCluster defines a new CNPG Cluster resource based on the merged configuration.
+func (r *ClusterReconciler) defineNewCNPGCluster(
+	clusterClass *enterprisev4.ClusterClass,
+	cluster *enterprisev4.Cluster,
+	mergedConfig *enterprisev4.ClusterSpec,
+) (*cnpgv1.Cluster, error) {
+
+	// Validate that required fields are present in the merged configuration before creating the CNPG Cluster.
+	if err := validateClusterConfig(mergedConfig, clusterClass); err != nil {
```
mploski: do we need this? why not enforce this on openAPI level i.e put a comment that the attribute is mandatory? It feels a bit redundant + with every change we will need to update this function

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,430 @@
...
+	clusterClass *enterprisev4.ClusterClass,
+	cnpgCluster *cnpgv1.Cluster,
+	mergedConfig *enterprisev4.ClusterSpec,
+) (bool, error) {
+
+	// Validate that required fields are present in the merged configuration before updating the CNPG Cluster.
+	if err := validateClusterConfig(mergedConfig, clusterClass); err != nil {
+		return false, err
+	}
+	// Validate that storage size is not decreased
```
mploski: this is already checked by CEL

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,430 @@
...
+				Owner:    "postgres",
+			},
+		},
+		StorageConfiguration: cnpgv1.StorageConfiguration{
+			Size: mergedConfig.Storage.String(),
+		},
+		Resources: resources,
+	}
+
+	if !equality.Semantic.DeepEqual(cnpgCluster.Spec, desiredState) {
```
mploski: can't we simply check if previous merged cluster config == current merged config?  We can just fetch the current cluster state and compare with mergedConfig. If there is any difference then initiate clusterSpec. In that case we dont even need separate function for a new cluster (defineNewCNPGCluster) vs updating cluster as at the end the process is exactly the same

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	mergedConfig := r.getMergedConfig(clusterClass, cluster)
+
+	// 4. Get or create the CNPG Cluster
+	cnpgCluster := &cnpgv1.Cluster{}
+	err := r.Get(ctx, types.NamespacedName{
+		Name:      cluster.Name,
+		Namespace: cluster.Namespace,
+	}, cnpgCluster)
+
+	isNewCluster := false
```
mploski: Right now, your Reconcile method is doing a manual branching using isNewCluster but we don't need it:

```
	// Merge ClusterClass spec into Cluster spec
	mergedConfig := r.getMergedConfig(clusterClass, cluster)

	// Build desired CNPG cluster object
	desiredCNPG := r.buildCNPGCluster(cluster, mergedConfig)

	// Get the existing CNPG cluster
	existingCNPG := &cnpgv1.Cluster{}
	err := r.Get(ctx, types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, existingCNPG)

	// Switch-based handling for create/update/no-op
	switch {
	case apierrors.IsNotFound(err):
		// Cluster doesn't exist → create
		if err := r.Create(ctx, desiredCNPG); err != nil {
			logger.Error(err, "Failed to create CNPG Cluster")
			return ctrl.Result{}, err
		}
     ... add status update, logger whatever
	case err != nil:
		// Some other error → return
		logger.Error(err, "Failed to get CNPG Cluster")
        ... update status 
		return ctrl.Result{}, err
	default:
		// Cluster exists → check if spec differs
		if !equality.Semantic.DeepEqual(existingCNPG.Spec, desiredCNPG.Spec) {
			existingCNPG.Spec = desiredCNPG.Spec
			if err := r.Update(ctx, existingCNPG); err != nil {
				logger.Error(err, "Failed to update CNPG Cluster")
				return ctrl.Result{}, err
			}
			logger.Info("CNPG Cluster updated", "name", existingCNPG.Name)
		} 
	}
```

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil && apierrors.IsNotFound(err) {
+		// Create new CNPG Cluster
+		cnpgCluster = r.buildCNPGCluster(cluster, mergedConfig)
+		if err := r.Create(ctx, cnpgCluster); err != nil {
+			logger.Error(err, "Failed to create CNPG Cluster")
```
mploski: we should update our cr status if we have a failure

DmytroPI-dev:
I have read the CLA Document and I hereby sign the CLA

PROJECT
```diff
@@ -131,4 +131,13 @@ resources:
   kind: DatabaseClass
   path: github.com/splunk/splunk-operator/api/v4
   version: v4
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: splunk.com
+  group: enterprise
+  kind: Cluster
```
mploski: Cluster or PostgresCluster?

DmytroPI-dev: Looks like a glitch, I have PostgresCluster in my branch after rebase

internal/controller/postgrescluster_controller.go
```diff
@@ -0,0 +1,306 @@
...
+
+// syncStatusAndResult is a helper function to handle the final steps of reconciliation, including syncing status and determining whether to requeue based on the CNPG Cluster state and any errors that occurred during reconciliation.
+func (r *PostgresClusterReconciler) syncStatusAndResult(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster, cnpgCluster *cnpgv1.Cluster, err error) (ctrl.Result, error) {
+	// Check if there was an error during sync, then return the original error.
+	if syncErr := r.syncStatus(ctx, postgresCluster, cnpgCluster, err); syncErr != nil {
+		// This joins the original logic error with the status update error
+		return ctrl.Result{}, errors.Join(err, fmt.Errorf("failed to sync status: %w", syncErr))
+	}
+	// 2. If the main logic failed, return that error now.
+	if err != nil {
```
mploski: as discussed yesterday this function should not be responsible for anything else than setting  phase and conditions. Lets move deciding whether to requeue or not or to raise the error to the reconcile loop.

internal/controller/postgrescluster_controller.go
```diff
@@ -0,0 +1,306 @@
...
+
+	return resultConfig
+}
+
+// buildCNPGClusterSpec builds the desired CNPG ClusterSpec and returns an error if mandatory fields are missing.
+func (r *PostgresClusterReconciler) buildCNPGClusterSpec(mergedConfig *enterprisev4.PostgresClusterSpec) cnpgv1.ClusterSpec {
+
+	resources := corev1.ResourceRequirements{}
+	if mergedConfig.Resources != nil {
+		resources = *mergedConfig.Resources
```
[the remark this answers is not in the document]
DmytroPI-dev: Will remove this, as we already check resources it `getMergedConfig` method, missed this.

internal/controller/postgrescluster_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// 4. Final state reached (Ready or Failed)
+	return ctrl.Result{}, nil
+}
+
+// syncStatus is a helper function to map CNPG Cluster state to PostgresCluster object and commit it to API.
+func (r *PostgresClusterReconciler) syncStatus(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster, cnpgCluster *cnpgv1.Cluster, err error) error {
+	// will use Patch as we did for main reconciliation loop.
+	latestPGCluster := client.MergeFrom(postgresCluster.DeepCopy())
```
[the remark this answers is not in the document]
DmytroPI-dev: I was using `Update` initially, but it causes error as below: 
```
"Operation cannot be fulfilled on postgresclusters.enterprise.splunk.com \"postgresql-cluster-dev-overridden\": the object has been modified; please apply your changes to the latest version and try again". 
``` 
As this was the very  same error I noticed earlier, when decided to switch from `Update` to `Patch` in `Reconcile` method, I decided to use `Patch` here as well. If the above error is not critical, we can use Update, of course, but I assume that any error logs will be confusing, though it resolves automatically in next reconciliation loop.

internal/controller/postgrescluster_controller.go
DmytroPI-dev: Refactored, moved logic to Reconcile

mploski (approved):

PR #1702, opened 2026-02-17
Cspl 4375 exclude tests when not needed

kubabuczak:
### Description

Don't run tests when there are no relevant changes

### DIFF
--- a/.github/workflows/distroless-build-test-push-workflow.yml
+++ b/.github/workflows/distroless-build-test-push-workflow.yml
@@ -7,15 +7,15 @@ permissions:
 on:
   pull_request:
     paths-ignore:
-    - 'docs/**'
-    - '*.md'
+      - 'docs/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/kubectl-splunk-workflow.yml
```diff
@@ -9,11 +9,11 @@ permissions:
...
   push:
-    branches: 
-        - feature/CSPL-3152 
+    branches:
+        - feature/CSPL-3152
   pull_request:
-    branches: 
-        - feature/CSPL-3152 
+    branches:
+        - feature/CSPL-3152
```
kasiakoziol: This branch doesn't exist anymore. I have no idea why it was used, but looks like we never execute this workflow and that's why it was not discoevered.

Igor-splunk (approved):

Issue #1652, opened 2025-12-17
Splunk Operator: CR label/annotation changes trigger unexpected rolling restarts


ductrung-nguyen:
### Please select the type of request

Enhancement

### Tell us more

**Describe the request**

When you add or modify labels/annotations on a Splunk CR (like IndexerCluster), it causes a full rolling restart of all pods. This is pretty disruptive - especially for indexer clusters where each pod has to go through the whole decommissioning cycle (reassigning primaries, graceful shutdown, etc.).

For example, just adding a simple label like `team: platform` to organize resources shouldn't bring down your entire cluster one pod at a time.

In enterprise environment, we need to use labels/annotations on K8s resources for many purpose: management, FinOps....
So, this issue is quite annoying for us.

**Expected behavior**

Adding metadata (labels/annotations) to a CR shouldn't automatically restart all the pods. Labels are commonly used for organizational purposes (cost tracking, team ownership, filtering in dashboards), and most operators treat them as non-disruptive changes.

Ideally:
- CR labels/annotations should propagate to the StatefulSet level only (which doesn't cause restarts)
- If users actually need labels on the pods themselves, there should be an explicit way to do that (like `spec.podLabels`) with clear docs that it'll trigger a rolling update

**Splunk setup on K8S**

Any Splunk deployment managed by the operator - Standalone, IndexerCluster, SearchHeadCluster, etc.

**Reproduction steps**

1. Deploy an IndexerCluster with 3 replicas
2. Wait for all pods to be ready
3. Add a label to the CR: `kubectl label indexercluster my-cluster team=platform`
4. Watch all 3 indexer pods get recycled one by one (with full decommissioning for each)

**K8s environment**

Happens on any K8s version - it's an operator behavior, not K8s specific.

**Additional context**

The root cause is that CR metadata gets propagated to the Pod Template (`.spec.template.metadata`), which changes the controller-revision-hash and triggers the operator's rolling update logic.

Most other operators either:
- Don't propagate CR metadata to pods at all
- Provide separate fields for "safe" metadata vs "pod-level" metadata
- Have an opt-in mechanism for changes that require restarts

PR #1720, opened 2026-02-24
feat: implement prefix-based metadata filtering for StatefulSet and P…

ductrung-nguyen:
### Description

This PR implements prefix-based metadata filtering to give users control over how labels and annotations propagate from the Custom Resource (CR) to the underlying StatefulSet and Pods.
**Note**: This is a part of https://github.com/splunk/splunk-operator/pull/1644.
It contains the first [commit](https://github.com/splunk/splunk-operator/pull/1644/changes/c61061dcf99c6022ee33e91652c64222211c5cb6) + a bug fix which we discovered after a long validation.

Previously, almost all metadata was copied everywhere. Now you can use specific prefixes to target where you want your labels to go:

*   **`sts-only.*`**: Keys starting with this are added *only* to the StatefulSet metadata (and the prefix is stripped!). They won't appear on the Pod Template, so adding them won't trigger a pod rollout.
*   **`pod-only.*`**: Keys are added *only* to the Pod Template (and excluded from the StatefulSet metadata).
*   **System prefixes**: We now automatically exclude `kubectl.kubernetes.io/*` and `operator.splunk.com/*` prefixes to prevent clutter and conflicts.

Regular labels without these prefixes work exactly as before—propagatating to both.

### Key Changes

*   Added the core filtering logic. It checks for `sts-only.` and `pod-only.` prefixes.
*   Implemented `SyncParentMetaToStatefulSet` which handles the transformation (stripping `sts-only.`) and syncing logic.
*   Updated the main reconciliation loop to use these new syncing functions instead of the old blind copy.
*   Added logic to detect if *only* StatefulSet metadata changed (so we can update the STS object without touching the Pod specs).

### Testing and Verification

Many unit tests are implemented.

For E2E validation:
1.  Update the existing Splunk CR.
2.  Add a label `sts-only.my-tag: true`.
    *   **Verifed**: The StatefulSet has `my-tag: true`.
    *   **Verifed**: The Pods do *not* have `my-tag` (and thus no rollout if added later).
3.  Add a label `pod-only.scrapper: true`.
    *   **Verified**: The Pods have `pod-only.scrapper: true`.
    *   **Verified**: The StatefulSet does *not* have this label.
4.  Add a generic label `environment: prod`.
    *   **Verified**: Both STS and Pods have `environment: prod`.

### Related Issues
https://github.com/splunk/splunk-operator/issues/1652

### PR Checklist

- [x] All tests pass locally.

### DIFF
--- a/pkg/splunk/common/util.go
+++ b/pkg/splunk/common/util.go
@@ -17,6 +17,7 @@ package common
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"math/rand"
@@ -29,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 func init() {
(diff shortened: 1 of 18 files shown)


pkg/splunk/common/util.go
```diff
@@ -47,23 +49,690 @@ func AsOwner(cr MetaObject, isController bool) metav1.OwnerReference {
...
+// The exclusion rules are:
+//   - kubectl.kubernetes.io/*: kubectl-managed metadata (e.g., last-applied-configuration)
+//   - operator.splunk.com/*: Operator-internal metadata not meant for pods
+//   - sts-only.*: StatefulSet-only labels that should not propagate to pods
+//
+// Note: pod-only.* keys are NOT excluded - they are transformed (prefix stripped)
+// during propagation, allowing users to specify pod-specific metadata on the CR.
+var podTemplateExcludedPrefixes = []string{
+	"kubectl.kubernetes.io/",
+	"operator.splunk.com/",
```
vivekr-splunk, line 80: We are actively spiking the StatefulSet lifecycle / reload / rolling restart direction in #1710, since that is where we are thinking through how we want StatefulSet behavior to be managed overall. Looking at the overlap, this metadata filtering work appears to be something we can layer on top of that path rather than a separate direction: I checked it against #1710 and the core source files auto-merge, with the remaining conflicts mostly in fixtures/tests plus the reserved metadata namespace choice. One thing I would like to align first is the operator-owned metadata namespace here (`operator.splunk.com/*`) versus the pod lifecycle metadata used in #1710 (`splunk.com/pod-intent`, `splunk.com/pod-cleanup`). Do you think the #1710 approach would solve the underlying StatefulSet behavior problem you are addressing here? If yes, my preference would be to settle that base behavior first and then rebase/add this PR on top of it.

vivekr-splunk (changes requested):
Requesting changes because this PR starts defining operator-reserved StatefulSet metadata behavior, and that needs to stay aligned with the intended StatefulSet lifecycle direction. The metadata filtering idea here is useful, but before we merge it I think we should make sure the reserved internal metadata namespace is consistent with the operator-owned pod lifecycle annotations/finalizers used in the ongoing StatefulSet lifecycle design work.

PR #1731, opened 2026-02-26
spike: reduce Monitoring Console restart churn from ConfigMap metadata updates

vivekr-splunk:
## Summary
- switch MC pod-template annotation `monitoringConsoleConfigRev` to a deterministic hash of MC ConfigMap data
- avoid rollouts when only ConfigMap metadata/resourceVersion changes
- add unit tests for hash determinism and statefulset annotation behavior

## Why
During C3-style bring-up, MC can restart multiple times. One trigger is that annotation currently follows ConfigMap `resourceVersion`, so metadata-only updates can force unnecessary restarts. This spike constrains restarts to effective data changes only.

## Testing
- go test ./pkg/splunk/enterprise -run "TestGetMonitoringConsoleConfigDataHash|TestGetMonitoringConsoleStatefulSetUsesConfigDataHash|TestGetMonitoringConsoleStatefulSet|TestApplyMonitoringConsoleEnvConfigMap" -count=1

### DIFF
--- a/pkg/splunk/enterprise/monitoringconsole.go
+++ b/pkg/splunk/enterprise/monitoringconsole.go
@@ -17,6 +17,8 @@ package enterprise
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/splunk/enterprise/monitoringconsole.go
```diff
@@ -193,19 +195,47 @@ func getMonitoringConsoleStatefulSet(ctx context.Context, client splcommon.Contr
...
 
 	// Setup App framework staging volume for apps
 	setupAppsStagingVolume(ctx, client, cr, &ss.Spec.Template, &cr.Spec.AppFrameworkConfig)
 	return ss, nil
 }
 
+// getMonitoringConsoleConfigDataHash returns a stable hash for MC configMap data.
+// Hashing data (instead of configMap resourceVersion) avoids unnecessary MC restarts
+// when only metadata changes.
+func getMonitoringConsoleConfigDataHash(data map[string]string) string {
```
kubabuczak: This looks like pretty general hashing mechanism, maybe we should move it to some util lib (`common/util.go`) or configmap specific one `pkg/splunk/splkcontroller/configmap.go`? 
Signature would be `func GetConfigMapDataHash(data map[string]string) string`

PR #1771, opened 2026-03-12
fix reconciliation storm issue for postgrescluster

DmytroPI-dev:
### Tests

created bash script to run tests locally: 

```
NAMESPACE=your-namespace UPGRADE_POSTGRES_VERSION=16 ./test/postgrescluster-retain-upgrade-flow.sh
```

### Description

Calmed reconciliation storm for `postgrescluster_controller.go`

### Key Changes

PostgresCluster does not show a runaway reconcile storm in this log. Its bursts are short, causal, and converge. 

Log analysis: 

The strongest reconcile storm in this log is PostgresDatabase, not PostgresCluster.
There is a short but real burst at operator.log (line 241): PostgresDatabase reconciles, patches users into PostgresCluster, then another PostgresDatabase reconcile starts almost immediately at operator.log (line 252) and hits a resource-version conflict at operator.log (line 256).
After that, PostgresDatabase re-enters again at operator.log (line 263) and still reports pending users at operator.log (line 272). That is classic concurrent-update churn.
The biggest burst is at operator.log (line 281) through operator.log (line 326): 5 separate PostgresDatabase reconciles in about 0.3s, all processing the same two databases and repeatedly logging CNPG Database created/updated successfully.
PostgresCluster also has a burst during recreate/provisioning at operator.log (line 343) through operator.log (line 387): multiple immediate reconciles caused by finalizer add, owner-ref restoration, CNPG patch, then pooler wait/retry.
The two Connection poolers are not ready yet, requeueing entries at operator.log (line 371) and operator.log (line 378) are bounded. They settle by operator.log (line 387), so this is not the main storm.
After that, PostgresCluster reconciles roughly every 10-19s at operator.log (line 388), operator.log (line 396), operator.log (line 404), operator.log (line 412), operator.log (line 420). That looks like normal CNPG phase-change traffic, not a hot loop.
Interpretation

The primary storm driver is PostgresDatabase reconciling the same object repeatedly while related objects are changing under it.
The clearest amplifier is the update conflict at operator.log (line 256).
The second amplifier is child-object churn: the controller appears to reconcile again after each CNPG database create/update, even when it is still converging the same desired state.
PostgresCluster does have short bursts, but they are bounded and consistent with create/delete/provisioning transitions.

**Bottom line**

Real storm: PostgresDatabase around operator.log (line 241) and especially operator.log (line 281).
Not a storm: the later PostgresCluster loop after operator.log (line 388); it looks event-driven and converges to healthy at operator.log (line 425).

### Logs files. 
[operator_before.log](https://github.com/user-attachments/files/26168590/operator_before.log) - logs before,
[operator_after.log](https://github.com/user-attachments/files/26168039/operator_after.log) - logs after.
[cluster.log](https://github.com/user-attachments/files/26165867/cluster.log) - cluster phase changes.
[postgrescluster.log](https://github.com/user-attachments/files/26165868/postgrescluster.log) - postgrescluster phase changes. 

### Issues fixed.

1. The old controller version had repeated status-update conflicts caused by multiple writes during one reconcile.
The clearest burst is at `19:46:34+01:00`:

- `Failed to sync pooler status` at operator.log:4782, operator.log:4845 operator.log:4887 operator.log:4929.
- each failure is `the object has been modified`
- at the same time, neighboring reconciles still finish with `Reconciliation complete` at operator.log:4817 operator.log:4824, operator.log:4831, operator.log:4880

That is classic self-induced churn from inline `Status().Update(...)` calls racing with other reconciles.

2. The previous controller version had pooler-related reconcile storms.
The log shows repeated successful reconciles mixed with pooler status failures in the same second:

- `Poolers are ready, syncing pooler status` at operator.log:104
- later repeated `Failed to sync pooler status` at operator.log:4782

The old flow updated pooler state in multiple places, including the tail of reconcile, which amplified conflicts and repeat work.

3. The old controller was patching managed roles repeatedly even when the role counts already matched.
Examples:

- `currentCount: 4, desiredCount: 4` but still “roles differ” at operator.log:612
- same pattern at operator.log:4829, operator.log:4836], operator.log:4927

That indicates unnecessary no-op role patching and extra reconcile noise.

4. The old controller was noisy even on healthy convergence.
It reached full success twice within milliseconds at operator.log:105 and operator.log:114, then many more times around operator.log:4817. That points to event amplification from owned resources plus repeated no-op status/resource updates.

5. The old controller had an issue with infinite loop on version major upgrade attempt. 

**Problems  fixed**

- Fixed infinite loop on version upgrade.
- Centralized status persistence addresses the old conflict storm. Moving from immediate status writes throughout reconcile to buffered status mutation plus `persistStatus()` / `persistStatusIfChanged()` directly targets the `object has been modified` failures.
- The pooler flow is now staged more cleanly. Pooler wait/delete paths use explicit requeue behavior, and pooler status is no longer updated in the same scattered way that produced the old `Failed to sync pooler status` bursts.
- The refactor removed the worst control-flow ambiguity in pooler reconciliation. The old code mixed pooler readiness, status writes, ConfigMap reconciliation, and final status sync in one tail section; the new flow separates these phases more clearly.
- Desired/current-state comparison is now more deliberate, which is the right direction for reducing the old managed-role churn where the controller kept patching despite `currentCount == desiredCount`.
- Fixed overall readability and phase separation. In the old logs, secret repair, cluster patching, managed roles, poolers, ConfigMap, and final status writes all interleaved tightly; the refactored controller makes these transitions explicit and easier to reason about.

**Bottom line**

The old log shows two main classes of problems that the refactor improved:
- reconciliation churn from repeated inline status updates and pooler status writes
- noisy no-op reconciliation work, especially around managed roles and tail-end success paths

### Related Issues

[CPI-1913](https://splunk.atlassian.net/browse/CPI-1913) - JIRA ticket
### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/internal/controller/postgrescluster_controller.go
+++ b/internal/controller/postgrescluster_controller.go
@@ -31,21 +31,34 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	client "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	logs "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-// PostgresClusterReconciler reconciles a PostgresCluster object
+// PostgresClusterReconciler reconciles PostgresCluster resources.
 type PostgresClusterReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 }
 
-// Used for applying changes both from PostgresCluster and PostgresClusterClass.
-type MergedConfig struct {
-	Spec *enterprisev4.PostgresClusterSpec
-	CNPG *enterprisev4.CNPGConfig
+// EffectiveClusterConfig holds the effective PostgresCluster spec and CNPG settings after class defaults are applied.
+type EffectiveClusterConfig struct {
+	ClusterSpec       *enterprisev4.PostgresClusterSpec
+	ProvisionerConfig *enterprisev4.CNPGConfig
+}
+
+// normalizedManagedRole holds only the fields this controller sets on a CNPG RoleConfiguration.
+// CNPG's admission webhook populates defaults (ConnectionLimit: -1, Inherit: true) that would
+// cause equality.Semantic.DeepEqual to always report a diff — we compare only what we own.
+type normalizedManagedRole struct {
+	Name           string
+	Ensure         cnpgv1.EnsureOption
+	Login          bool
+	PasswordSecret string
 }
 
 // +kubebuilder:rbac:groups=enterprise.splunk.com,resources=postgresclusters,verbs=get;list;watch;create;update;patch;delete
@@ -57,7 +70,7 @@ type MergedConfig struct {
 // +kubebuilder:rbac:groups=postgresql.cnpg.io,resources=poolers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=postgresql.cnpg.io,resources=poolers/status,verbs=get
 
-// Main reconciliation loop for PostgresCluster.
+// Reconcile drives PostgresCluster toward the desired CNPG resources and status.
 func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := logs.FromContext(ctx)
 	logger.Info("Reconciling PostgresCluster", "name", req.Name, "namespace", req.Namespace)
@@ -67,7 +80,7 @@ func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	var postgresSecretName string
 	secret := &corev1.Secret{}
 
-	// 1. Fetch the PostgresCluster instance, stop, if not found.
+	// Phase: ResourceFetch
 	postgresCluster := &enterprisev4.PostgresCluster{}
 	if getPGClusterErr := r.Get(ctx, req.NamespacedName, postgresCluster); getPGClusterErr != nil {
 		if apierrors.IsNotFound(getPGClusterErr) {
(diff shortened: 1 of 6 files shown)


internal/controller/postgrescluster_controller.go
```diff
@@ -1006,10 +1038,10 @@ func (r *PostgresClusterReconciler) generateSecret(ctx context.Context, postgres
 	} else if err != nil {
 		return err
 	}
-	if postgresCluster.Status.Resources == nil {
-		postgresCluster.Status.Resources = &enterprisev4.PostgresClusterResources{}
+	if bytes.Equal(existing.Data["username"], []byte(superUsername)) && existing.Data["password"] != nil {
```
mploski: could you explain why we added this if?

internal/controller/postgrescluster_controller.go
```diff
@@ -395,21 +400,39 @@ func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	if cnpgCluster.Status.Phase == cnpgv1.PhaseHealthy && r.arePoolersReady(ctx, postgresCluster) {
-		logger.Info("Poolers are ready, syncing pooler status")
-		r.syncPoolerStatus(ctx, postgresCluster)
+		if err := r.syncPoolerStatus(ctx, postgresCluster); err != nil {
+			logger.Error(err, "Failed to sync pooler status")
+			return ctrl.Result{}, err
+		}
+	}
+
+	if specChanged {
```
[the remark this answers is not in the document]
DmytroPI-dev: I debugged  this, as we have `specChanged := postgresCluster.Status.ObservedGeneration != postgresCluster.Generation`, so I assume this makes sense, as We reach this line only if reconcile did not return earlier. A lag exists there when the CR spec was changed before, so Kubernetes already bumped Generation, but controller has not yet finished a successful status update that copies it into Status.ObservedGeneration. So I assume we should leave it as is.

internal/controller/postgrescluster_controller.go
```diff
@@ -1195,14 +1396,67 @@ func (r *PostgresClusterReconciler) patchObject(ctx context.Context, original cl
...
+		CreateFunc: func(event.CreateEvent) bool {
+			return true
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldObj, okOld := e.ObjectOld.(*enterprisev4.PostgresCluster)
+			newObj, okNew := e.ObjectNew.(*enterprisev4.PostgresCluster)
+			if !okOld || !okNew {
```
limak9182: nit: are we checking errors here? var names are a little bit misleading to me

DmytroPI-dev: Those booleans are not error checks. They are the second return value from a Go type assertion, so they only indicate whether e.ObjectOld / e.ObjectNew had the expected concrete type.
See https://go.dev/doc/effective_go#interface_conversions

internal/controller/postgrescluster_controller.go
```diff
@@ -869,66 +1021,82 @@ func (r *PostgresClusterReconciler) arePoolersReady(ctx context.Context, postgre
...
+func buildCNPGRole(role enterprisev4.ManagedRole) cnpgv1.RoleConfiguration {
+	cnpgRole := cnpgv1.RoleConfiguration{
+		Name:   role.Name,
+		Ensure: cnpgv1.EnsurePresent,
+		Login:  true,
+	}
+	if role.Ensure == "absent" {
+		cnpgRole.Ensure = cnpgv1.EnsureAbsent
+		cnpgRole.Login = false
+		cnpgRole.Login = true
```
limak9182: some copy paste artefact I guess

limak9182: I believe it should be set to true, as we want to give users power to log into retained DBs

internal/controller/postgrescluster_controller.go
```diff
@@ -77,345 +90,455 @@ func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 		if apierrors.IsNotFound(finalizerErr) {
 			logger.Info("PostgresCluster already deleted, skipping finalizer update")
 			return ctrl.Result{}, nil
 		}
+
 		logger.Error(finalizerErr, "Failed to handle finalizer")
-		if statusErr := updateStatus(clusterReady, metav1.ConditionFalse, reasonClusterDeleteFailed, fmt.Sprintf("Failed to delete resources during cleanup: %v", finalizerErr), failedClusterPhase); statusErr != nil {
-			logger.Error(statusErr, "Failed to update status")
-		}
+		updateStatus(
```
M4KIF, line 119: Is linting standardized across the codebase? Ie. do we have a working lint script?

internal/controller/postgrescluster_controller.go
```diff
@@ -869,66 +1021,82 @@ func (r *PostgresClusterReconciler) arePoolersReady(ctx context.Context, postgre
...
+}
 
-		if role.Ensure == "absent" {
-			cnpgRole.Ensure = cnpgv1.EnsureAbsent
-		} else {
-			cnpgRole.Ensure = cnpgv1.EnsurePresent
-			cnpgRole.Login = true
-		}
+// buildCNPGRole converts a single PostgresCluster ManagedRole to its CNPG RoleConfiguration equivalent.
+// Login is only set for present roles — absent roles must not be able to authenticate.
```
limak9182: I believe absent roles should be able to authenticate to enable users to debug in retained databases without modifying them?
What are your thoughts on that?

api/v4/postgrescluster_types.go
```diff
@@ -157,6 +157,11 @@ type PostgresClusterStatus struct {
 	// Resources contains references to related Kubernetes resources like ConfigMaps and Secrets.
 	// +optional
 	Resources *PostgresClusterResources `json:"resources,omitempty"`
+
+	// ObservedGeneration represents the .metadata.generation that the status was set based upon.
+	// Used to determine if the spec has changed since the last status update.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
limak9182: is it used in the current flow?
I don't see any logic around it in code

internal/controller/postgresoperator_common_types.go
```diff
@@ -75,6 +74,7 @@ const (
 	databasesReady  conditionTypes = "DatabasesReady"
 	secretsReady    conditionTypes = "SecretsReady"
 	configMapsReady conditionTypes = "ConfigMapsReady"
+	configMapReady  conditionTypes = "ConfigMapReady"
```
limak9182, line 77: shouldn't configMapsReady be used?

internal/controller/postgrescluster_controller.go
```diff
@@ -31,23 +31,36 @@ import (
...
 
-// PostgresClusterReconciler reconciles a PostgresCluster object
+// PostgresClusterReconciler reconciles PostgresCluster resources.
 type PostgresClusterReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 }
 
-// Used for applying changes both from PostgresCluster and PostgresClusterClass.
+// MergedConfig holds the effective PostgresCluster spec and CNPG settings after class defaults are applied.
```
M4KIF: Instead of a comment explaining the reason behind struct existence, the struct member names + struct name could better represent what are they doing. 
Because later in the code it would be MergedConfig.Spec instead of let's say MergedConfig.PgClusterSpec, MergedConfig.Pg(Db)ConnectorSpec(To start decoupling from the fact that we indeed rely on cnpg.).

If there could be a better struct x members name combination, ie. more descriptive It would also be fine to apply It. Dunno, a UnifiedClusteringCfg {Postgres / Connector}. ->  not a suggestion, just thinking of trying to describe being more descriptive from the get go.

M4KIF: but that changing the naming is also outside of the task scope

internal/controller/postgrescluster_controller.go
```diff
@@ -763,51 +887,86 @@ func (r *PostgresClusterReconciler) syncStatus(ctx context.Context, postgresClus
...
-	ctx context.Context,
 	postgresCluster *enterprisev4.PostgresCluster,
 	conditionType conditionTypes,
 	status metav1.ConditionStatus,
 	reason conditionReasons,
 	message string,
-	clusterPhase reconcileClusterPhases,
-) error {
-	postgresCluster.Status.Phase = string(clusterPhase)
+	phase reconcileClusterPhases,
```
M4KIF, line 902: Isn't deleting the error handling going to hurt future work on observability?
This function now has side-effects which can't be directly and explicitly tracked.

internal/controller/postgrescluster_controller.go
```diff
@@ -1195,14 +1396,67 @@ func (r *PostgresClusterReconciler) patchObject(ctx context.Context, original cl
...
 
-// SetupWithManager sets up the controller with the Manager.
+// SetupWithManager registers the controller for PostgresCluster resources and owned CNPG Clusters.
 func (r *PostgresClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&enterprisev4.PostgresCluster{}).
-		Owns(&cnpgv1.Cluster{}).
-		Owns(&cnpgv1.Pooler{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&corev1.Secret{}).
```
limak9182, line 1205: is it our intention to also delete Owns(...) here? Don't we want to rely on predicates to filter those event out?
In case of deletion of e.g. secret it won't be recreated.

internal/controller/postgrescluster_controller.go
DmytroPI-dev: Fixed

internal/controller/postgrescluster_controller.go
DmytroPI-dev: Not really. There is no single standardized lint entrypoint in this repo.
What exists:
make scheck runs staticcheck from Makefile (line 131)
make vet runs go vet from Makefile (line 135)
make lang runs a biased-language checker from Makefile (line 329)
Also, make run also runs go vet and go fmt.

api/v4/postgrescluster_types.go
DmytroPI-dev: Fixed. It is not required anymore.

internal/controller/postgresoperator_common_types.go
DmytroPI-dev: I used configMapReady (in singular), as we have only one ConfigMap created by PostgresCluster, containing connection data, thus ConfigMapsReady are used in postgresDatabase controller, where we generate several CM. Does it makes sense?

internal/controller/postgrescluster_controller.go
DmytroPI-dev: I update naming for better clarity.

internal/controller/postgrescluster_controller.go
DmytroPI-dev: I assume, here we can actually remove error returns,  because the work they do is in-memory only:

assign `postgresCluster.Status.Phase`
call `meta.SetStatusCondition(...)` which returns `bool` (func SetStatusCondition(conditions *[]metav1.Condition, newCondition metav1.Condition) (changed bool) {})

Neither operation returns an error. The actual observable failure point is the status write in 
`persistStatus` and `persistStatusIfChanged` methods, which return errors.

internal/controller/postgrescluster_controller.go
DmytroPI-dev: You are correct, good catch!

internal/controller/postgrescluster_controller.go
DmytroPI-dev: I assume, if a role is absent, the contract should be: the operator wants it removed. Letting it authenticate contradicts that intent and makes the field unclear. You end up with “delete this role, but also keep it usable,” which is contradictory.

Maybe, we should discuss (TODO) the following model: 
For retained databases, the cleaner model is:

retained DB
retained roles
roles stay present
(optionally) separate policy, like  “retained-read-only” or “retained-unmanaged”
We already have the retained-role path keeps those roles present.

limak9182: that makes sense, I believe situation might be different for cluster and databases

PR #1789, opened 2026-03-20
unit tests for postgres cluster for helpers

limak9182:
### Description

_What does this PR have in it?_

### Key Changes

_Highlight the updates in specific files_

### Testing and Verification

### Related Issues

### PR Checklist

- [ ] All tests pass locally.

Local testing:
```bash
# Creation
kubectl apply -f - <<'EOF'
  apiVersion: enterprise.splunk.com/v4
  kind: PostgresCluster
  metadata:
    name: test-cluster-2
    namespace: test1
  spec:
    class: postgresql-dev
    connectionPoolerEnabled: true
  EOF

kubectl apply -f - <<'EOF'
  apiVersion: enterprise.splunk.com/v4
  kind: PostgresDatabase
  metadata:
    name: myapp-db
    namespace: test1
  spec:
    cluster: test-cluster-2
    databases:
      - name: myapp
    users:
      - name: myapp_user
        passwordSecretRef:
          name: myapp-user-secret
  EOF

# validation
--- PostgresCluster status ---
NAME             CLASS            PHASE   AGE
test-cluster-2   postgresql-dev   Ready   5m19s

--- CNPG Cluster ---
NAME             AGE    INSTANCES   READY   STATUS                     PRIMARY
test-cluster-2   101s   1           1       Cluster in healthy state   test-cluster-2-1

--- Pods ---
NAME                                        READY   STATUS    RESTARTS   AGE
test-cluster-2-1                            1/1     Running   0          92s
test-cluster-2-pooler-ro-67d44b9757-8clcb   1/1     Running   0          81s
test-cluster-2-pooler-ro-67d44b9757-hnrwm   1/1     Running   0          81s
test-cluster-2-pooler-rw-9588d4b69-2mzfh    1/1     Running   0          81s
test-cluster-2-pooler-rw-9588d4b69-cpbg5    1/1     Running   0          81s

-- PostgresDatabase ---
NAME       CLUSTER          PHASE   AGE
cr-beta    test-cluster     Ready   2d19h
myapp-db   test-cluster-2   Ready   2m22s

```


M4KIF:
1. nit: unit tests in go have a naming convention of <functionality_filename+test.go>, so It would be `postgrescluster_controler_test.go`. It exists in the package that It's meant to test, that == a unit test in go world.
2. nit: The tests could (if possible) use of assert package, to streamline the handling of string comparsion, bool predicates etc. so then instead of a if branch with string ==, or boolean conditions It would come down to a line with assert.Equal/Contains/True etc.(t, <condition-tested>, <expected-res>)
3. Those tests do contain the "integrations" with pgbouncer logic, or k8s api's. So there's a breach between the business domain (the provisions and business operations that we want to set, provide to the cluster, let's say some rules, values for resource management, etc. other decision steps) logic with the dependencies (which are pgbouncer api's, k8s's api's etc.). 
But I don't know whether we want to follow with DDD and P&A. Nor do I know when We will split into business <pause> dependencies <pause> application level (integration of dependencies with business logic through ports). That's a discussion to be had?

In general, all above come to choice and the decisions that we've made and are focused on sticking too. I would suggest Michał to take a look too.

internal/controller/postgrescluster_controller.go
```diff
@@ -855,18 +822,22 @@ func (r *PostgresClusterReconciler) getPoolerInstanceCount(pooler *cnpgv1.Pooler
 // arePoolersReady checks if both RW and RO poolers have all instances scheduled.
 func (r *PostgresClusterReconciler) arePoolersReady(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster) bool {
 	rwPooler := &cnpgv1.Pooler{}
-	rwErr := r.Get(ctx, types.NamespacedName{
```
M4KIF: nit: Do we agree on refactoring on the go, or sticking to separate task/domains? 
I don't know tbh, whether we are sticking to something or do we have a team rule.

internal/controller/postgrescluster_controller.go
```diff
@@ -158,13 +158,16 @@ func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		if generateSecretErr := r.generateSecret(ctx, postgresCluster, postgresSecretName, secret); generateSecretErr != nil {
+		createdSecret, generateSecretErr := r.generateSecret(ctx, postgresCluster, postgresSecretName)
+		if generateSecretErr != nil {
 			logger.Error(generateSecretErr, "Failed to ensure PostgresCluster secret", "name", postgresSecretName)
 			if statusErr := updateStatus(clusterReady, metav1.ConditionFalse, reasonUserSecretFailed, fmt.Sprintf("Failed to generate PostgresCluster secret: %v", generateSecretErr), failedClusterPhase); statusErr != nil {
 				logger.Error(statusErr, "Failed to update status")
 			}
 			return ctrl.Result{}, generateSecretErr
 		}
+		secret = createdSecret
```
mploski: why do we need it?

internal/controller/postgrescluster_controller.go
```diff
@@ -524,16 +493,20 @@ func (r *PostgresClusterReconciler) createOrUpdateConnectionPooler(
...
 	}
 
-	// Create/Update RO Pooler
-	if err := r.createConnectionPooler(ctx, postgresCluster, mergedConfig, cnpgCluster, readOnlyEndpoint); err != nil {
+	roCreated, err := r.createConnectionPooler(ctx, postgresCluster, mergedConfig, cnpgCluster, readOnlyEndpoint)
+	if err != nil {
 		return fmt.Errorf("failed to reconcile RO pooler: %w", err)
 	}
 
+	if rwCreated || roCreated {
```
mploski: why not let reconciller to update status and reque if poolers are not operational yet?

internal/controller/postgrescluster_controller.go
```diff
@@ -855,18 +822,22 @@ func (r *PostgresClusterReconciler) getPoolerInstanceCount(pooler *cnpgv1.Pooler
 // arePoolersReady checks if both RW and RO poolers have all instances scheduled.
 func (r *PostgresClusterReconciler) arePoolersReady(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster) bool {
 	rwPooler := &cnpgv1.Pooler{}
-	rwErr := r.Get(ctx, types.NamespacedName{
+	if err := r.Get(ctx, types.NamespacedName{
```
mploski: this should be fetched on the reconciler level and pass down to the check function

internal/controller/postgrescluster_controller.go
```diff
@@ -959,13 +930,9 @@ func normalizeCNPGClusterSpec(spec cnpgv1.ClusterSpec, customDefinedParameters m
...
 }
 
-// generateConfigMap generates a ConfigMap with connection details for the PostgresCluster.
-func (r *PostgresClusterReconciler) generateConfigMap(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster, cnpgCluster *cnpgv1.Cluster, secretName string) (*corev1.ConfigMap, error) {
-	configMapName := fmt.Sprintf("%s%s", postgresCluster.Name, defaultConfigMapSuffix)
-	if postgresCluster.Status.Resources != nil && postgresCluster.Status.Resources.ConfigMapRef != nil {
-		configMapName = postgresCluster.Status.Resources.ConfigMapRef.Name
-	} 
-	
+// buildConfigMapData is a pure function — no API calls. Builds the data map for the
```
mploski: remove the comment?

internal/controller/postgrescluster_controller.go
```diff
@@ -974,71 +941,89 @@ func (r *PostgresClusterReconciler) generateConfigMap(ctx context.Context, postg
...
+	if poolersExist {
 		data["CLUSTER_POOLER_RW_ENDPOINT"] = fmt.Sprintf("%s.%s", poolerResourceName(cnpgCluster.Name, readWriteEndpoint), cnpgCluster.Namespace)
 		data["CLUSTER_POOLER_RO_ENDPOINT"] = fmt.Sprintf("%s.%s", poolerResourceName(cnpgCluster.Name, readOnlyEndpoint), cnpgCluster.Namespace)
 	}
+	return data
+}
+
+// reconcileConfigMap resolves the ConfigMap name, checks pooler existence, builds and
+// creates or updates the ConfigMap with connection details for the PostgresCluster.
+func (r *PostgresClusterReconciler) reconcileConfigMap(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster, cnpgCluster *cnpgv1.Cluster, secretName string) (string, error) {
```
mploski: thats too much for one function. reconcileConfigMap should be responsible for config map build not fetching if pooler exists. can you pass this info as a parameter?

internal/controller/postgrescluster_controller.go
```diff
@@ -974,71 +941,89 @@ func (r *PostgresClusterReconciler) generateConfigMap(ctx context.Context, postg
...
-		return err
+		return nil
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to reconcile ConfigMap %s: %w", configMapName, err)
 	}
-	if postgresCluster.Status.Resources == nil {
-		postgresCluster.Status.Resources = &enterprisev4.PostgresClusterResources{}
+
+	switch result {
```
mploski: why do we even care if it was update/create/or no-op?

internal/controller/postgrescluster_unit_test.go
```diff
@@ -0,0 +1,1247 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			got := normalizeCNPGClusterSpec(tt.spec, tt.customDefinedParameters)
+
+			if !equality.Semantic.DeepEqual(got, tt.expected) {
```
mploski: it would be better to use standard testing libs for comparision i.e https://pkg.go.dev/github.com/stretchr/testify/assert as a default choice it can also compare objects https://levinology.com/blog/2022-10-01-comparing-structs-testify/

internal/controller/postgrescluster_unit_test.go
```diff
@@ -0,0 +1,1247 @@
...
+				"hostssl all all 0.0.0.0/0 scram-sha-256",
+				"host replication all 10.0.0.0/8 md5",
+			},
+			Resources: &corev1.ResourceRequirements{},
+		},
+	}
+
+	spec := r.buildCNPGClusterSpec(cfg, "my-secret")
+
+	t.Run("image name includes version", func(t *testing.T) {
```
mploski: we dont need to paralellise attribute checking via running this in separate t.run. Lets keep those tests consise

mploski:
> 1. nit: unit tests in go have a naming convention of <functionality_filename+test.go>, so It would be `postgrescluster_controler_test.go`. It exists in the package that It's meant to test, that == a unit test in go world.
> 2. nit: The tests could (if possible) use of assert package, to streamline the handling of string comparsion, bool predicates etc. so then instead of a if branch with string ==, or boolean conditions It would come down to a line with assert.Equal/Contains/True etc.(t, , )
> 3. Those tests do contain the "integrations" with pgbouncer logic, or k8s api's. So there's a breach between the business domain (the provisions and business operations that we want to set, provide to the cluster, let's say some rules, values for resource management, etc. other decision steps) logic with the dependencies (which are pgbouncer api's, k8s's api's etc.).
>    But I don't know whether we want to follow with DDD and P&A. Nor do I know when We will split into business  dependencies  application level (integration of dependencies with business logic through ports). That's a discussion to be had?
> 
> In general, all above come to choice and the decisions that we've made and are focused on sticking too. I would suggest Michał to take a look too.

2. Agree lets use https://pkg.go.dev/github.com/stretchr/testify/assert for comparision
3. Indeed we have pure unit tests and tests that tests integration with k8s api's/cnpg api. I think this is the time next week to sit down and draw the business vs integration separation logic

internal/controller/postgrescluster_controller.go
limak9182: We agreed to do some really small refactors

internal/controller/postgrescluster_controller.go
limak9182: N/A pulled from target branch, changed there

internal/controller/postgrescluster_controller.go
limak9182: done

internal/controller/postgrescluster_controller.go
limak9182: done

internal/controller/postgrescluster_controller.go
limak9182: resolved by pullling from main

internal/controller/postgrescluster_unit_test.go
limak9182: moved to testify - assert

internal/controller/postgrescluster_unit_test.go
limak9182: done

internal/controller/postgrescluster_controller.go
limak9182: done

internal/controller/postgrescluster_controller_unit_test.go
```diff
@@ -0,0 +1,1010 @@
...
+		require.NoError(t, err)
+		assert.NotNil(t, cfg.ClusterSpec.PostgreSQLConfig)
+		assert.NotNil(t, cfg.ClusterSpec.PgHBA)
+		assert.NotNil(t, cfg.ClusterSpec.Resources)
+	})
+}
+
+func TestBuildCNPGClusterSpec(t *testing.T) {
+	r := &PostgresClusterReconciler{}
+
```
DmytroPI-dev: do we need to validate not only the basic CNPG cluster spec fields but also asserts propagation of class-level CNPG spec fields?

pkg/postgresql/cluster/core/cluster_unit_test.go
```diff
@@ -0,0 +1,1010 @@
...
+		modified := original.DeepCopy()
+		modified.Data = map[string][]byte{"key": []byte("value")}
+
+		err := r.patchObject(context.Background(), original, modified, "Secret")
+
+		assert.NoError(t, err)
+	})
+}
+
+func TestDeleteCNPGCluster(t *testing.T) {
```
DmytroPI-dev, line 626: do we need to check if delete returns an error?

internal/controller/postgrescluster_controller_unit_test.go
limak9182: I believe this unit test should be only responsible for testing this particular functionality, functional tests might cover the whole case of checking parameters between PostgresCluster and PostgresClusterClass.

pkg/postgresql/cluster/core/cluster_unit_test.go
limak9182: What I wanted to check here is the scenerio when the cluster was already deleted, we don't want an error here but just nil. In the future when this function might be modified we have an extra guard that our reconciliation logic relies on that.

mploski (approved):

PR #1791, opened 2026-03-23
Move postgres controller code to package, cleanup crds, fix minor issues

mploski:
### Description

_What does this PR have in it?_

### Key Changes

_Highlight the updates in specific files_

### Testing and Verification

### Related Issues

### PR Checklist

- [ ] All tests pass locally.


DmytroPI-dev (commented):
We have pkg/postgresql/database/core/types.go and pkg/postgresql/cluster/core/types.go, both in core package, can we have some conflicts here?

mploski:
> We have pkg/postgresql/database/core/types.go and pkg/postgresql/cluster/core/types.go, both in core package, can we have some conflicts here?

We might, in this case we can have separate "shared" directory with reusable types later on. Currently most of the things are not shared

internal/controller/postgrescluster_controller.go
```diff
@@ -18,35 +18,26 @@ package controller
...
 	Scheme *runtime.Scheme
 }
 
-// Used for applying changes both from PostgresCluster and PostgresClusterClass.
-type MergedConfig struct {
-	Spec *enterprisev4.PostgresClusterSpec
-	CNPG *enterprisev4.CNPGConfig
-}
+const (
+	ClusterTotalWorker int = 2
```
limak9182: Nit: Is that a requirement for SOK? I wasn't testing the code against two parallel goroutines running at the same time. Does it work such that one worker starts on an event, but if it's still processing and another event comes in, it doesn't just sit in the queue, it can be picked up by another worker? I’d like to trigger that for future local tests, that's why I’m asking.

limak9182 (approved):

DmytroPI-dev (approved):

internal/controller/postgrescluster_controller.go
mploski: They just asked us to provide this info explicit. Yes your assumption is correct how it works, with one exception: if changes is related to the same CR it still will be queued to provide proper serialisation

PR #1793, opened 2026-03-24
unit tests for postgres database

DmytroPI-dev:
### Description

_Unit tests for postgresql/database/core_

### Key Changes

_Add database_unit_test.go_

### Testing and Verification

_The PR is for adding unit tests._

### Related Issues

[CPI-1915](https://splunk.atlassian.net/browse/CPI-1915)

### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/pkg/postgresql/database/core/database.go
+++ b/pkg/postgresql/database/core/database.go
@@ -41,7 +41,7 @@ func PostgresDatabaseService(
 	logger.Info("Reconciling PostgresDatabase", "name", postgresDB.Name, "namespace", postgresDB.Namespace)
 
 	updateStatus := func(conditionType conditionTypes, conditionStatus metav1.ConditionStatus, reason conditionReasons, message string, phase reconcileDBPhases) error {
-		return setStatus(ctx, c, postgresDB, conditionType, conditionStatus, reason, message, phase)
+		return persistStatus(ctx, c, postgresDB, conditionType, conditionStatus, reason, message, phase)
 	}
 
 	// Finalizer: cleanup on deletion, register on creation.
@@ -68,23 +68,24 @@ func PostgresDatabaseService(
 	}
 
 	// Phase: ClusterValidation
-	cluster, clusterStatus, err := ensureClusterReady(ctx, c, postgresDB)
+	cluster, err := fetchCluster(ctx, c, postgresDB)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			if err := updateStatus(clusterReady, metav1.ConditionFalse, reasonClusterNotFound, "Cluster CR not found", pendingDBPhase); err != nil {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{RequeueAfter: clusterNotFoundRetryDelay}, nil
+		}
 		if statusErr := updateStatus(clusterReady, metav1.ConditionFalse, reasonClusterInfoFetchFailed,
 			"Can't reach Cluster CR due to transient errors", pendingDBPhase); statusErr != nil {
 			logger.Error(statusErr, "Failed to update status")
 		}
 		return ctrl.Result{}, err
 	}
+	clusterStatus := getClusterReadyStatus(cluster)
 	logger.Info("Cluster validation done", "clusterName", postgresDB.Spec.ClusterRef.Name, "status", clusterStatus)
 
 	switch clusterStatus {
-	case ClusterNotFound:
-		if err := updateStatus(clusterReady, metav1.ConditionFalse, reasonClusterNotFound, "Cluster CR not found", pendingDBPhase); err != nil {
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{RequeueAfter: clusterNotFoundRetryDelay}, nil
-
 	case ClusterNotReady, ClusterNoProvisionerRef:
 		if err := updateStatus(clusterReady, metav1.ConditionFalse, reasonClusterProvisioning, "Cluster is not in ready state yet", pendingDBPhase); err != nil {
 			return ctrl.Result{}, err
@@ -299,23 +300,27 @@ func reconcileRWRolePrivileges(
 	return stderrors.Join(errs...)
 }
 
-func ensureClusterReady(ctx context.Context, c client.Client, postgresDB *enterprisev4.PostgresDatabase) (*enterprisev4.PostgresCluster, clusterReadyStatus, error) {
+func fetchCluster(ctx context.Context, c client.Client, postgresDB *enterprisev4.PostgresDatabase) (*enterprisev4.PostgresCluster, error) {
 	logger := log.FromContext(ctx)
 	cluster := &enterprisev4.PostgresCluster{}
 	if err := c.Get(ctx, types.NamespacedName{Name: postgresDB.Spec.ClusterRef.Name, Namespace: postgresDB.Namespace}, cluster); err != nil {
 		if errors.IsNotFound(err) {
-			return nil, ClusterNotFound, nil
+			return nil, err
 		}
 		logger.Error(err, "Failed to fetch Cluster", "name", postgresDB.Spec.ClusterRef.Name)
-		return nil, ClusterNotReady, err
+		return nil, err
 	}
+	return cluster, nil
+}
+
+func getClusterReadyStatus(cluster *enterprisev4.PostgresCluster) clusterReadyStatus {
 	if cluster.Status.Phase == nil || *cluster.Status.Phase != string(ClusterReady) {
-		return cluster, ClusterNotReady, nil
+		return ClusterNotReady
 	}
 	if cluster.Status.ProvisionerRef == nil {
-		return cluster, ClusterNoProvisionerRef, nil
+		return ClusterNoProvisionerRef
 	}
-	return cluster, ClusterReady, nil
+	return ClusterReady
 }
 
 func getDesiredUsers(postgresDB *enterprisev4.PostgresDatabase) []string {
@@ -383,30 +388,8 @@ func parseRoleNames(raw []byte) []string {
 
 func patchManagedRoles(ctx context.Context, c client.Client, postgresDB *enterprisev4.PostgresDatabase, cluster *enterprisev4.PostgresCluster) error {
 	logger := log.FromContext(ctx)
-	allRoles := make([]enterprisev4.ManagedRole, 0, len(postgresDB.Spec.Databases)*2)
-	for _, dbSpec := range postgresDB.Spec.Databases {
-		allRoles = append(allRoles,
-			enterprisev4.ManagedRole{
-				Name:   adminRoleName(dbSpec.Name),
-				Exists: true,
-				PasswordSecretRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: roleSecretName(postgresDB.Name, dbSpec.Name, secretRoleAdmin)},
-					Key: secretKeyPassword},
-			},
-			enterprisev4.ManagedRole{
-				Name:   rwRoleName(dbSpec.Name),
-				Exists: true,
-				PasswordSecretRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: roleSecretName(postgresDB.Name, dbSpec.Name, secretRoleRW)},
-					Key: secretKeyPassword},
-			})
-	}
-	rolePatch := &unstructured.Unstructured{
-		Object: map[string]any{
-			"apiVersion": cluster.APIVersion,
-			"kind":       cluster.Kind,
-			"metadata":   map[string]any{"name": cluster.Name, "namespace": cluster.Namespace},
-			"spec":       map[string]any{"managedRoles": allRoles},
-		},
-	}
+	allRoles := buildManagedRoles(postgresDB.Name, postgresDB.Spec.Databases)
+	rolePatch := buildManagedRolesPatch(cluster, allRoles)
 	fieldManager := fieldManagerName(postgresDB.Name)
 	if err := c.Patch(ctx, rolePatch, client.Apply, client.FieldOwner(fieldManager)); err != nil {
 		logger.Error(err, "Failed to add users to PostgresCluster", "postgresDatabase", postgresDB.Name)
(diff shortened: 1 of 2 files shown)


pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+		{
+			Manager: "postgresdatabase-newer",
+			FieldsV1: &metav1.FieldsV1{
+				Raw: managedRolesFieldsRaw(t, "k:"+string(roleKey)),
+			},
+		},
+	}
+
+	got := managedRoleOwners(managedFields)
+	want := map[string]string{
```
mploski, line 236: In general want should be a part of Arrange if we want to follow AAA pattern - this is general comment

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+			wantName:   "primary",
+		},
+		{
+			name: "returns no provisioner ref when phase is ready but ref is missing",
+			cluster: &enterprisev4.PostgresCluster{
+				ObjectMeta: metav1.ObjectMeta{Name: "primary", Namespace: "dbs"},
+				Status: enterprisev4.PostgresClusterStatus{
+					Phase: func() *string {
+						v := string(ClusterReady)
+						return &v
```
mploski: why do we need this, maybe we can also levarage  Ptr helper function `strPtr(s string) *string`

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+		{Name: "payments", DeletionPolicy: deletionPolicyRetain},
+		{Name: "analytics"},
+		{Name: "audit", DeletionPolicy: deletionPolicyRetain},
+	}
+
+	got := buildDeletionPlan(databases)
+
+	require.Len(t, got.retained, 2)
+	require.Len(t, got.deleted, 1)
+	assert.Equal(t, "payments", got.retained[0].Name)
```
mploski: can we levarage assert.ElementsMatch to avoid hardcoding the order?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+		require.NoError(t, orphanConfigMaps(context.Background(), c, postgresDB, databases))
+	})
+
+	t.Run("orphanSecrets skips already retained secret", func(t *testing.T) {
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        "primary-payments-admin",
+				Namespace:   "dbs",
+				Annotations: map[string]string{annotationRetainedFrom: postgresDB.Name},
+				OwnerReferences: []metav1.OwnerReference{
```
mploski: since you wan to test retained secret, shouldn't owner references be empty?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+		t.Run(tst.name, func(t *testing.T) {
+			repos := make(map[string]*stubDBRepo, len(tst.dbNames))
+			repoCalls := make([]string, 0, len(tst.dbNames))
+
+			for _, dbName := range tst.dbNames {
+				repos[dbName] = &stubDBRepo{execErr: tst.execErrs[dbName]}
+			}
+
+			newDBRepo := func(_ context.Context, host, dbName, password string) (DBRepo, error) {
+				assert.Equal(t, "rw.example.internal", host)
```
mploski: do we need this asserts here? what is the benefit?

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: In `database.go`, in `orphanSecret` method, we have the following: 
1. Fetch the secret 
2. if `secret.Annotations[annotationRetainedFrom] == postgresDB.Name`, we `continue` immediately, and only otherwise it strips owner ref and add annotation. 

So for this specific test, the point is to verify the early-return path: if a secret is already marked retained, the helper should leave it alone. Leaving  an owner reference in the fixture might be useful, as it proves the function did not mutate the object. If there would be empty owner ref, test would not show if helper truly skipped mutation.
Does it makes sense?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+// - handleDeletion
+// - orphanRetainedResources
+// - deleteRemovedResources
+// - cleanupManagedRoles
+// - patchManagedRolesOnDeletion
+// - reconcileUserSecrets
+// - ensureSecret
+// - createUserSecret
+// - reconcileRoleConfigMaps
+// - generatePassword
```
mploski: I think we can still test it if the secret has specific size etc

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+			want: nil,
+		},
+	}
+
+	for _, tst := range tests {
+
+		t.Run(tst.name, func(t *testing.T) {
+			got := slices.Clone(parseRoleNames(tst.raw))
+			want := slices.Clone(tst.want)
+			slices.Sort(got)
```
mploski: I guess we dont need to sort them and compare like that, we can use asserts https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+	roleKey, err := json.Marshal(map[string]string{"name": "main_db_admin"})
+	require.NoError(t, err)
+	secondRoleKey, err := json.Marshal(map[string]string{"name": "main_db_rw"})
+	require.NoError(t, err)
+
+	managedFields := []metav1.ManagedFieldsEntry{
+		{Manager: "ignored"},
+		{
+			Manager: "postgresdatabase-other",
+			FieldsV1: &metav1.FieldsV1{
```
mploski, line 221: should w ealso check things like FieldsV1 being nil, or Fields not containg managedRoles field?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+						Raw: managedRolesFieldsRaw(t, "k:"+string(sameOwnerKey)),
+					},
+				},
+			},
+		},
+	}
+
+	got := getRoleConflicts(postgresDB, cluster)
+	want := []string{"main_db_admin (owned by postgresdatabase-legacy)"}
+	slices.Sort(got)
```
mploski: why we sort?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+			wantNotReady: []string{"main_db_rw"},
+		},
+		{
+			name:          "returns empty when all roles are reconciled",
+			expectedUsers: []string{"main_db_admin"},
+			cluster: &cnpgv1.Cluster{
+				Status: cnpgv1.ClusterStatus{
+					ManagedRolesStatus: cnpgv1.ManagedRoles{
+						ByStatus: map[cnpgv1.RoleStatus][]string{
+							cnpgv1.RoleStatusReconciled: {"main_db_admin"},
```
mploski, line 347: should we also check a case when role in pending reconcillation? RoleStatusPendingReconciliation?

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+
+			require.Error(t, err)
+			for _, wantMsg := range tst.wantErrContains {
+				assert.ErrorContains(t, err, wantMsg)
+			}
+		})
+	}
+}
+
+// Uses a fake client because the helper's decision depends on reading a Cluster object.
```
mploski: maybe then instead of fetching the object insisde of the function we should pass enterprisev4.PostgresCluster{} instead of: postgresDB *enterprisev4.PostgresDatabase into it?

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: Removed.

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1060 @@
...
+
+		secret, err := getSecret(context.Background(), c, "dbs", "db-secret")
+
+		require.NoError(t, err)
+		require.NotNil(t, secret)
+		assert.Equal(t, existing.Name, secret.Name)
+		assert.Equal(t, "value", string(secret.Data[secretKeyPassword]))
+	})
+
+	t.Run("returns nil nil when secret is absent", func(t *testing.T) {
```
mploski, line 716: remove one nil :-)

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: Isn't FieldsV1 == nil is already exercised inside `TestManagedRoleOwners`, with the first entry:
`{Manager: "ignored"}`? As well as missing `managedRoles` already covered in `TestParseRoleNames` by the "returns empty when managed roles missing"?

DmytroPI-dev: Added test to `TestParseRoleNames` for returns empty when `spec` field is missing entirely.

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: this is correct, as in `getSecret` we have 
```
	if errors.IsNotFound(err) {
		return nil, nil
```

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: Added tests for those.

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: Made small refactoring: splitted into two methods: `fetchCluster` and `getClusterReadyStatus`, thus first keeps the API read, second checks readiness as pure function. Changed tests accordingly.

DmytroPI-dev (commented):
Refactor database helpers for narrower unit coverage:

Split cluster validation into two helpers: `fetchCluster` handles API lookup and `getClusterReadyStatus` evaluates readiness from a loaded PostgresCluster. The reconcile flow now handles NotFound explicitly after fetch, while preserving the previous status/requeue behavior.

Extract pure builders from inline reconciliation logic to improve testability without changing behavior:

buildManagedRoles for managed role generation
buildManagedRolesPatch for SSA patch payload assembly
buildCNPGDatabaseSpec for CNPG DatabaseSpec construction
applyStatus for in-memory status mutation, leaving setStatus as the persistence wrapper
Rewrite and extend unit tests accordingly:
replace ensureClusterReady tests with separate coverage for fetchCluster and getClusterReadyStatus
add direct tests for setStatus
add direct tests for reconcileUserSecrets, including idempotency
add pure unit tests for buildManagedRoles, buildManagedRolesPatch, and buildCNPGDatabaseSpec

pkg/postgresql/database/core/database.go
```diff
@@ -704,8 +682,12 @@ func deleteSecrets(ctx context.Context, c client.Client, postgresDB *enterprisev
 }
 
 func buildRetainedRoles(postgresDBName string, retained []enterprisev4.DatabaseDefinition) []enterprisev4.ManagedRole {
```
mploski: do we need this function?

pkg/postgresql/database/core/database.go
```diff
@@ -489,6 +463,11 @@ func verifyDatabasesReady(ctx context.Context, c client.Client, postgresDB *ente
 }
 
 func setStatus(ctx context.Context, c client.Client, db *enterprisev4.PostgresDatabase, conditionType conditionTypes, conditionStatus metav1.ConditionStatus, reason conditionReasons, message string, phase reconcileDBPhases) error {
```
mploski: maybe we should rename setStatus to something like PersistStatus?

pkg/postgresql/database/core/database_unit_test.go
mploski: But this is not what happens in the code. In orphanSecrets the process is the following:
1. find the databases
2. find secrets per role
3. if annotation set return
4. remove owner references
5. set annotation

In ensureSecret function we do:
1. find secret by convention
2. if doesnt exist, create one
3. if exist check if annotation is set, if yes adaptResource

In adaptResource
1. delete retain from annotation
2. set controller reference


As you can see There is no possibility that  secret has both annotation and reference and we do not allow this, and we should not add this capability into our testing OR we should serve that state in our app code

pkg/postgresql/database/core/database.go
DmytroPI-dev: Removed.

pkg/postgresql/database/core/database_unit_test.go
```diff
@@ -0,0 +1,1650 @@
...
+	}
+	databases := []enterprisev4.DatabaseDefinition{{Name: "payments"}}
+
+	t.Run("deleteCNPGDatabases removes existing object", func(t *testing.T) {
+		db := &cnpgv1.Database{ObjectMeta: metav1.ObjectMeta{Name: "primary-payments", Namespace: "dbs"}}
+		c := testClient(t, scheme, db)
+
+		require.NoError(t, deleteCNPGDatabases(context.Background(), c, postgresDB, databases))
+
+		err := c.Get(context.Background(), types.NamespacedName{Name: db.Name, Namespace: db.Namespace}, &cnpgv1.Database{})
```
mploski: do we need this? Maybe simple checking if function dont return errors is enough? the same way you do in deleteConfigMaps and others

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: Fixed this

pkg/postgresql/database/core/database_unit_test.go
DmytroPI-dev: made all three (deleteSecrets, deleteCNPGDatabases, deleteConfigMaps) to do the same checks.

mploski (approved):

PR #1795, opened 2026-03-26
docs: restore missing 3.0.0 Helm chart entries in index.yaml

PhilipSchmid:
Upstream commit 38373ce accidentally dropped the 3.0.0 entries for both splunk-operator and splunk-enterprise from [docs/index.yaml](https://splunk.github.io/splunk-operator/index.yaml) while publishing 3.1.0, which breaks existing installation automations that refer to 3.0.0. The 3.0.0 chart tarballs are still available at their direct URLs (e.g., https://splunk.github.io/splunk-operator/splunk-operator-3.0.0.tgz), so the index should reference them.

Restored entries:
- splunk-operator 3.0.0 (digest bd318b1f, URL splunk-operator-3.0.0.tgz)
- splunk-enterprise 3.0.0 (3 entries with digests ae82f6c8, 9a4133ae, db8d639d, all pointing to splunk-enterprise-3.0.0.tgz)

### DIFF
--- a/docs/index.yaml
+++ b/docs/index.yaml
@@ -56,6 +56,88 @@ entries:
     - https://splunk.github.io/splunk-operator/splunk-enterprise-3.1.0.tgz
     version: 3.1.0
   - apiVersion: v2
+    appVersion: 3.0.0
+    created: "2025-09-22T12:48:31.589008-07:00"
+    dependencies:
+    - condition: splunk-operator.enabled
+      name: splunk-operator
+      repository: file://splunk-operator/helm-chart/splunk-operator
+      version: 3.0.0
+    description: A Helm chart for Splunk Enterprise managed by the Splunk Operator
+    digest: ae82f6c8edee4d827817fe6c9c6447c422a03c59595a0f6e779cef847a83b611
+    maintainers:
+    - email: vivekr@splunk.com
+      name: Vivek Reddy
+    - email: rlieberman@splunk.com
+      name: Raizel Lieberman
+    - email: patrykw@splunk.com
+      name: Patryk Wasielewski
+    - email: igorg@splunk.com
+      name: Igor Grzankowski
+    - email: kkoziol@splunk.com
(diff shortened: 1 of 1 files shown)


PhilipSchmid:
I have read the CLA Document and I hereby sign the CLA

PhilipSchmid:
I have read the Code of Conduct and I hereby sign the COC

docs/index.yaml
```diff
@@ -55,6 +55,88 @@ entries:
...
+    - email: igorg@splunk.com
+      name: Igor Grzankowski
+    - email: kkoziol@splunk.com
+      name: Kasia Kozioł
+    name: splunk-enterprise
+    type: application
+    urls:
+    - https://splunk.github.io/splunk-operator/splunk-enterprise-3.0.0.tgz
+    version: 3.0.0
+  - apiVersion: v2
```
vivekr-splunk, line 84: This adds multiple conflicting `splunk-enterprise` entries for version `3.0.0` that all point to the same chart URL. In the checked-out branch there are now three `3.0.0` entries under `entries.splunk-enterprise`; only the first digest (`ae82f6...`) matches `docs/splunk-enterprise-3.0.0.tgz`, while the later ones have different digests, and one of them even carries `appVersion: 2.8.1`. From a Helm repo consumer point of view, that leaves the index ambiguous and partially incorrect, so I don’t think this is safe to merge as-is.

vivekr-splunk (changes requested):
I found one blocking issue in the restored Helm repo metadata.

vivekr-splunk (commented):
This also looks duplicate/superseded by the later Helm 3.0.0 index-restore PRs in the repo. I would consolidate on a single corrected restore PR instead of merging multiple overlapping attempts in this area.

PhilipSchmid:
Ah, perfect, then let's leave it to the other PR from one of your colleagues to address it. 😄 

https://github.com/splunk/splunk-operator/pull/1832

PR #1796, opened 2026-03-26
Added needed infra setup + pgcluster integration test base

M4KIF:
### Description

- added a synchronous reconciler calls based integration test for the pgcluster controler
- checks expected state of: secrets, finaliser addtion, objet refs, spec drift, invalid setup.

### Key Changes

- pgcluster integration tests

### Testing and Verification

- tests run using `make test` out of the source root.

### Related Issues

### PR Checklist

- [x] All tests pass locally.

### DIFF
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018-2022 Splunk Inc. All rights reserved.
+Copyright 2026.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


internal/controller/suite_test.go
```diff
@@ -19,155 +19,140 @@ package controller
...
+	var err error
+	err = enterprisev4.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+	err = cnpgv1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:scheme
 
+	By("bootstrapping test environment")
+	cnpgCRDDirectory, err := getCNPGCRDDirectory()
```
mploski: we should not leak anything specific to cnpg to this file as cnpg is low level details.

mploski: why do we need this function?

internal/controller/suite_test.go
```diff
@@ -19,155 +19,140 @@ package controller
...
+			return filepath.Join(basePath, entry.Name())
+		}
 	}
-	if err := (&IndexerClusterReconciler{
-		Client: k8sManager.GetClient(),
-		Scheme: k8sManager.GetScheme(),
-	}).SetupWithManager(k8sManager); err != nil {
-		Expect(err).NotTo(HaveOccurred())
-	}
-	if err := (&IngestorClusterReconciler{
```
mploski, line 137: why we did removed it?

internal/controller/suite_test.go
```diff
@@ -1,5 +1,5 @@
 /*
```
mploski, line 1: we dont own this file I'd leave it alone or just make the simplest changes possible

internal/controller/suite_test.go
M4KIF: So If we need to setup the env for the pgcluster controller, ie. setup It's dependency we should try to place It into the cluster setup fixture. That maybe trivial, Is that possible? ie. isn't the state of the k8s api and test env decided before the we pass down to per test fixtures?
Ie. lets compare It to the database transaction, isn't the "state" transaction already committed when we would try and alter It? (state stabilised and we try to append cnpg "resource"?)

In general I agree and hope that's possible to do

M4KIF: tbh, I haven't yet tried to minimise It. And tbh, I didn't exactly understand what happens here. So preparing cngp requires us to have the path of the cnpg(?).
Which is available as a go module, and the method tries to retrieve the path to cnpg "CRD"'s(?) which could or couldn't be available in the GOPATH. And statements in this method try to be as universal as possible by trying to consider every possible way this module is placed in the filesystem.

M4KIF: after another round of reading I'm trying to use more domain specific language. Please correct me whenever I talk gibberish.

internal/controller/suite_test.go
M4KIF: my bad, was too focused on reaching sort of executability of the tests. Linked the cnpg and missed a change made.

limak9182 (approved):

DmytroPI-dev (approved):

DmytroPI-dev (commented):
Don't forget to stash commits 😸

PR #1797, opened 2026-03-26
add integration tests for database reconcilier

DmytroPI-dev:
### Description

_Integration test for postgresDatabase_

### Key Changes

_Created  `postgresdatabase_controller_test.go`_

### Testing and Verification

_This MR is an integration test itself_

### Related Issues

[CPI-1917](https://splunk.atlassian.net/browse/CPI-1917)

### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/internal/controller/postgresdatabase_controller_test.go
+++ b/internal/controller/postgresdatabase_controller_test.go
@@ -18,67 +18,539 @@ package controller
 
 import (
 	"context"
+	"fmt"
+	"slices"
+	"time"
 
+	cnpgv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	enterprisev4 "github.com/splunk/splunk-operator/api/v4"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-var _ = Describe("Database Controller", func() {
-	Context("When reconciling a resource", func() {
-		const resourceName = "test-resource"
+const postgresDatabaseFinalizer = "postgresdatabases.enterprise.splunk.com/finalizer"
 
-		ctx := context.Background()
+func reconcilePostgresDatabase(ctx context.Context, nn types.NamespacedName) (ctrl.Result, error) {
+	reconciler := &PostgresDatabaseReconciler{
+		Client: k8sClient,
+		Scheme: k8sClient.Scheme(),
+	}
+	return reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: nn})
+}
 
-		typeNamespacedName := types.NamespacedName{
-			Name:      resourceName,
-			Namespace: "default", // TODO(user):Modify as needed
-		}
-		database := &enterprisev4.PostgresDatabase{}
+func managedRoleNames(roles []enterprisev4.ManagedRole) []string {
+	names := make([]string, 0, len(roles))
+	for _, role := range roles {
+		names = append(names, role.Name)
+	}
+	return names
+}
 
-		BeforeEach(func() {
-			By("creating the custom resource for the Kind Database")
-			err := k8sClient.Get(ctx, typeNamespacedName, database)
-			if err != nil && errors.IsNotFound(err) {
-				resource := &enterprisev4.PostgresDatabase{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      resourceName,
-						Namespace: "default",
-					},
-					// TODO(user): Specify other spec details if needed.
-				}
-				Expect(k8sClient.Create(ctx, resource)).To(Succeed())
-			}
-		})
+func adminRoleNameForTest(dbName string) string {
+	return dbName + "_admin"
+}
 
-		AfterEach(func() {
-			// TODO(user): Cleanup logic after each test, like removing the resource instance.
-			resource := &enterprisev4.PostgresDatabase{}
-			err := k8sClient.Get(ctx, typeNamespacedName, resource)
-			Expect(err).NotTo(HaveOccurred())
+func rwRoleNameForTest(dbName string) string {
+	return dbName + "_rw"
+}
 
-			By("Cleanup the specific resource instance Database")
-			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
-		})
-		It("should successfully reconcile the resource", func() {
-			By("Reconciling the created resource")
-			controllerReconciler := &PostgresDatabaseReconciler{
-				Client: k8sClient,
-				Scheme: k8sClient.Scheme(),
-			}
+func ownedByPostgresDatabase(postgresDB *enterprisev4.PostgresDatabase) []metav1.OwnerReference {
+	controller := true
+	blockOwnerDeletion := true
+	return []metav1.OwnerReference{{
+		APIVersion:         enterprisev4.GroupVersion.String(),
+		Kind:               "PostgresDatabase",
+		Name:               postgresDB.Name,
+		UID:                postgresDB.UID,
+		Controller:         &controller,
+		BlockOwnerDeletion: &blockOwnerDeletion,
+	}}
+}
 
-			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
-				NamespacedName: typeNamespacedName,
+func createPostgresDatabaseResource(ctx context.Context, namespace, resourceName, clusterName string, databases []enterprisev4.DatabaseDefinition, finalizers ...string) *enterprisev4.PostgresDatabase {
+	postgresDB := &enterprisev4.PostgresDatabase{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       resourceName,
+			Namespace:  namespace,
+			Finalizers: finalizers,
+		},
+		Spec: enterprisev4.PostgresDatabaseSpec{
+			ClusterRef: corev1.LocalObjectReference{Name: clusterName},
+			Databases:  databases,
+		},
+	}
+	Expect(k8sClient.Create(ctx, postgresDB)).To(Succeed())
+	return postgresDB
+}
+
+func createPostgresClusterResource(ctx context.Context, namespace, clusterName string) *enterprisev4.PostgresCluster {
+	postgresCluster := &enterprisev4.PostgresCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterName,
+			Namespace: namespace,
+		},
+		Spec: enterprisev4.PostgresClusterSpec{
+			Class: "dev",
+		},
+	}
+	Expect(k8sClient.Create(ctx, postgresCluster)).To(Succeed())
+	return postgresCluster
+}
+
+func markPostgresClusterReady(ctx context.Context, postgresCluster *enterprisev4.PostgresCluster, cnpgClusterName, namespace string, poolerEnabled bool) {
+	clusterPhase := "Ready"
+	postgresCluster.Status.Phase = &clusterPhase
+	postgresCluster.Status.ProvisionerRef = &corev1.ObjectReference{
+		APIVersion: cnpgv1.SchemeGroupVersion.String(),
+		Kind:       "Cluster",
+		Name:       cnpgClusterName,
+		Namespace:  namespace,
+	}
+	if poolerEnabled {
+		postgresCluster.Status.ConnectionPoolerStatus = &enterprisev4.ConnectionPoolerStatus{Enabled: true}
+	}
+	Expect(k8sClient.Status().Update(ctx, postgresCluster)).To(Succeed())
+}
+
+func createCNPGClusterResource(ctx context.Context, namespace, cnpgClusterName string) *cnpgv1.Cluster {
+	cnpgCluster := &cnpgv1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cnpgClusterName,
+			Namespace: namespace,
+		},
+		Spec: cnpgv1.ClusterSpec{
+			Instances: 1,
+			StorageConfiguration: cnpgv1.StorageConfiguration{
+				Size: "1Gi",
+			},
+		},
+	}
+	Expect(k8sClient.Create(ctx, cnpgCluster)).To(Succeed())
+	return cnpgCluster
+}
+
+func markCNPGClusterReady(ctx context.Context, cnpgCluster *cnpgv1.Cluster, reconciledRoles []string, writeService, readService string) {
+	cnpgCluster.Status.ManagedRolesStatus = cnpgv1.ManagedRoles{
+		ByStatus: map[cnpgv1.RoleStatus][]string{
+			cnpgv1.RoleStatusReconciled: reconciledRoles,
+		},
+	}
+	cnpgCluster.Status.WriteService = writeService
+	cnpgCluster.Status.ReadService = readService
+	Expect(k8sClient.Status().Update(ctx, cnpgCluster)).To(Succeed())
+}
+
+type readyClusterScenario struct {
+	namespace       string
+	resourceName    string
+	clusterName     string
+	cnpgClusterName string
+	dbName          string
+	requestName     types.NamespacedName
+}
+
+func newReadyClusterScenario(namespace, resourceName, clusterName, cnpgClusterName, dbName string) readyClusterScenario {
+	return readyClusterScenario{
+		namespace:       namespace,
+		resourceName:    resourceName,
+		clusterName:     clusterName,
+		cnpgClusterName: cnpgClusterName,
+		dbName:          dbName,
+		requestName:     types.NamespacedName{Name: resourceName, Namespace: namespace},
+	}
+}
+
+func seedReadyClusterScenario(ctx context.Context, scenario readyClusterScenario, poolerEnabled bool) {
+	createPostgresDatabaseResource(ctx, scenario.namespace, scenario.resourceName, scenario.clusterName, []enterprisev4.DatabaseDefinition{{Name: scenario.dbName}})
+	postgresCluster := createPostgresClusterResource(ctx, scenario.namespace, scenario.clusterName)
+	markPostgresClusterReady(ctx, postgresCluster, scenario.cnpgClusterName, scenario.namespace, poolerEnabled)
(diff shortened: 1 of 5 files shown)


pkg/postgresql/database/core/database-controller-integration_test-plan.md
```diff
@@ -0,0 +1,337 @@
+# PostgresDatabase Controller Integration Test Plan
```
limak9182: do we want this MD file to be committed?

DmytroPI-dev: I thought it might be helpful, no?

pkg/postgresql/database/core/database.go
```diff
@@ -705,8 +705,8 @@ func buildManagedRoles(postgresDBName string, databases []enterprisev4.DatabaseD
 func buildManagedRolesPatch(cluster *enterprisev4.PostgresCluster, roles []enterprisev4.ManagedRole) *unstructured.Unstructured {
 	return &unstructured.Unstructured{
 		Object: map[string]any{
-			"apiVersion": cluster.APIVersion,
-			"kind":       cluster.Kind,
+			"apiVersion": enterprisev4.GroupVersion.String(),
+			"kind":       "PostgresCluster",
```
limak9182: hmm I don't fully understand this change, why it's preferred to `cluster.Kind` hardcoded? 🤔

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,591 @@ package controller
...
+		result, err := reconcilePostgresDatabase(ctx, requestName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(result).To(Equal(ctrl.Result{}))
+
+		current := &enterprisev4.PostgresDatabase{}
+		Expect(k8sClient.Get(ctx, requestName, current)).To(Succeed())
+		current.Status.Databases = []enterprisev4.DatabaseInfo{{Name: dbName}}
+		Expect(k8sClient.Status().Update(ctx, current)).To(Succeed())
+
+		result, err = reconcilePostgresDatabase(ctx, requestName)
```
limak9182: Just an open question. I see we are calling reconcile function in sequence. Is it possible to test the whole workflow when the controller re-enters the reconciliation and makes progress when the requeue fires. That would test the whole process and prove we are requeueing correctly.

pkg/postgresql/database/core/database-controller-integration_test-plan.md
DmytroPI-dev: Removed to Confluence page.

pkg/postgresql/database/core/database.go
DmytroPI-dev: Integration tests drops the following error: 
```
github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3
        /Users/dpishchenkov/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.28.1/internal/suite.go:946
  [FAILED] in [It] - /Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:596 @ 03/27/26 11:01:00.73
• [FAILED] [0.022 seconds]
PostgresDatabase Controller [It] adds pooler endpoints to the generated ConfigMap when connection pooling is enabled
/Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:521

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0x140011f9d20>: 
      patching managed roles for PostgresDatabase pooler-cluster: Object 'Kind' is missing in 'unstructured object has no kind'
      {
          msg: "patching managed roles for PostgresDatabase pooler-cluster: Object 'Kind' is missing in 'unstructured object has no kind'",
          err: <*runtime.missingKindErr | 0x140009494f0>{
              data: "unstructured object has no kind",
          },
      }
  occurred
  In [It] at: /Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:596 @ 03/27/26 11:01:00.73

  Full Stack Trace
  
  
Summarizing 3 Failures:
  [FAIL] PostgresDatabase Controller [It] reconciles Kubernetes resources for a ready cluster without invoking live grants
  /Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:223
  [FAIL] PostgresDatabase Controller [It] orphan-retains selected resources, deletes others, and patches managed roles on deletion
  /Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:474
  [FAIL] PostgresDatabase Controller [It] adds pooler endpoints to the generated ConfigMap when connection pooling is enabled
  /Users/dpishchenkov/splunk-operator/internal/controller/postgresdatabase_controller_test.go:596

Ran 5 of 56 Specs in 26.554 seconds
```

DmytroPI-dev: Fixed this.

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Added full-sequence tests.

limak9182: Great, thank you.

pkg/postgresql/database/core/database.go
limak9182: Awesome, great, closing this thread then.

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+			continue
+		}
+		if ref.Controller != nil && *ref.Controller {
+			return true
+		}
+	}
+	return false
+}
+
+func namespacedName(namespace, name string) types.NamespacedName {
```
mploski: I think this function is redundant it takes the same number of params as function you wrapped.

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+	}
+	return reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: nn})
+}
+
+func controlledBy(owner metav1.Object, obj metav1.Object) bool {
+	for _, ref := range obj.GetOwnerReferences() {
+		if ref.UID != owner.GetUID() {
+			continue
+		}
+		if ref.Controller != nil && *ref.Controller {
```
mploski: Can we use metav1.GetControllerOf(obj)/metav1.IsControlledBy(obj, owner) to verify ownership? Why do we need custom function?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+		APIVersion:         enterprisev4.GroupVersion.String(),
+		Kind:               "PostgresDatabase",
+		Name:               postgresDB.Name,
+		UID:                postgresDB.UID,
+		Controller:         &controller,
+		BlockOwnerDeletion: &blockOwnerDeletion,
+	}}
+}
+
+func startPostgresDatabaseManager(ctx context.Context) {
```
mploski: lets not use Manager if we dont need to - it cause our test to be non deterministic as we dont control manager once it is triggered and what events it starts to consume - so we might have concurrency problems, can't we run reconciller explicitly without it?

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Removed

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Fixed.

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Removed, fixed.

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+		errCh <- mgr.Start(managerCtx)
+	}()
+
+	DeferCleanup(func() {
+		cancel()
+		Eventually(errCh, time.Second).Should(Receive(BeNil()))
+	})
+}
+
+var _ = Describe("PostgresDatabase Controller", func() {
```
mploski, line 345: can we follow the structure explained here:
https://github.com/onsi/ginkgo i.e when -> context -> it? It is then much easier to read those tests, and debug:

take a look also on: https://onsi.github.io/ginkgo/#organizing-specs-with-container-nodes

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+
+		result, err := reconcilePostgresDatabase(ctx, requestName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(result).To(Equal(ctrl.Result{}))
+
+		current := &enterprisev4.PostgresDatabase{}
+		Expect(k8sClient.Get(ctx, requestName, current)).To(Succeed())
+		Expect(current.Finalizers).To(ContainElement(postgresDatabaseFinalizer))
+
+		result, err = reconcilePostgresDatabase(ctx, requestName)
```
mploski: what is the scope of this test? in it we do: adds a finalizer before processing a missing cluster but here we do a second reconcillation,  check that requeue after is set to 30 seconds, then we fetch the current once again and check phases and statuses. We should probably follow the context, when and it pattern and make this more targeted ( this is general comment)

Also, why do we do 2 object fetch? Maybe if you expect some specific condition at the end we should rely on `eventually` assert?

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: changed to use when -> context -> It structure.

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+	})
+
+	It("reconciles Kubernetes resources for a ready cluster without invoking live grants", func() {
+		resourceName := "ready-cluster"
+		clusterName := "tenant-cluster"
+		cnpgClusterName := "tenant-cnpg"
+		dbName := "appdb"
+		requestName := namespacedName(namespace, resourceName)
+
+		postgresDB := &enterprisev4.PostgresDatabase{
```
mploski: I see that in every test we create Postgres Daabase, cluster, set the needed phases to prepare environment so we start from expected state. maybe we can offload it to the helper function to lower down the repetition in every test?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+		Expect(cnpgDatabase.Spec.Name).To(Equal(dbName))
+		Expect(cnpgDatabase.Spec.Owner).To(Equal("appdb_admin"))
+		Expect(cnpgDatabase.Spec.ClusterRef.Name).To(Equal(cnpgClusterName))
+		Expect(controlledBy(current, cnpgDatabase)).To(BeTrue())
+
+		applied := true
+		cnpgDatabase.Status.Applied = &applied
+		Expect(k8sClient.Status().Update(ctx, cnpgDatabase)).To(Succeed())
+
+		result, err = reconcilePostgresDatabase(ctx, requestName)
```
mploski: Can we split those 4 stages via ordered container https://onsi.github.io/ginkgo/#ordered-containers?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,749 @@ package controller
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(result).To(Equal(ctrl.Result{}))
+
+		Expect(k8sClient.Get(ctx, requestName, current)).To(Succeed())
+		Expect(current.Status.Phase).NotTo(BeNil())
```
mploski: I'd suggest instead of checking every attribute one after another, prepare few simple constructs what you expect from every reconciller ( depends on the input CR ofcourse) and compare those attributes with current.Status

internal/controller/suite_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018-2022 Splunk Inc. All rights reserved.
+Copyright 2026.
```
mploski: we shouldnt modify existing file Copyright

internal/controller/suite_test.go
```diff
@@ -61,8 +72,12 @@ var _ = BeforeSuite(func(ctx context.Context) {
 
 	By("bootstrapping test environment")
 
+	cnpgModuleDir := resolveCNPGModuleDir()
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
+		CRDDirectoryPaths: []string{
+			filepath.Join("..", "..", "config", "crd", "bases"),
+			filepath.Join(cnpgModuleDir, "config", "crd", "bases"),
```
mploski, line 79: I wonder why we need to have cnpg crd's deployed like that? Can't it work without it?

pkg/postgresql/database/core/database.go
```diff
@@ -702,20 +707,27 @@ func buildManagedRoles(postgresDBName string, databases []enterprisev4.DatabaseD
 	return roles
 }
 
-func buildManagedRolesPatch(cluster *enterprisev4.PostgresCluster, roles []enterprisev4.ManagedRole) *unstructured.Unstructured {
+func buildManagedRolesPatch(cluster *enterprisev4.PostgresCluster, roles []enterprisev4.ManagedRole, scheme *runtime.Scheme) (*unstructured.Unstructured, error) {
+	gvk, err := apiutil.GVKForObject(cluster, scheme)
```
mploski, line 719: I wonder what problem does it solve?

DmytroPI-dev: this problem: https://github.com/splunk/splunk-operator/pull/1797#discussion_r2999898804

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Reason:

reconcile 1 adds the finalizer and returns
reconcile 2 creates secrets/configmaps and patches managed roles, then requeues
reconcile 3 creates the CNPG Database, then requeues because it is not yet applied
reconcile 4 finishes status after you mark cnpgDatabase.Status.Applied = true

AI recommends the following: 
1. These 4 stages are not independent specs, they are one scenario progressing through mutable shared state
2. Ginkgo explicitly warns that Ordered containers are an exception to normal spec independence and should be used only when needed
3. if one stage fails, later stages in the ordered container are skipped by default
4. it increases coupling and makes the suite more vulnerable to spec-pollution style issues
Now we have: 
one scenario = one spec
deterministic debugging
clearer failure output per stage
no ordered-spec coupling

Does this makes sense?

internal/controller/suite_test.go
DmytroPI-dev: As tests use real CNPG CRs like Cluster and Database and `envtest` runs a K8s API server, it will  accept those resources if their CRDs are installed, so we need the CNPG CRDs must also be loaded into envtest

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Add helpers

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Add  status assertion helpers  for:

- expectStatusPhase(...)
- expectStatusCondition(...)
- expectReadyStatus(...)

the happy-path final status check now uses `expectReadyStatus(...)`
the missing-cluster and role-conflict cases now use the smaller status helpers instead of repeating all checks inline.

internal/controller/postgresdatabase_controller_test.go
mploski: good points, lets not incorporate those tradeoffs and instead write repeatable code inside of helpers so we increase test readability

pkg/postgresql/database/core/database.go
mploski: out of curiosity, do we know why Kind was empty? It suppose to take value from Cluster. Do we uderstand the root cause?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,522 @@ package controller
...
+		conditionType string
+		reason        string
+	}{
+		{conditionType: "ClusterReady", reason: "ClusterAvailable"},
+		{conditionType: "SecretsReady", reason: "SecretsCreated"},
+		{conditionType: "ConfigMapsReady", reason: "ConfigMapsCreated"},
+		{conditionType: "RolesReady", reason: "UsersAvailable"},
+		{conditionType: "DatabasesReady", reason: "DatabasesAvailable"},
+	} {
+		expectStatusCondition(current, expected.conditionType, metav1.ConditionTrue, expected.reason)
```
mploski: what this line and 180 is used for? Im not sure we should hide the map of states here - can't we simply pass the enum to this function what we expect from conditionType and reason perspective? In fact only ConditionType is enough. In fact I see we use this function once, so maybe we can rid of it and use expectStatusCondition directly?

pkg/postgresql/database/core/database.go
DmytroPI-dev: Root case analysis: 
Root cause: `Unstructured` does not infer `kind` from the Go type. It only reads it from its internal `Object` map.

In your failing path in [database.go](/Users/dpishchenkov/splunk-operator/pkg/postgresql/database/core/database.go#L712), you built the patch like this:

```go
Object: map[string]any{
    "apiVersion": cluster.APIVersion,
    "kind":       cluster.Kind,
    ...
}
```

The problem is that `cluster` there is a typed `*enterprisev4.PostgresCluster` fetched from the API. After a normal `Get(...)`, its `TypeMeta` fields (`APIVersion`, `Kind`) are often empty or not reliable.

So what happened was:

1. `cluster.APIVersion == ""`
2. `cluster.Kind == ""`
3. you copied those empty values into `unstructured.Unstructured.Object`
4. then server-side apply tried to use that object as a patch
5. the patch payload had no valid `kind`
6. Kubernetes returned:
   - `Object 'Kind' is missing in 'unstructured object has no kind'`

Why `Unstructured` did not save you:
- `GetObjectKind()` just returns the object itself
- `GroupVersionKind()` on `Unstructured` reads:
  - `u.GetAPIVersion()`
  - `u.GetKind()`
- those in turn come from the `Object` map
- so if the map contains empty or missing `apiVersion` / `kind`, `GroupVersionKind()` is also empty

In other words:
- `Unstructured` implements `ObjectKind`
- but it is only a wrapper around the data you put into `Object`
- it does not derive GVK from the Go type automatically

So the real bug was not in `Unstructured`.
It was here:
- taking GVK from `cluster.APIVersion` / `cluster.Kind`
- instead of from a stable source like:
  - the scheme via `apiutil.GVKForObject(...)`
  - or `enterprisev4.GroupVersion.WithKind("PostgresCluster")`

Why tests exposed it:
- the patching path definitely ran in `envtest`
- the fetched typed object had empty `TypeMeta`
- so the bug became deterministic there

So the short version is:

- typed objects: GVK can come from scheme / decoding, but `TypeMeta` is not reliable after `Get(...)`
- unstructured objects: GVK comes only from `"apiVersion"` and `"kind"` in the map
- you copied empty `TypeMeta` into that map
- SSA rejected the patch because it had no valid `kind`

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -18,67 +18,522 @@ package controller
...
 			})
-			Expect(err).NotTo(HaveOccurred())
-			// TODO(user): Add more specific assertions depending on your controller's reconciliation logic.
-			// Example: If you expect a certain status condition after reconciliation, verify it here.
 		})
 	})
 })
+
+func containsFinalizer(finalizers []string, target string) bool {
+	return slices.Contains(finalizers, target)
```
mploski: why not use Contains directly?

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Add helpers for repeatable code.

mploski (approved):

PR #1799, opened 2026-03-27
  fix(postgres): address PR review — API validation, type cleanup, controller correctness

mploski:
### Description
Addresses  issues raised in code review against the postgres cluster, cluster class, and database controllers.

##### ConnectionPoolerConfig removed from PostgresClusterSpec (postgrescluster_types.go)

  ConnectionPoolerConfig was dead code — the field was declared on PostgresClusterSpec but never read by the reconciler, which exclusively uses the class-level config. Removed
  the field and its DeepCopyInto generated code to eliminate the misleading API surface.

##### DatabaseDefinition.Name pattern validation (postgresdatabase_types.go)

  Added +kubebuilder:validation:Pattern=^[a-z_][a-z0-9_]*$ and MinLength=1. Without this, names such as "123db" or "my-db" passed admission but failed at the PostgreSQL layer
  with an opaque SQL error. Validation now enforced at the API boundary.

##### PostgresClusterClass immutability enforced (postgresclusterclass_types.go)

  Added +kubebuilder:validation:XValidation:rule="self == oldSelf" to PostgresClusterClassSpec. Classes serve as immutable provisioning templates — previously nothing prevented
  mutation after clusters were already referencing a class.

##### JSON tag inconsistency fixed: ConfigMapRef (postgresdatabase_types.go)

  DatabaseInfo.ConfigMapRef was serialising as "configMap" due to a mismatched json tag. Corrected to "configMapRef" for consistency with all other ref fields. This is a
  breaking wire-format change for any existing stored objects using this field.

##### Typo fixed: PosgresClusterClassConfig (postgresclusterclass_types.go)

  Renamed to PostgresClusterClassConfig. The transposed s and g had been propagated into the generated CRD and deep copy code.

##### Name length constraint added to PostgresCluster and PostgresDatabase (postgrescluster_types.go, postgresdatabase_types.go)

  Added +kubebuilder:validation:XValidation:rule="size(self.metadata.name) <= 50" to both types to prevent derived resource name overflow (secrets, configmaps, poolers all
  append suffixes to the CR name).

  ---
  #### CRD & Config Fixes

##### Broken print columns in PostgresClusterClass CRD

  Column JSONPaths referenced spec.postgresClusterConfig.* which no longer existed after the field rename to spec.config.*. Updated all three columns (Instances, Storage,
  Version). CRDs regenerated.

##### Database CRDs missing from preserveUnknownFields patch (patch_preserve_unknown_fields.yaml)

  Added postgresclusters, postgresclusterclasses, postgresdatabases to the existing patch for consistency with all other operator-managed CRDs. No functional change —
  apiextensions.k8s.io/v1 defaults to false — but aligns new types with the established convention.

##### Samples kustomization referenced non-existent files

  Removed stale file references in the samples kustomization that were pointing to files removed in a prior refactor.

  ---
  #### Controller Fixes

##### poolerExists errors propagated (cluster.go)

  Changed poolerExists return type from bool to (bool, error). Transient API errors were previously silently treated as "pooler not found", causing the controller to incorrectly
   attempt recreation. Updated all three call sites — main service loop errors joined with errors.Join, generateConfigMap propagates as (nil, err).

##### Secret deletion no longer breaks retained CNPG clusters (cluster.go)

  When ClusterDeletionPolicy: Retain is set, the finalizer handler now removes the owner reference from the superuser secret before removing the finalizer. Previously, the
  secret was deleted via garbage collection when the PostgresCluster was removed, leaving the retained CNPG cluster without credentials.

##### Conflict handling for finalizer updates in PostgresDatabase

  The finalizer update path in the PostgresDatabase reconciler was not handling apierrors.IsConflict. Added Requeue: true response consistent with the existing pattern in
  PostgresCluster.

##### setStatus skips write when nothing changed (cluster.go)

  Added a DeepEqual guard: snapshots cluster.Status before applying mutations and skips Status().Update() when the result is semantically identical. Eliminates redundant API
  writes on stable healthy clusters. Uses equality.Semantic.DeepEqual to correctly handle Kubernetes type semantics (resource.Quantity normalisation, nil vs empty map).
  Mid-reconcile status checkpoints are preserved, the guard only prevents no-op writes, not intentional state transitions.

  ---
### Testing and Verification

### Related Issues

### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/internal/controller/postgrescluster_controller.go
+++ b/internal/controller/postgrescluster_controller.go
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
@@ -61,6 +60,7 @@ func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 // SetupWithManager registers the controller and owned resource watches.
 func (r *PostgresClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		WithEventFilter(predicate.Funcs{GenericFunc: func(event.GenericEvent) bool { return false }}).
 		For(&enterprisev4.PostgresCluster{}, builder.WithPredicates(postgresClusterPredicator())).
 		Owns(&cnpgv1.Cluster{}, builder.WithPredicates(cnpgClusterPredicator())).
 		Owns(&cnpgv1.Pooler{}, builder.WithPredicates(cnpgPoolerPredicator())).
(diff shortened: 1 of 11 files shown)


internal/controller/postgrescluster_controller.go
```diff
@@ -73,107 +73,71 @@ func (r *PostgresClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	return !equality.Semantic.DeepEqual(oldObj.GetOwnerReferences(), newObj.GetOwnerReferences())
 }
 
-// postgresClusterPredicator triggers on generation changes, deletion, and finalizer transitions.
+// postgresClusterPredicator triggers on spec changes, deletion, and finalizer transitions.
 func postgresClusterPredicator() predicate.Predicate {
-	return predicate.Funcs{
-		CreateFunc: func(event.CreateEvent) bool { return true },
-		DeleteFunc: func(event.DeleteEvent) bool { return true },
-		UpdateFunc: func(e event.UpdateEvent) bool {
```
DmytroPI-dev, line 89: You've simplified predicators a lot, looks great!

DmytroPI-dev (approved):

limak9182 (approved):

PR #1808, opened 2026-04-01
Add PostgreSQL observability telemetry exposure

DmytroPI-dev:
### Description

Adds PostgreSQL observability telemetry for `PostgresCluster` using Prometheus pod-annotation-based scraping. Metrics are exposed by CNPG's built-in exporters on PostgreSQL pods (port `9187`) and PgBouncer pooler pods (port `9127`). The operator controls whether annotations are injected via class- and cluster-level configuration, with no dedicated metrics `Service` or `ServiceMonitor` required for PostgreSQL or PgBouncer scraping.

A `ServiceMonitor` is still supported for operator-controller metrics as an optional step.

### Key Changes

`api/v4/postgresclusterclass_types.go`  
Added class-level observability configuration (`monitoring.postgresqlMetrics.enabled`, `monitoring.connectionPoolerMetrics.enabled`) that controls whether scrape annotations are injected into CNPG pods.

`api/v4/postgrescluster_types.go`  
Added cluster-level disable-only overrides (`spec.monitoring.postgresqlMetrics.disabled`, `spec.monitoring.connectionPoolerMetrics.disabled`) allowing per-cluster opt-out without changing the class.

`pkg/postgresql/cluster/core/cluster.go`  
Wired observability flag resolution into `PostgresCluster` reconciliation. When enabled, sets `InheritedMetadata.Annotations` on the CNPG `Cluster` (for PostgreSQL pods) and `Template.ObjectMeta.Annotations` on CNPG `Pooler` resources (for PgBouncer pods).

`pkg/postgresql/cluster/core/monitoring.go`  
Added `isPostgreSQLMetricsEnabled` / `isConnectionPoolerMetricsEnabled` flag resolution helpers.  
Added `buildPostgresScrapeAnnotations` / `buildPoolerScrapeAnnotations` annotation builders.  
Added `removeScrapeAnnotations` for the disable path.

`pkg/postgresql/cluster/core/monitoring_unit_test.go`  
Added unit tests for flag resolution, scrape annotation builders, and annotation removal.

`internal/controller/postgrescluster_controller_test.go`  
Added integration tests verifying that `InheritedMetadata` annotations are set on the CNPG `Cluster` when monitoring is enabled and removed when disabled by cluster override.

`docs/PostgreSQLObservabilityDashboard.json`  
Reference Grafana dashboard covering PostgreSQL target count, RW/RO PgBouncer availability, WAL activity, database sizes, PgBouncer client load, controller reconcile metrics, and domain fleet metrics.

`docs/postgresSQLMonitoring-e2e.md`  
End-to-end validation guide for the annotation-based scraping flow on KIND.

### Testing and Verification

Added unit tests in `pkg/postgresql/cluster/core/monitoring_unit_test.go` for:
- class/cluster observability enablement logic
- scrape annotation builders for PostgreSQL (port `9187`) and PgBouncer (port `9127`)
- annotation removal on the disable path

Added integration tests in `internal/controller/postgrescluster_controller_test.go` verifying:
- `InheritedMetadata.Annotations` presence when monitoring is enabled
- annotation removal when disabled by cluster-level override

### Related Issues

[CPI-1853](https://splunk.atlassian.net/browse/CPI-1853) — related JIRA ticket.

### Grafana screenshot:
Screenshot 2026-04-15 at 14 33 08

### PR Checklist
- [X] All tests pass locally.


docs/PostgreSQLObservabilityDashboard.md
```diff
@@ -0,0 +1,65 @@
+# PostgreSQL Observability Dashboard Example
+
+This file provides a reference Grafana dashboard for the PostgreSQL observability model described in the PostgreSQL observability notes.
+
+The dashboard JSON lives at:
+
+- [PostgreSQLObservabilityDashboard.json](/Users/dpishchenkov/splunk-operator/docs/PostgreSQLObservabilityDashboard.json)
```
limak9182: It links to your personal file.

pkg/postgresql/cluster/core/cluster.go
limak9182, line 260: nit: should we also use here your `isConnectionPoolerMetricsEnabled` to not have same logic to maintain in 2 places?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -396,6 +396,37 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 		rc.emitPoolerReadyTransition(postgresCluster, oldConditions)
 	}
 
+	if err := reconcilePostgreSQLMetricsService(ctx, c, rc.Scheme, postgresCluster, isPostgreSQLMetricsEnabled(postgresCluster, clusterClass)); err != nil {
```
limak9182: I'm just wondering if placing it after Configmap is created and Everything is ready? At the very end of reconciliation function? That way we would make sure that all setup is ready before we start scraping metrics. It sounds also more natural to me, what are your thoughts?

docs/PostgreSQLObservabilityDashboard.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Purpose
+
+This dashboard is a reference artifact only.
```
limak9182, line 11: could you please provide a relative path to the dashboard file? It'll get lost in between all docs

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,483 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// metrics
+	postgresMetricsServiceSuffix = "-postgres-metrics"
+	postgresMetricsPortName      = "metrics"
+	postgresMetricsPort          = int32(9187)
```
limak9182: is it a standard port for metrics? I believe splunk operator service monitor uses a different one, maybe we should stay consistent?

pkg/postgresql/cluster/core/cluster.go
limak9182: In `buildCNPGClusterSpec()` does not set Spec.Monitoring on the CNPG Cluster object. Same for `buildCNPGPooler()`. Is it our decision or just overlook?

Maybe that would give us some interesting metrics we could use:
https://cloudnative-pg.io/docs/1.29/cloudnative-pg.v1#monitoringconfiguration

pkg/postgresql/cluster/core/monitoring.go
DmytroPI-dev: Controller-manager uses `https:8443`, but I assume we should use `9127` and `9187`, as per [docs](https://cloudnative-pg.io/docs/1.29/monitoring/).

pkg/postgresql/cluster/core/cluster.go
DmytroPI-dev: CNPG already exposes standard PostgreSQL metrics on port 9187 (metrics), so Cluster.spec.monitoring is not needed for basic scraping, I assume. It might be useful for use later for extra control such as custom queries, disabling default queries, query TTL, or TLS on the metrics endpoint.

DmytroPI-dev: This line, as I understand, decides whether the poolers themselves should exist. On contrary, `isConnectionPoolerMetricsEnabled(...)` answers a different question: whether pooler metrics should be exposed, which depends on both:
- if poolers are  enabled
- if observability is enabled for PgBouncer

pkg/postgresql/cluster/core/cluster.go
DmytroPI-dev: I don't think so, as this will not give us  any benefit, because the metrics resources do not depend on the CM at all. In our case, CM is just a connection metadata for consumers, moving observability after CM would make the reconcile flow less clear: 
- cluster and poolers are the producers of metrics
- observability resources belong with those producers
- ConfigMap is a separate output artifact

pkg/postgresql/cluster/core/cluster.go
limak9182: make sense, thank you for explaining that

pkg/postgresql/cluster/core/monitoring.go
limak9182: make sense

pkg/postgresql/cluster/core/cluster.go
limak9182: your points are valid, I get it now

pkg/postgresql/cluster/core/cluster.go
limak9182: I get it, thanks

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -396,6 +396,37 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		ctx, c, rc.Scheme, postgresCluster, readWriteEndpoint, rwPoolerMetricsEnabled,
+	); err != nil {
+		return ctrl.Result{}, err
+	}
+	if err := reconcileConnectionPoolerMetricsServiceMonitor(
+		ctx, c, rc.Scheme, postgresCluster, readOnlyEndpoint, roPoolerMetricsEnabled,
+	); err != nil {
+		return ctrl.Result{}, err
+	}
+
```
limak9182: It's another block for our reconciliation metric, maybe it's worth to emit event in case of success? or issue?

api/v4/postgrescluster_types.go
```diff
@@ -108,6 +108,27 @@ type PostgresClusterSpec struct {
...
 	// +optional
 	ClusterDeletionPolicy *string `json:"clusterDeletionPolicy,omitempty"`
+
+	// Observability contains configuration for metrics exposure features.
+	// +optional
+	Observability *PostgresObservabilityOverride `json:"observability,omitempty"`
+}
+
+// PostgresObservabilityOverride overrides observability configuration options for PostgresClusterClass.
+type PostgresObservabilityOverride struct {
```
mploski: PostgresObservabilityOverride  we should follow the same pattern we have for ConnectionPoolerEnabled
So maybe ConnectionPoolerMetricsEnabled and PostgreSQLMetricsEnabled?

api/v4/postgrescluster_types.go
```diff
@@ -108,6 +108,27 @@ type PostgresClusterSpec struct {
...
+
+	// +optional
+	PostgreSQL *FeatureDisableOverride `json:"postgresql,omitempty"`
+
+	// +optional
+	PgBouncer *FeatureDisableOverride `json:"pgbouncer,omitempty"`
```
mploski: in other provider we might not have pgbouncer ( aws for example) lets call it generic way ( connectionPooler). Also we should probably have CEL logic that doesnt allow connection pooler metrics enabled if connection pooler itself is disabled

api/v4/postgresclusterclass_types.go
```diff
@@ -99,6 +99,13 @@ type PostgresClusterClassConfig struct {
 	// +kubebuilder:default=false
 	// +optional
 	ConnectionPoolerEnabled *bool `json:"connectionPoolerEnabled,omitempty"`
+
+	// Observability contains configuration for metrics exposure.
+	// When enabled, creates metrics resources for clusters using this class.
+	// Can be overridden in PostgresCluster CR.
+	// +kubebuilder:default={}
+	// +optional
+	Observability *PostgresObservabilityClassConfig `json:"observability,omitempty"`
```
mploski: Similar to previous comment :-)

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,483 @@
...
+		return false
+	}
+	if cluster == nil || cluster.Spec.ConnectionPoolerEnabled == nil {
+		return true
+	}
+	return *cluster.Spec.ConnectionPoolerEnabled
+}
+
+func isConnectionPoolerMetricsEnabled(cluster *enterprisev4.PostgresCluster, class *enterprisev4.PostgresClusterClass) bool {
+	if !isConnectionPoolerEnabled(cluster, class) {
```
mploski: this check shouldnt be a part of this function I believe

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,483 @@
...
+		return false
+	}
+	if cluster == nil || cluster.Spec.Observability == nil || cluster.Spec.Observability.PostgreSQL == nil {
+		return true
+	}
+	override := cluster.Spec.Observability.PostgreSQL.Disabled
+	return override == nil || !*override
+}
+
+func isConnectionPoolerEnabled(cluster *enterprisev4.PostgresCluster, class *enterprisev4.PostgresClusterClass) bool {
```
mploski: should this function be a part of connection pooler not monitoring?

pkg/postgresql/cluster/core/cluster.go
mploski: also what about extending our status with information if this failed/succeeded i.e add new condition?

pkg/postgresql/cluster/core/monitoring.go
mploski: @DmytroPI-dev If I read correctly doc you shared, the port 9187 is internal port that postgresql instance manager use to expose prometheus metrics, pooler against 9127. cnpg Operator pod exposes their own metrics against port 8080.   Considering that Shouldn't we expose the common port for everything? At the end this will simplify scrapping via automated labels and we don't need to worry about specific different port for different things

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,483 @@
...
+		return false
+	}
+	if cluster == nil || cluster.Spec.Observability == nil || cluster.Spec.Observability.PgBouncer == nil {
+		return true
+	}
+	override := cluster.Spec.Observability.PgBouncer.Disabled
+	return override == nil || !*override
+}
+
+func buildPostgreSQLMetricsService(scheme *runtime.Scheme, cluster *enterprisev4.PostgresCluster) (*corev1.Service, error) {
```
mploski: Out of curiosity why we need to create k8s service to expose those information? Service is effectively a load balancer that use round robin. If we have many postgres instances every call to that endpoint can fetch metrics from different instance, which can be different depending how users are connected. Is my understanding correct?

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,483 @@
...
+		}
+		return nil
+	}
+
+	desired, err := buildPostgreSQLMetricsService(scheme, cluster)
+	if err != nil {
+		return fmt.Errorf("building PostgreSQL metrics Service: %w", err)
+	}
+
+	live := &corev1.Service{
```
mploski: why do we need this, cant we use desired directly?

pkg/postgresql/cluster/core/cluster.go
DmytroPI-dev: Updated with adding new conditions on that.

api/v4/postgresclusterclass_types.go
DmytroPI-dev: Didn't get you correctly 😞 , what previous comment do you mean? [This one](https://github.com/splunk/splunk-operator/pull/1808#discussion_r3064253129)?

pkg/postgresql/cluster/core/monitoring.go
DmytroPI-dev: We don't need it at all, as we have `poolerEnabled` in cluster.go

pkg/postgresql/cluster/core/monitoring.go
DmytroPI-dev: Using `desired` directly would mix desired state with server-populated/immutable `Service` fields like `clusterIP` `ipFamilies`, `resourceVersion` and related defaults, which can cause unnecessary diffs or update failures, so we should  use a minimal object instead. For `Service` specifically, this matters more than for `ConfigMap` or `ServiceMonitor`, because Service has immutable/defaulted networking fields we'd not want to stomp.

pkg/postgresql/cluster/core/monitoring.go
DmytroPI-dev: As per the [Prometheus configuration docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/) , the dedicated `Service` here is mainly a stable discovery contract for Prometheus Operator, not a client-style load balancer for metrics consumers.

Prometheus’ Kubernetes service discovery docs distinguish between service-level and endpoint-level discovery:

- For `service` discovery: _“The address will be set to the Kubernetes DNS name of the service and respective service port.”_
- For `endpoints` discovery: _“The endpoints role discovers targets from listed endpoints of a service. For each endpoint address *one target* is discovered per port.”_
- For `endpointslice` discovery: _“The endpointslice role discovers targets from existing endpointslices. For each endpoint address referenced in the endpointslice object one target is discovered.”_

So the intent here is not “scrape one `Service` and round-robin across instances”. The `ServiceMonitor` uses the `Service` as the discovery entry point, and Prometheus then discovers/scrapes the backing endpoints separately. That gives us per-instance targets, which is what we want for PostgreSQL metrics.

That is also why we'd need to create new `Service`s instead using the existing client-facing `Service`s by default: they are built for application traffic, may not expose the metrics port at all, and don’t provide an explicit observability contract.
Does this makes sense?

mploski: It does thank you! This reveal different discussion though. We cannot enforce usage of prometheus operator and it seems we are adding this as a dependency by using Service Monitor. The reason is that, the platform will be responsible for scrapping AND SOK aim to use otel collector for this. In that model we discover endpoints to be scrapped by providing appropiate annotiation to the PODs: https://www.dash0.com/guides/opentelemetry-prometheus-receiver#scrapeconfigs. So it seems we could simplify this code, by removing ServiceMonitor/Service but only populate requested annotiations so otel collector can scrape it. Let's discuss it offline :-)

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -397,6 +397,56 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	monitoringEnabled := postgresMetricsEnabled || (poolerMetricsEnabled && poolerEnabled)
+
+	monitoringFailure := func(reason conditionReasons, eventReason, message string, err error) (ctrl.Result, error) {
+		return ctrl.Result{}, handleMonitoringFailure(ctx, c, rc, postgresCluster, reason, eventReason, message, err)
+	}
+
+	oldConditions := make([]metav1.Condition, len(postgresCluster.Status.Conditions))
+	copy(oldConditions, postgresCluster.Status.Conditions)
+
+	if err := reconcilePostgreSQLMetricsService(ctx, c, rc.Scheme, postgresCluster, postgresMetricsEnabled); err != nil {
```
M4KIF: that could potentially be packed into unit of work style of code block, as It's very repeatable across the logic statements, It could get packed into and executable interface which handles the componentMetrics or something in this line of thought.
The code would then get separated into testable blocks and orchestrated cleanly.

pkg/postgresql/cluster/core/monitoring_unit_test.go
M4KIF, line 1: units, nice!

pkg/postgresql/cluster/core/monitoring.go
```diff
@@ -0,0 +1,467 @@
...
+func isPostgreSQLMetricsEnabled(cluster *enterprisev4.PostgresCluster, class *enterprisev4.PostgresClusterClass) bool {
+	if class == nil || class.Spec.Config == nil || class.Spec.Config.Monitoring == nil {
+		return false
+	}
+	classCfg := class.Spec.Config.Monitoring.PostgreSQLMetrics
+	if classCfg == nil || classCfg.Enabled == nil || !*classCfg.Enabled {
+		return false
+	}
+	if cluster == nil || cluster.Spec.Monitoring == nil || cluster.Spec.Monitoring.PostgreSQLMetrics == nil {
+		return true
```
M4KIF: what does that mean? areMetrics enabled return true on nil value's? Could you please explain?

api/v4/postgrescluster_types.go
```diff
@@ -107,6 +107,27 @@ type PostgresClusterSpec struct {
...
+type PostgresClusterMonitoring struct {
+
+	// +optional
+	PostgreSQLMetrics *FeatureDisableOverride `json:"postgresqlMetrics,omitempty"`
+
+	// +optional
+	ConnectionPoolerMetrics *FeatureDisableOverride `json:"connectionPoolerMetrics,omitempty"`
+}
+
+type FeatureDisableOverride struct {
```
mploski: could we allign to what we already have https://github.com/splunk/splunk-operator/pull/1808/changes#diff-e68b5e4731e03120f6625b75b8562e7845606ddaeae77144b79fc4368171e848L101?

docs/postgresSQLMonitoring-e2e.md
```diff
@@ -0,0 +1,321 @@
+# PostgreSQL Monitoring E2E on KIND
```
mploski: this file probably shouldn't be here, it sound like a internal testing approach, maybe move it to confluence?

docs/postgresSQLMonitoring-otel-e2e.md
```diff
@@ -0,0 +1,434 @@
+# PostgreSQL Monitoring E2E with OTel Collector
```
mploski: same here

docs/postgresSQLMonitoring-otel-e2e.md
```diff
@@ -0,0 +1,434 @@
...
+
+If the `kube-prometheus` release already exists, use:
+
+```bash
+helm upgrade kube-prometheus prometheus-community/kube-prometheus-stack \
+  --namespace monitoring \
+  -f test/postgresql/monitoring/prometheus-via-otel-values.yaml
+```
+
+This is important: Prometheus should scrape the OTel Collector exporter, not the PostgreSQL and PgBouncer pods directly. Otherwise Grafana will bypass OTel or you will get duplicate series.
```
mploski: What do you mean by this statement? We dont have Prometheus in this setup?

docs/postgresSQLMonitoring-otel-e2e.md
```diff
@@ -0,0 +1,434 @@
...
+Look for metrics such as:
+
+- `cnpg_pg_postmaster_start_time`
+- `cnpg_pg_database_size_bytes`
+- `cnpg_collector_pg_wal`
+- `cnpg_pg_stat_archiver_archived_count`
+- `cnpg_pgbouncer_last_collection_error`
+- `cnpg_pgbouncer_pools_cl_active`
+
+## 7. Verify Prometheus is scraping OTel
```
mploski: prometheus shouldnt be involved at all here otel collector should send it directly to grafana: https://grafana.com/docs/opentelemetry/#ingest-otlp-data

internal/controller/postgrescluster_controller.go
```diff
@@ -62,7 +62,13 @@ type PostgresClusterReconciler struct {
 // +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
 
 func (r *PostgresClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	rc := &clustercore.ReconcileContext{Client: r.Client, Scheme: r.Scheme, Recorder: r.Recorder, Metrics: r.Metrics}
+	metrics := r.Metrics
```
mploski: what problem do we solve here? Our reconciler should expect proper metrics being initiated, noop recorder is only for testing

internal/controller/postgrescluster_controller_test.go
```diff
@@ -79,13 +78,37 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
...
+	)
+
 	reconcileNTimes := func(times int) {
 		for i := 0; i < times; i++ {
 			_, err := reconciler.Reconcile(ctx, req)
 			Expect(err).NotTo(HaveOccurred())
 		}
 	}
 
+	recreateClusterClass := func(modify func(*enterprisev4.PostgresClusterClass)) {
```
mploski: It is not readable, can we make it straightforward? What problem we try to solve here? Why do we need to recreate class? especially in real live class cannot be recreated if it is already used. Maybe instead of this lets create few classes that we use in different scenarios and we change the cluster config itself

internal/controller/postgresdatabase_controller.go
```diff
@@ -76,7 +76,11 @@ func (r *PostgresDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 		return ctrl.Result{}, err
 	}
-	rc := &dbcore.ReconcileContext{Client: r.Client, Scheme: r.Scheme, Recorder: r.Recorder, Metrics: r.Metrics}
+	metrics := r.Metrics
```
mploski: As previously I dont think we should have it like that. metrics should be initiaited in reconciler

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -344,6 +346,9 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 		rc.emitPoolerReadyTransition(postgresCluster, oldConditions)
 	}
 
+	oldConditions := make([]metav1.Condition, len(postgresCluster.Status.Conditions))
```
mploski: maybe instead of duplicating making a oldConditions copy here and in line 337 we can have it before our switch and copy just once?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -526,6 +537,9 @@ func normalizeCNPGClusterSpec(spec cnpgv1.ClusterSpec, customDefinedParameters m
 	if len(spec.PostgresConfiguration.PgHBA) > 0 {
 		normalized.PgHBA = spec.PostgresConfiguration.PgHBA
 	}
+	if spec.InheritedMetadata != nil && len(spec.InheritedMetadata.Annotations) > 0 {
+		normalized.InheritedAnnotations = spec.InheritedMetadata.Annotations
```
mploski, line 538: I probably read it wrong, but why do we set normalized.InheritedAnnotations if we already have it injected into spec?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -119,6 +119,8 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 		return ctrl.Result{}, errors.Join(err, statusErr)
 	}
 
+	postgresMetricsEnabled := isPostgreSQLMetricsEnabled(postgresCluster, clusterClass)
```
M4KIF, line 122: Just an idea, could that be merged into a general port like "Are the component displaying metrics == enabled"? Because there is always a possibility to expand then without adding isComponent1MetricsEnabled, isComponentNMetricsEnabled. Just the method, getComponentMetricsSettings(...) return map[string(component)]bool. One config poll for any future coming.

pkg/postgresql/cluster/core/cluster.go
M4KIF, line 522: we could potentially map It via json contract. Unless we have tags busy in our specs, which we probably have. If not, It would be mapped straight into cnpg spec.
Btw. wha do ingeritedAnnotations mean tech wise?

pkg/postgresql/cluster/core/cluster_unit_test.go
```diff
@@ -453,14 +491,16 @@ func TestBuildCNPGCluster(t *testing.T) {
...
 	assert.Equal(t, "my-cluster", cluster.Name)
 	assert.Equal(t, "db-ns", cluster.Namespace)
 	require.Len(t, cluster.OwnerReferences, 1)
 	assert.Equal(t, "pg-uid", string(cluster.OwnerReferences[0].UID))
 	assert.Equal(t, 3, cluster.Spec.Instances)
+	require.NotNil(t, cluster.Spec.InheritedMetadata)
+	assert.Equal(t, postgresMetricsPortString, cluster.Spec.InheritedMetadata.Annotations[prometheusPortAnnotation])
 }
 
 func TestClusterSecretExists(t *testing.T) {
```
M4KIF, line 508: nit: Isn't It more readable to split the units into chunks?
a naming idea, just for reference. TestClusterSecret_with_n_expected_rwro_poolers_exist

pkg/postgresql/cluster/core/cluster.go
mploski: inherited annotations set a anottations on the k8s pod. Thanks do this we have a way to discover every pod with those annotations and this is what otel collector use to find pod endpoints to scrape from

pkg/postgresql/cluster/core/monitoring.go
mploski: I refactored this logic so it should be easier to read

pkg/postgresql/cluster/core/cluster.go
mploski: lets do this as a part of refactor initiative. Not sure if this should be separated port or not as this is tightly coupled to low level cluster provider though

pkg/postgresql/cluster/core/cluster_unit_test.go
mploski: We decided to follow table driven tests for tests that use the same methods but have different input/output expectations

pkg/postgresql/cluster/core/cluster.go
M4KIF: as the cluster provider is still orchestrated/managed by k8s I would say It's "portable". Yeah, let's analyse It along the way later.

pkg/postgresql/cluster/core/cluster.go
M4KIF: thank you

pkg/postgresql/cluster/core/cluster_unit_test.go
M4KIF: I thought that It might be the case. Backing off of the nit, all fine!

M4KIF (approved):

PR #1812, opened 2026-04-02
remove observedGeneration from postgresDatabase

DmytroPI-dev:
### Description

This PR aligns `PostgresDatabase` with the watch-driven reconciliation approach used for `PostgresCluster`. It removes the reconcile skip based on `ObservedGeneration`, reacts to owned `Secret`/`ConfigMap` drift even when the spec is unchanged, repairs configmap drift declaratively, and degrades status instead of recreating previously provisioned managed secrets that were deleted.

### Key Changes

- `postgresdatabase_controller.go` add explicit drift-trigger predicates for owned `Secret`, `ConfigMap`, and relevant CNPG `Database` changes. Conflict retries are requeued without surfacing noisy reconcile errors.
- `database.go` remove the `ObservedGeneration` short-circuit and revalidated stages on every reconcile. Add drift handling for managed user secrets, configmap repair/re-adoption, retained-resource adoption, conflict-friendly database reconcile handling, and quieter ready/not-found paths.
- `events.go` Add events for drift and retained-resource adoption.
- `cluster.go` reduced log noise on transient status-update conflicts.

### Testing and Verification

- Update controller tests in `postgresdatabase_controller_test.go` for:
  - configmap drift repair
  - deleted configmap recreation
  - deleted managed secret degradation without recreation
  - secret ownership re-adoption
  - adding a new database to an already provisioned resource
  - secondary-resource watch predicates
  - cluster-missing and role-conflict regressions
- Add unit tests in `database_unit_test.go` for secret/configmap drift and adoption paths.
- Verified behavior from controller logs/events during delete/recreate flows with both `Delete` and `Retain` policies.
- Automated tests were updated. All tests passed

### Related Issues

[CPI-1950](https://splunk.atlassian.net/browse/CPI-1950)

### PR Checklist

- [X] All tests pass locally.


internal/controller/postgresdatabase_controller.go
```diff
@@ -72,7 +73,12 @@ func (r *PostgresDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 	rc := &dbcore.ReconcileContext{Client: r.Client, Scheme: r.Scheme, Recorder: r.Recorder}
-	return dbcore.PostgresDatabaseService(ctx, rc, postgresDB, dbadapter.NewDBRepository)
+	result, err := dbcore.PostgresDatabaseService(ctx, rc, postgresDB, dbadapter.NewDBRepository)
+	if apierrors.IsConflict(err) {
```
mploski: dont we catch and requeue those errors already in service function?

internal/controller/postgresdatabase_controller.go
```diff
@@ -108,18 +120,47 @@ func (r *PostgresDatabaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Owns(&corev1.Secret{}, builder.WithPredicates(postgresDatabaseSecretPredicator())).
+		Owns(&corev1.ConfigMap{}, builder.WithPredicates(postgresDatabaseConfigMapPredicator())).
 		Named("postgresdatabase").
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: DatabaseTotalWorker,
 		}).
 		Complete(r)
 }
+
+func postgresDatabaseCNPGDatabasePredicator() predicate.Predicate {
```
mploski: could you refactor all of those predicates the same way I simpified them for cluster? We dont need to craft custom predicate Funcs everywhere. We should probably reuse what k8s give us by default

internal/controller/postgresdatabase_controller.go
DmytroPI-dev: missed this one, fixed

internal/controller/postgresdatabase_controller.go
```diff
@@ -95,31 +95,45 @@ func (r *PostgresDatabaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		},
+	)
+}
+
+func postgresDatabaseCNPGDatabasePredicator() predicate.Predicate {
+	return predicate.Or(
+		predicate.GenerationChangedPredicate{},
+		predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldObj, okOld := e.ObjectOld.(*cnpgv1.Database)
```
mploski, line 140: whats the difference between this predicate and GenerationChangedPredicate?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -332,10 +334,8 @@ func expectStatusCondition(current *enterprisev4.PostgresDatabase, conditionType
 	Expect(condition.Reason).To(Equal(expectedReason), "unexpected reason for %s", conditionType)
 }
 
-func expectReadyStatus(current *enterprisev4.PostgresDatabase, generation int64, expectedDatabase enterprisev4.DatabaseInfo) {
+func expectReadyStatus(current *enterprisev4.PostgresDatabase, expectedDatabase enterprisev4.DatabaseInfo) {
```
mploski: I'd still leave generation in status. Reason - we will know what is the CRD version status was emited for

pkg/postgresql/database/core/database.go
```diff
@@ -29,18 +29,37 @@ import (
...
 	ctx context.Context,
 	rc *ReconcileContext,
 	postgresDB *enterprisev4.PostgresDatabase,
 	newDBRepo NewDBRepoFunc,
-) (ctrl.Result, error) {
+) (result ctrl.Result, err error) {
 	c := rc.Client
 	logger := log.FromContext(ctx).WithValues("postgresDatabase", postgresDB.Name)
 	ctx = log.IntoContext(ctx, logger)
+	defer func() {
```
mploski: Why do we want to do defer here? We always catch the errors explicitly close to the called function so what benefit this brings?

pkg/postgresql/database/core/database.go
```diff
@@ -29,18 +29,37 @@ import (
...
+		if !errors.IsConflict(err) {
+			return
+		}
+		logger.Info("Conflict during PostgresDatabase reconciliation, will requeue")
+		result = ctrl.Result{Requeue: true}
+		err = nil
+	}()
 	logger.Info("Reconciling PostgresDatabase")
+	wasReady := postgresDB.Status.Phase != nil && *postgresDB.Status.Phase == string(readyDBPhase)
```
mploski, line 75: Could you explain what it does and how we use it?

pkg/postgresql/database/core/events.go
```diff
@@ -23,6 +23,7 @@ const (
 	EventClusterNotReady               = "ClusterNotReady"
 	EventRoleConflict                  = "RoleConflict"
 	EventUserSecretsFailed             = "UserSecretsFailed"
+	EventUserSecretsDriftDetected      = "UserSecretsDriftDetected"
```
mploski: TIL: we should start using roles instead of users

pkg/postgresql/database/core/database.go
```diff
@@ -787,31 +818,81 @@ func adoptResource(ctx context.Context, c client.Client, scheme *runtime.Scheme,
 	return c.Update(ctx, obj)
 }
 
-func reconcileUserSecrets(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase) error {
+func reconcileUserSecrets(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase, existingDatabases map[string]struct{}) error {
```
mploski: this function should not be bothered with logic checking what db is already provisioned, it should be clean

pkg/postgresql/database/core/database.go
```diff
@@ -787,31 +818,81 @@ func adoptResource(ctx context.Context, c client.Client, scheme *runtime.Scheme,
...
-		if err := ensureSecret(ctx, c, scheme, postgresDB, rwRoleName(dbSpec.Name), roleSecretName(postgresDB.Name, dbSpec.Name, secretRoleRW)); err != nil {
+		if err := ensureSecret(ctx, c, scheme, postgresDB, rwRoleName(dbSpec.Name), roleSecretName(postgresDB.Name, dbSpec.Name, secretRoleRW), databaseAlreadyProvisioned); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func ensureSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase, roleName, secretName string) error {
+func ensureSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase, roleName, secretName string, databaseAlreadyProvisioned bool) error {
```
mploski: this function should not be responsible for checking if db is already provisioned

pkg/postgresql/database/core/database.go
```diff
@@ -787,31 +818,81 @@ func adoptResource(ctx context.Context, c client.Client, scheme *runtime.Scheme,
...
+	default:
+		owner := metav1.GetControllerOf(secret)
+		return &secretReconcileError{
+			message: fmt.Sprintf("Managed Secret %s is controlled by %s %s", secretName, owner.Kind, owner.Name),
+			reason:  reasonSecretsDriftDetected,
+		}
+	}
+}
+
+func validateManagedSecret(secret *corev1.Secret, roleName string) error {
```
mploski: I wonder why we need this in the first place? What would be the use-case for example username not matching roleName?

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Returned this back.

internal/controller/postgresdatabase_controller.go
DmytroPI-dev: If my understanding is correct, GenerationChangedPredicate is triggered only when CNPG Database spec changes. If we have status-only update, generation might not always change. Owner reference changes on the owned CNPG Database would also stop triggering reconciliation.

pkg/postgresql/database/core/database.go
DmytroPI-dev: Remove defer, add explicit  requeue on conflict.

pkg/postgresql/database/core/database.go
DmytroPI-dev: The idea was to reduce EventPostgresDatabaseReady on every successful reconcile.

pkg/postgresql/database/core/events.go
DmytroPI-dev: Renamed.

pkg/postgresql/database/core/database.go
DmytroPI-dev: Removed logic checking.

pkg/postgresql/database/core/database.go
DmytroPI-dev: Removed logic checking.

pkg/postgresql/database/core/database.go
DmytroPI-dev: To prevent adopting secret in case of manual changes, for example. Does this makes sense?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -389,14 +401,41 @@ func expectStatusCondition(current *enterprisev4.PostgresDatabase, conditionType
...
 	expectStatusPhase(current, phaseReady)
-	Expect(current.Status.ObservedGeneration).NotTo(BeNil())
-	Expect(*current.Status.ObservedGeneration).To(Equal(generation))
 	Expect(current.Status.Databases).To(HaveLen(1))
 	Expect(current.Status.Databases[0].Name).To(Equal(expectedDatabase.Name))
 	Expect(current.Status.Databases[0].Ready).To(Equal(expectedDatabase.Ready))
 	Expect(current.Status.Databases[0].AdminUserSecretRef).NotTo(BeNil())
 	Expect(current.Status.Databases[0].RWUserSecretRef).NotTo(BeNil())
 	Expect(current.Status.Databases[0].ConfigMapRef).NotTo(BeNil())
+	Expect(current.Status.ObservedGeneration).NotTo(BeNil())
```
M4KIF, line 413: Isn't that supposed to be removed from the general logic?

pkg/postgresql/database/core/database.go
```diff
@@ -99,13 +126,19 @@ func PostgresDatabaseService(
...
+				return result, conflictErr
+			}
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: retryDelay}, nil
 
 	case ClusterReady:
 		rc.emitOnConditionTransition(postgresDB, postgresDB.Status.Conditions, clusterReady, EventClusterValidated, "Referenced PostgresCluster is ready")
 		if err := updateStatus(clusterReady, metav1.ConditionTrue, reasonClusterAvailable, "Cluster is operational", provisioningDBPhase); err != nil {
+			if result, conflictErr, ok := requeueOnConflict(err, "persisting cluster ready status"); ok {
```
M4KIF: Do I understand correctly, adding this logic is a replacement for observed generation?

pkg/postgresql/database/core/database.go
```diff
@@ -134,40 +170,73 @@ func PostgresDatabaseService(
...
 	}
 
 	// Phase: CredentialProvisioning — secrets must exist before roles are patched.
 	// CNPG rejects a PasswordSecretRef pointing at a missing secret.
-	if err := reconcileUserSecrets(ctx, c, rc.Scheme, postgresDB); err != nil {
+	if err := reconcileUserSecrets(ctx, c, rc.Scheme, postgresDB, previouslyProvisionedDatabases); err != nil {
+		if result, conflictErr, ok := requeueOnConflict(err, "reconciling user secrets"); ok {
+			return result, conflictErr
+		}
+		var driftErr *secretReconcileError
```
M4KIF: are we sure, that It will always be drift related?
What I'm after is having It coupled to the name secretReconcileError, which seems quite generic. Even If the underlaying struct could be verbose and drift only in values.

pkg/postgresql/database/core/database.go
```diff
@@ -41,14 +57,26 @@ func PostgresDatabaseService(
 	logger := log.FromContext(ctx).WithValues("postgresDatabase", postgresDB.Name)
 	ctx = log.IntoContext(ctx, logger)
 	logger.Info("Reconciling PostgresDatabase")
+	wasReady := postgresDB.Status.Phase != nil && *postgresDB.Status.Phase == string(readyDBPhase)
+	previouslyProvisionedDatabases := existingDatabaseStatus(postgresDB)
 
 	updateStatus := func(conditionType conditionTypes, conditionStatus metav1.ConditionStatus, reason conditionReasons, message string, phase reconcileDBPhases) error {
 		return persistStatus(ctx, c, postgresDB, conditionType, conditionStatus, reason, message, phase)
 	}
+	requeueOnConflict := func(err error, action string) (ctrl.Result, error, bool) {
```
M4KIF: more of a nit: Will having a lamda created here not cause to miss scope for us. Ie. If we have k8s conflict error occuring in a few places. Will the action log be sufficient and fully satisfying our business needs? Also, there could be some metrics done in the future from reconcilation misses(conflicts) which could prove useful for system admins.

internal/controller/postgresdatabase_controller_test.go
DmytroPI-dev: Yes, we removed observedGeneration to act as a database readiness check, but left  it to have an idea know what is the CRD version status was emitted for. See [here](https://github.com/splunk/splunk-operator/pull/1812#discussion_r3043402649)

pkg/postgresql/database/core/database.go
DmytroPI-dev: If asking about this: 
```
if result, conflictErr, ok := requeueOnConflict(err, "persisting cluster ready status"); ok {
	return result, conflictErr
}
```
- this is to catch errors on version conflicts and requeue immediately to soothe reconciliation storm.

pkg/postgresql/database/core/database.go
DmytroPI-dev: right now, we have `reason: reasonSecretsDriftDetected` (see [here](https://github.com/splunk/splunk-operator/pull/1812/files#diff-ec6b127d1a5889cad3cf908a8af626f6bf1bc6bbf65cde2d0d36ad402014a0e5R1007) and [here](https://github.com/splunk/splunk-operator/pull/1812/files#diff-ec6b127d1a5889cad3cf908a8af626f6bf1bc6bbf65cde2d0d36ad402014a0e5R1021) ), but this makes sense in general, will rename it as `secretErr`

pkg/postgresql/database/core/database.go
DmytroPI-dev: Add typed conflict categories instead of lambda

internal/controller/postgresdatabase_controller.go
mploski: but then if you expect object to be triggered with every change including status, cant we simply remove predicate and react to every trigger? What events we filter out right now?

pkg/postgresql/database/core/database.go
```diff
@@ -29,6 +29,36 @@ import (
...
+	return e.message
+}
+
+func requeueOnConflict(ctx context.Context, err error, category reconcileConflictCategory, action string) (ctrl.Result, error, bool) {
+	if !errors.IsConflict(err) {
+		return ctrl.Result{}, err, false
+	}
+
+	// Keep the category stable so future metrics or events can aggregate conflict sources.
+	log.FromContext(ctx).Info(
```
mploski, line 55: Plase follow https://splunk.atlassian.net/wiki/spaces/CCP/pages/1079831167399/PostgreSQL+Controllers+Logging+Strategy for the logging approach

pkg/postgresql/database/core/database.go
mploski: make sense!

pkg/postgresql/database/core/database.go
```diff
@@ -840,31 +964,116 @@ func adoptResource(ctx context.Context, c client.Client, scheme *runtime.Scheme,
...
 
-func reconcileUserSecrets(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase) error {
+func secretMissingPolicyForDB(dbName string, existingDBs map[string]struct{}) secretMissingPolicy {
+	if _, exists := existingDBs[dbName]; exists {
+		return reportSecretDriftIfMissing
+	}
+	return createSecretIfMissing
+}
+
+func reconcileUserSecrets(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase, existingDatabases map[string]struct{}) error {
```
mploski: maybe replace user with role everywhere?

pkg/postgresql/database/core/database.go
```diff
@@ -840,31 +964,116 @@ func adoptResource(ctx context.Context, c client.Client, scheme *runtime.Scheme,
...
-		if err := ensureSecret(ctx, c, scheme, postgresDB, rwRoleName(dbSpec.Name), roleSecretName(postgresDB.Name, dbSpec.Name, secretRoleRW)); err != nil {
+		if err := reconcileUserRoleSecret(ctx, c, scheme, postgresDB, dbSpec.Name, secretRoleRW, missingPolicy); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func reconcileUserRoleSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, postgresDB *enterprisev4.PostgresDatabase, dbName, role string, missingPolicy secretMissingPolicy) error {
+	roleName := adminRoleName(dbName)
```
mploski: I'd remove the understanding whether the role is admin or rw and just pass correct name in the parameters to the function

pkg/postgresql/database/core/database.go
mploski: I wonder if this is not too much. If we need to analyse if every object we create is accurate then we should also do the same for configmaps, secrets etc. This is basically our contract that shouldnt be changed and if user does, this is their problem. So the question would be, how defensive we want to be and if we want to stick to it should we do a checking for other things as well?

M4KIF:
apart from the logging and some other nits (which ofc should be addressed) there is a nice structure seen -> reconcile flow rework -> tests expanded accordingly to catch the change and validate It. I like that

internal/controller/postgresdatabase_controller.go
DmytroPI-dev: As we update status at least in three places, (414, 97, 627), and this custom predicate triggers reconcile in cases when deletion starts, finalizers changes, and ignore other update-only metadata/status changes. We filter out condition-only updates, like ClusterReady, SecretsReady, ConfigMapsReady, RolesReady, etc. This was one of the sources of reconciliation storm we were struggling.

PR #1814, opened 2026-04-02
Cluster sync adj in p&a flavour

M4KIF:
### Description

Rewritten the sync logic
- moved each component checks to separate objects all implementing the Condition interface by which we iterate and check the health.

### Key Changes

- interfaced the health check
- implemented the above for provisioner, pooler, configmap, secret
- short unit test added
- added constants and state check via bitmask

### Testing and Verification

local integ suite passes (make test) and units (pkg/postgres/cluster/core go test) passes

### Related Issues

### PR Checklist

- [nie wiem] Code changes adhere to the project's coding standards.
- [x] All tests pass locally.

### DIFF
--- a/internal/controller/postgrescluster_controller_test.go
+++ b/internal/controller/postgrescluster_controller_test.go
@@ -19,7 +19,9 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -52,6 +54,20 @@ import (
 * PC-09 ignores no-op updates
  */
 
+func containsEvents(events *[]string, recorder *record.FakeRecorder, eventType string, event string) bool {
+	for {
+		select {
+		case e := <-recorder.Events:
+			*events = append(*events, e)
+			if strings.Contains(e, eventType) && strings.Contains(e, event) {
+				return true
+			}
+		default:
+			return false
+		}
+	}
+}
+
 var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 
 	const (
@@ -79,6 +95,7 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 		pgClusterClassKey types.NamespacedName
 		reconciler        *PostgresClusterReconciler
 		req               reconcile.Request
+		fakeRecorder      *record.FakeRecorder
 	)
 
 	reconcileNTimes := func(times int) {
@@ -162,13 +179,17 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 			Spec: enterprisev4.PostgresClusterSpec{
 				Class:                 className,
 				ClusterDeletionPolicy: ptr.To(deletePolicy),
+				ManagedRoles: []enterprisev4.ManagedRole{
+					{Name: "app_user", Exists: true},
+					{Name: "app_user_rw", Exists: true},
+				},
 			},
 		}
-
+		fakeRecorder = record.NewFakeRecorder(100)
 		reconciler = &PostgresClusterReconciler{
 			Client:         k8sClient,
 			Scheme:         k8sClient.Scheme(),
-			Recorder:       record.NewFakeRecorder(100),
+			Recorder:       fakeRecorder,
 			Metrics:        &pgprometheus.NoopRecorder{},
 			FleetCollector: pgprometheus.NewFleetCollector(),
 		}
@@ -255,12 +276,17 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				cond := meta.FindStatusCondition(pc.Status.Conditions, "ClusterReady")
 				Expect(cond).NotTo(BeNil())
 				Expect(cond.Status).To(Equal(metav1.ConditionFalse))
-				Expect(cond.Reason).To(Equal("ClusterBuildSucceeded"))
+				Expect(cond.Reason).To(Equal("CNPGClusterProvisioning"))
 
 				// Simulate external CNPG controller status progression.
 				cnpg := &cnpgv1.Cluster{}
 				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
 				cnpg.Status.Phase = cnpgv1.PhaseHealthy
+				cnpg.Status.ManagedRolesStatus = cnpgv1.ManagedRoles{
+					ByStatus: map[cnpgv1.RoleStatus][]string{
+						cnpgv1.RoleStatusReconciled: {"app_user", "app_user_rw"},
+					},
+				}
 				Expect(k8sClient.Status().Update(ctx, cnpg)).To(Succeed())
 				reconcileNTimes(1)
 
@@ -270,9 +296,47 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				Expect(cond).NotTo(BeNil())
 				Expect(cond.Status).To(Equal(metav1.ConditionTrue))
 				Expect(cond.Reason).To(Equal("CNPGClusterHealthy"))
+
+				secretCond := meta.FindStatusCondition(pc.Status.Conditions, "SecretsReady")
+				Expect(secretCond).NotTo(BeNil())
+				Expect(secretCond.Status).To(Equal(metav1.ConditionTrue))
+				Expect(secretCond.Reason).To(Equal("SuperUserSecretReady"))
+
+				configMapCond := meta.FindStatusCondition(pc.Status.Conditions, "ConfigMapsReady")
+				Expect(configMapCond).NotTo(BeNil())
+				Expect(configMapCond.Status).To(Equal(metav1.ConditionTrue))
+				Expect(configMapCond.Reason).To(Equal("ConfigMapReconciled"))
+
+				managedRolesCond := meta.FindStatusCondition(pc.Status.Conditions, "ManagedRolesReady")
+				Expect(managedRolesCond).NotTo(BeNil())
+				Expect(managedRolesCond.Status).To(Equal(metav1.ConditionTrue))
+				Expect(managedRolesCond.Reason).To(Equal("ManagedRolesReconciled"))
+
+				// Pooler is disabled in this suite fixture, but converge publishes PoolerReady=True with disabled message.
+				poolerCond := meta.FindStatusCondition(pc.Status.Conditions, "PoolerReady")
+				Expect(poolerCond).NotTo(BeNil())
+				Expect(poolerCond.Status).To(Equal(metav1.ConditionTrue))
+				Expect(poolerCond.Reason).To(Equal("AllInstancesReady"))
+				Expect(poolerCond.Message).To(Equal("Connection pooler disabled"))
+
+				Expect(pc.Status.ManagedRolesStatus).NotTo(BeNil())
+				Expect(pc.Status.ManagedRolesStatus.Reconciled).To(ContainElements("app_user", "app_user_rw"))
+
+				Expect(pc.Status.Phase).NotTo(BeNil())
+				Expect(*pc.Status.Phase).To(Equal("Ready"))
+				Expect(pc.Status.ProvisionerRef).NotTo(BeNil())
+				Expect(pc.Status.ProvisionerRef.Kind).To(Equal("Cluster"))
+				Expect(pc.Status.ProvisionerRef.Name).To(Equal(clusterName))
+
 				Expect(pc.Status.Resources).NotTo(BeNil())
 				Expect(pc.Status.Resources.SuperUserSecretRef).NotTo(BeNil())
 				Expect(pc.Status.Resources.ConfigMapRef).NotTo(BeNil())
+
+				received := make([]string, 0, 8)
+				Expect(containsEvents(
+					&received, fakeRecorder,
+					v1.EventTypeNormal, core.EventClusterReady,
+				)).To(BeTrue(), "events seen: %v", received)
 			})
 
 			// PC-07
@@ -308,6 +372,7 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 		Context("with PostgreSQL metrics enabled in class", func() {
 			BeforeEach(func() {
 				pgCluster.Spec.Class = classNameMetrics
+				pgCluster.Spec.ManagedRoles = nil
 			})
 
 			It("adds scrape annotations to the CNPG Cluster", func() {
@@ -316,6 +381,17 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 
 				cnpg := &cnpgv1.Cluster{}
 				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
+
+				cnpg.Status.Phase = cnpgv1.PhaseHealthy
+				cnpg.Status.ManagedRolesStatus = cnpgv1.ManagedRoles{
+					ByStatus: map[cnpgv1.RoleStatus][]string{
+						cnpgv1.RoleStatusReconciled: {"app_user", "app_user_rw"},
+					},
+				}
+				Expect(k8sClient.Status().Update(ctx, cnpg)).To(Succeed())
+
+				reconcileNTimes(1)
+
 				Expect(cnpg.Spec.InheritedMetadata).NotTo(BeNil())
 				Expect(cnpg.Spec.InheritedMetadata.Annotations).To(HaveKeyWithValue(scrapeAnnotationKey, "true"))
 				Expect(cnpg.Spec.InheritedMetadata.Annotations).To(HaveKeyWithValue(pathAnnotationKey, metricsPath))
@@ -353,6 +429,7 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 		Context("with connection pooler metrics enabled in class", func() {
 			BeforeEach(func() {
 				pgCluster.Spec.Class = classNamePooler
+				pgCluster.Spec.ManagedRoles = nil
 			})
 
 			It("adds scrape annotations to poolers only after the CNPG cluster becomes healthy", func() {
@@ -439,7 +516,7 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 	When("reconciling with invalid or drifted dependencies", func() {
 		// PC-05
 		Context("when referenced class does not exist", func() {
-			It("fails with class-not-found condition", func() {
+			It("fails with class-not-found condition and emits a warning event", func() {
 				badName := "bad-" + clusterName
 				badKey := types.NamespacedName{Name: badName, Namespace: namespace}
 
@@ -464,6 +541,12 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 					cond := meta.FindStatusCondition(current.Status.Conditions, "ClusterReady")
 					return cond != nil && cond.Reason == "ClusterClassNotFound"
 				}, "20s", "250ms").Should(BeTrue())
+
+				received := make([]string, 0, 8)
+				Expect(containsEvents(
+					&received, fakeRecorder,
+					v1.EventTypeWarning, core.EventClusterClassNotFound,
+				)).To(BeTrue(), "events seen: %v", received)
 			})
 		})
 
--- a/internal/controller/postgrescluster_controller_test.go
+++ b/internal/controller/postgrescluster_controller_test.go
@@ -54,18 +54,24 @@ import (
 * PC-09 ignores no-op updates
  */
 
-func containsEvents(events *[]string, recorder *record.FakeRecorder, eventType string, event string) bool {
+func CollectEvents(events *[]string, recorder *record.FakeRecorder) {
 	for {
 		select {
 		case e := <-recorder.Events:
 			*events = append(*events, e)
-			if strings.Contains(e, eventType) && strings.Contains(e, event) {
-				return true
-			}
 		default:
-			return false
+			return
 		}
 	}
+}
+
+func ContainsEvent(events []string, eventType string, event string) bool {
+	for _, e := range events {
+		if strings.Contains(e, eventType) && strings.Contains(e, event) {
+			return true
+		}
+	}
+	return false
 }
 
 var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
@@ -278,10 +284,31 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				Expect(cond.Status).To(Equal(metav1.ConditionFalse))
 				Expect(cond.Reason).To(Equal("CNPGClusterProvisioning"))
 
-				// Simulate external CNPG controller status progression.
+				secretCond := meta.FindStatusCondition(pc.Status.Conditions, "SecretsReady")
+				Expect(secretCond).NotTo(BeNil())
+				Expect(secretCond.Status).To(Equal(metav1.ConditionTrue))
+				Expect(secretCond.Reason).To(Equal("SuperUserSecretReady"))
+
+				configMapCond := meta.FindStatusCondition(pc.Status.Conditions, "ConfigMapsReady")
+				// ConfigMap converge runs in the runtime phase; at this point reconcile may
+				// still be returning from provisioner pending and not have written it yet.
+				Expect(configMapCond).To(BeNil())
+
+				// Simulate CNPG becoming healthy first, but without managed roles status published yet.
 				cnpg := &cnpgv1.Cluster{}
 				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
 				cnpg.Status.Phase = cnpgv1.PhaseHealthy
+				Expect(k8sClient.Status().Update(ctx, cnpg)).To(Succeed())
+				reconcileNTimes(1)
+
+				Expect(k8sClient.Get(ctx, pgClusterKey, pc)).To(Succeed())
+				managedRolesCond := meta.FindStatusCondition(pc.Status.Conditions, "ManagedRolesReady")
+				Expect(managedRolesCond).NotTo(BeNil())
+				Expect(managedRolesCond.Status).To(Equal(metav1.ConditionFalse))
+				Expect(managedRolesCond.Reason).To(Equal("ManagedRolesPending"))
+
+				// Simulate external CNPG controller publishing managed roles status.
+				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
 				cnpg.Status.ManagedRolesStatus = cnpgv1.ManagedRoles{
 					ByStatus: map[cnpgv1.RoleStatus][]string{
 						cnpgv1.RoleStatusReconciled: {"app_user", "app_user_rw"},
@@ -297,17 +324,17 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				Expect(cond.Status).To(Equal(metav1.ConditionTrue))
 				Expect(cond.Reason).To(Equal("CNPGClusterHealthy"))
 
-				secretCond := meta.FindStatusCondition(pc.Status.Conditions, "SecretsReady")
+				secretCond = meta.FindStatusCondition(pc.Status.Conditions, "SecretsReady")
 				Expect(secretCond).NotTo(BeNil())
 				Expect(secretCond.Status).To(Equal(metav1.ConditionTrue))
 				Expect(secretCond.Reason).To(Equal("SuperUserSecretReady"))
 
-				configMapCond := meta.FindStatusCondition(pc.Status.Conditions, "ConfigMapsReady")
+				configMapCond = meta.FindStatusCondition(pc.Status.Conditions, "ConfigMapsReady")
 				Expect(configMapCond).NotTo(BeNil())
 				Expect(configMapCond.Status).To(Equal(metav1.ConditionTrue))
 				Expect(configMapCond.Reason).To(Equal("ConfigMapReconciled"))
 
-				managedRolesCond := meta.FindStatusCondition(pc.Status.Conditions, "ManagedRolesReady")
+				managedRolesCond = meta.FindStatusCondition(pc.Status.Conditions, "ManagedRolesReady")
 				Expect(managedRolesCond).NotTo(BeNil())
 				Expect(managedRolesCond.Status).To(Equal(metav1.ConditionTrue))
 				Expect(managedRolesCond.Reason).To(Equal("ManagedRolesReconciled"))
@@ -333,8 +360,13 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				Expect(pc.Status.Resources.ConfigMapRef).NotTo(BeNil())
 
 				received := make([]string, 0, 8)
-				Expect(containsEvents(
-					&received, fakeRecorder,
+				CollectEvents(&received, fakeRecorder)
+				Expect(ContainsEvent(
+					received,
+					v1.EventTypeNormal, core.EventConfigMapReconciled,
+				)).To(BeTrue(), "events seen: %v", received)
+				Expect(ContainsEvent(
+					received,
 					v1.EventTypeNormal, core.EventClusterReady,
 				)).To(BeTrue(), "events seen: %v", received)
 			})
@@ -442,6 +474,13 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				Expect(apierrors.IsNotFound(k8sClient.Get(ctx, rwKey, &cnpgv1.Pooler{}))).To(BeTrue())
 				Expect(apierrors.IsNotFound(k8sClient.Get(ctx, roKey, &cnpgv1.Pooler{}))).To(BeTrue())
 
+				pc := &enterprisev4.PostgresCluster{}
+				Expect(k8sClient.Get(ctx, pgClusterKey, pc)).To(Succeed())
+				poolerCond := meta.FindStatusCondition(pc.Status.Conditions, "PoolerReady")
+				// Pooler component is gated behind provisioner readiness, so before CNPG
+				// becomes healthy the condition may not be written yet.
+				Expect(poolerCond).To(BeNil())
+
 				cnpg := &cnpgv1.Cluster{}
 				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
 				cnpg.Status.Phase = cnpgv1.PhaseHealthy
@@ -464,6 +503,28 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 					g.Expect(ro.Spec.Template.ObjectMeta.Annotations).To(HaveKeyWithValue(scrapeAnnotationKey, "true"))
 					g.Expect(ro.Spec.Template.ObjectMeta.Annotations).To(HaveKeyWithValue(pathAnnotationKey, metricsPath))
 					g.Expect(ro.Spec.Template.ObjectMeta.Annotations).To(HaveKeyWithValue(portAnnotationKey, poolerPort))
+
+					// Simulate CNPG pooler controller publishing status progression.
+					if rw.Status.Instances < 2 {
+						rw.Status.Instances = 2
+						g.Expect(k8sClient.Status().Update(ctx, rw)).To(Succeed())
+					}
+					if ro.Status.Instances < 2 {
+						ro.Status.Instances = 2
+						g.Expect(k8sClient.Status().Update(ctx, ro)).To(Succeed())
+					}
+				}, "20s", "250ms").Should(Succeed())
+
+				Eventually(func(g Gomega) {
+					_, err := reconciler.Reconcile(ctx, req)
+					g.Expect(err).NotTo(HaveOccurred())
+
+					updated := &enterprisev4.PostgresCluster{}
+					g.Expect(k8sClient.Get(ctx, pgClusterKey, updated)).To(Succeed())
+					poolerReadyCond := meta.FindStatusCondition(updated.Status.Conditions, "PoolerReady")
+					g.Expect(poolerReadyCond).NotTo(BeNil())
+					g.Expect(poolerReadyCond.Status).To(Equal(metav1.ConditionTrue))
+					g.Expect(poolerReadyCond.Reason).To(Equal("AllInstancesReady"))
 				}, "20s", "250ms").Should(Succeed())
 			})
 		})
@@ -543,8 +604,9 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				}, "20s", "250ms").Should(BeTrue())
 
 				received := make([]string, 0, 8)
-				Expect(containsEvents(
-					&received, fakeRecorder,
+				CollectEvents(&received, fakeRecorder)
+				Expect(ContainsEvent(
+					received,
 					v1.EventTypeWarning, core.EventClusterClassNotFound,
 				)).To(BeTrue(), "events seen: %v", received)
 			})
@@ -564,6 +626,70 @@ var _ = Describe("PostgresCluster Controller", Label("postgres"), func() {
 				reconcileNTimes(2)
 				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
 				Expect(cnpg.Spec.Instances).To(Equal(int(clusterMemberCount)))
+			})
+		})
+
+		Context("when a configmap spec changes", func() {
+			BeforeEach(func() {
+				// Keep this test focused on ConfigMap behavior; otherwise reconcile can
+				// stop on ManagedRolesPending before ConfigMap status is written.
+				pgCluster.Spec.ManagedRoles = nil
+			})
+
+			It("emits ConfigMapReconciled event on configmap update", func() {
+				Expect(k8sClient.Create(ctx, pgCluster)).To(Succeed())
+				reconcileNTimes(2)
+
+				// Make sure runtime can proceed (if needed in your fixture)
+				cnpg := &cnpgv1.Cluster{}
+				Expect(k8sClient.Get(ctx, pgClusterKey, cnpg)).To(Succeed())
+				cnpg.Status.Phase = cnpgv1.PhaseHealthy
+				Expect(k8sClient.Status().Update(ctx, cnpg)).To(Succeed())
+				reconcileNTimes(1)
+
+				// Drain baseline events so we don't match the initial "created" event.
+				received := make([]string, 0, 16)
+				CollectEvents(&received, fakeRecorder)
+				received = received[:0]
+
+				// Drift the managed ConfigMap.
+				pc := &enterprisev4.PostgresCluster{}
+				Eventually(func() bool {
+					if err := k8sClient.Get(ctx, pgClusterKey, pc); err != nil {
+						return false
+					}
+					return pc.Status.Resources != nil && pc.Status.Resources.ConfigMapRef != nil
+				}, "5s", "100ms").Should(BeTrue())
+
+				cmKey := types.NamespacedName{
+					Name:      pc.Status.Resources.ConfigMapRef.Name,
+					Namespace: namespace,
+				}
+				cm := &v1.ConfigMap{}
+				Expect(k8sClient.Get(ctx, cmKey, cm)).To(Succeed())
+				delete(cm.Data, "CLUSTER_RW_ENDPOINT") // force reconciliation update
+				Expect(k8sClient.Update(ctx, cm)).To(Succeed())
+
+				// Reconcile and assert updated event.
+				reconcileNTimes(1)
+
+				Eventually(func() bool {
+					CollectEvents(&received, fakeRecorder)
+
+					// reason match
+					if !ContainsEvent(received, v1.EventTypeNormal, core.EventConfigMapReconciled) {
+						return false
+					}
+					// message-level match for update (not create)
+					for _, e := range received {
+						if strings.Contains(e, v1.EventTypeNormal) &&
+							strings.Contains(e, core.EventConfigMapReconciled) &&
+							strings.Contains(e, "updated") {
+							return true
+						}
+					}
+					return false
+				}, "5s", "100ms").Should(BeTrue(), "events seen: %v", received)
 			})
 		})
 	})
--- a/pkg/postgresql/cluster/core/cluster.go
+++ b/pkg/postgresql/cluster/core/cluster.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	cnpgv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+	"github.com/go-logr/logr"
 	password "github.com/sethvargo/go-password/password"
 	enterprisev4 "github.com/splunk/splunk-operator/api/v4"
 	pgcConstants "github.com/splunk/splunk-operator/pkg/postgresql/cluster/core/types/constants"
@@ -165,49 +166,92 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 	secretComponent := newSecretModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, postgresSecretName)
 	clusterComponent := newClusterModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, clusterClass, mergedConfig, postgresSecretName)
 
-	bootstrapComponents := []component{
-		secretComponent,
-		clusterComponent,
+	bootstrapManager := &componentManager{
+		components: []component{
+			secretComponent,
+			clusterComponent,
+		},
+		logger: logger,
+	}
+	result, err := bootstrapManager.Handle(ctx)
+	if err != nil {
+		return result, err
+	}
+	if result != (ctrl.Result{}) {
+		return result, nil
 	}
 
-	phase := func(component component) (ctrl.Result, error) {
-		componentLogger := logger.WithValues("component", component.Name())
-		gate, gateErr := component.EvaluatePrerequisites(ctx)
-		if gateErr != nil {
-			if isTransientError(gateErr) {
-				componentLogger.Error(gateErr, "Component prerequisite transient error, requeueing", "step", "prerequisites")
-				return transientResult(gateErr), nil
-			}
-			componentLogger.Error(gateErr, "Component prerequisite evaluation failed", "step", "prerequisites")
-			return ctrl.Result{}, fmt.Errorf("%s prerequisites: %w", component.Name(), gateErr)
+	cnpgCluster = clusterComponent.cnpgCluster
+	runtimeView := clusterRuntimeViewAdapter{model: clusterComponent}
+
+	runtimeManager := &componentManager{
+		components: []component{
+			newManagedRolesModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
+			newPoolerModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, clusterClass, mergedConfig, cnpgCluster, poolerEnabled, poolerConfigPresent),
+			newConfigMapModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
+		},
+		logger: logger,
+	}
+
+	result, err = runtimeManager.Handle(ctx)
+	if err != nil {
+		return result, err
+	}
+	if result != (ctrl.Result{}) {
+		return result, nil
+	}
+
+	logger.Info("Reconciliation complete")
+	if err := updatePhaseStatus(readyClusterPhase); err != nil {
+		if apierrors.IsConflict(err) {
+			return ctrl.Result{Requeue: true}, nil
 		}
-		if !gate.Allowed {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+func isTransientError(err error) bool {
+	return apierrors.IsConflict(err) ||
+		apierrors.IsServerTimeout(err) ||
+		apierrors.IsTooManyRequests(err) ||
+		apierrors.IsTimeout(err)
+}
+
+func transientResult(err error) ctrl.Result {
+	if apierrors.IsConflict(err) {
+		return ctrl.Result{Requeue: true}
+	}
+	return ctrl.Result{RequeueAfter: retryDelay}
+}
+
+func writeComponentStatus(updateStatus healthStatusUpdater, health componentHealth) error {
+	if updateStatus == nil {
+		return nil
+	}
+	return updateStatus(health)
+}
+
+type componentManager struct {
+	components []component
+	logger     logr.Logger
+}
+
+func (m *componentManager) Handle(ctx context.Context) (ctrl.Result, error) {
+	for _, component := range m.components {
+		componentLogger := m.logger.WithValues("component", component.Name())
+		gate := component.EvaluatePrerequisites(ctx)
+
+		if gate.Allowed {
+			component.Actuate(ctx)
+		} else {
 			componentLogger.Info("Component blocked by prerequisites",
 				"step", "prerequisites",
 				"condition", gate.Health.Condition,
 				"reason", gate.Health.Reason,
 				"phase", gate.Health.Phase,
 				"requeueAfter", gate.Health.Result.RequeueAfter)
-			health, err := component.Converge(ctx)
-			if err != nil && isTransientError(err) {
-				return transientResult(err), nil
-			}
-			if err != nil {
-				componentLogger.Error(err, "Blocked component convergence failed", "step", "converge")
-				return health.Result, fmt.Errorf("%s converge (blocked): %w", component.Name(), err)
-			}
-			return health.Result, nil
 		}
-
-		if err := component.Actuate(ctx); err != nil {
-			if isTransientError(err) {
-				componentLogger.Error(err, "Component actuation transient error, requeueing", "step", "actuate")
-				return transientResult(err), nil
-			}
-			componentLogger.Error(err, "Component actuation failed", "step", "actuate")
-			return ctrl.Result{}, fmt.Errorf("%s actuate: %w", component.Name(), err)
-		}
-		componentLogger.Info("Component actuation completed", "step", "actuate")
 
 		health, err := component.Converge(ctx)
 		if err != nil && isTransientError(err) {
@@ -243,65 +287,8 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 				"requeueAfter", health.Result.RequeueAfter)
 			return health.Result, nil
 		}
-		return ctrl.Result{}, nil
-	}
-
-	for _, component := range bootstrapComponents {
-		result, err := phase(component)
-		if err != nil {
-			return result, err
-		}
-		if result != (ctrl.Result{}) {
-			return result, nil
-		}
-	}
-
-	cnpgCluster = clusterComponent.cnpgCluster
-	runtimeView := clusterRuntimeViewAdapter{model: clusterComponent}
-	runtimeComponents := []component{
-		newManagedRolesModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
-		newPoolerModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, clusterClass, mergedConfig, cnpgCluster, poolerEnabled, poolerConfigPresent),
-		newConfigMapModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
-	}
-
-	for _, component := range runtimeComponents {
-		result, err := phase(component)
-		if err != nil {
-			return result, err
-		}
-		if result != (ctrl.Result{}) {
-			return result, nil
-		}
-	}
-	logger.Info("Reconciliation complete")
-	if err := updatePhaseStatus(readyClusterPhase); err != nil {
-		if apierrors.IsConflict(err) {
-			return ctrl.Result{Requeue: true}, nil
-		}
-		return ctrl.Result{}, err
 	}
 	return ctrl.Result{}, nil
-}
-
-func isTransientError(err error) bool {
-	return apierrors.IsConflict(err) ||
-		apierrors.IsServerTimeout(err) ||
-		apierrors.IsTooManyRequests(err) ||
-		apierrors.IsTimeout(err)
-}
-
-func transientResult(err error) ctrl.Result {
-	if apierrors.IsConflict(err) {
-		return ctrl.Result{Requeue: true}
-	}
-	return ctrl.Result{RequeueAfter: retryDelay}
-}
-
-func writeComponentStatus(updateStatus healthStatusUpdater, health componentHealth) error {
-	if updateStatus == nil {
-		return nil
-	}
-	return updateStatus(health)
 }
 
 // types/dto candidate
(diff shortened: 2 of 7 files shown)


pkg/postgresql/cluster/business/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
-	if postgresCluster.Status.Phase != nil {
-		oldPhase = *postgresCluster.Status.Phase
+	/*
+				rewrite to consider taking state of other objects into account
+				before declaring readyness.
+
+		CNPG cluster
+		Poolers
+		Access resources (configmap and secret)
+		And all of them needs to be set to Ready for our PostgresCluster phase to become Ready?
```
limak9182: yes, that's correct

pkg/postgresql/cluster/business/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
 
-	// Final status sync.
-	var oldPhase string
-	if postgresCluster.Status.Phase != nil {
-		oldPhase = *postgresCluster.Status.Phase
+	/*
+				rewrite to consider taking state of other objects into account
+				before declaring readyness.
+
+		CNPG cluster
```
limak9182: Correct me if I'm wrong, but what I'm missing in the code is checking if CNPG cluster is ready and if yes then updating our ClusterReady condition to true, so at the end (here) we can check if all conditions are true and set whole custom resource status ready to true

pkg/postgresql/cluster/business/core/cluster.go
limak9182: Here we are updating the status with clusterReady condition False if we patched CNPG Cluster and requeue.
Shouldnt we in the next reconciliation go again to this check again if CNPG cluster is in desired state:
```go
!equality.Semantic.DeepEqual(currentNormalized, desiredNormalized)
```
and if it is (we are not going inside this if), set the clusterReady condition to true? 
Something like:
```go
statusErr := updateStatus(clusterReady, metav1.ConditionTrue...
```
just after this big `if` block?

Maybe one more CNPG cluster check is needed, just to be sure it's in healthy and ready state, if not then requeue or leave with error?

limak9182: actually after thinking about it, it should be probably after `reconcileManagedRoles` as it's the last thing we are doing with CNPG cluster.

pkg/postgresql/cluster/business/core/cluster.go
M4KIF: Currently? Yes, currently It's mostly a scaffolding to which I will place any business logic.

pkg/postgresql/cluster/business/core/cluster.go
M4KIF: In general I think It's a bit misleading that we do cluster ready and condtion == false. At least for me, It should be sth like ClusterErrorRetry, like http codes.

limak9182: Sorry I think I don't not fully understand that. By cluster do you mean our PostgresClusterCR or Cluster Ready condition (actual CNPG cluster)?

mploski: > In general I think It's a bit misleading that we do cluster ready and condtion == false. At least for me, It should be sth like ClusterErrorRetry, like http codes. 

This is standard pattern in k8s, we should stick to it. Take a look at our design docs where we map phases and conditions to current situation

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	// 		copy(poolerOldConditions, postgresCluster.Status.Conditions)
+	// 		_ = syncPoolerStatus(ctx, c, postgresCluster)
+	// 		rc.emitPoolerReadyTransition(postgresCluster, poolerOldConditions)
+	// 	}
+	// }
+	// logger.Info("Reconciliation complete")
+	// return ctrl.Result{}, nil
+}
+
+type clusterReadynessCheck interface {
```
mploski: sound like a great thing to add to the specific port capabilities :-)

pkg/postgresql/cluster/business/ports/secondary/provisioner.go
```diff
@@ -0,0 +1,12 @@
+package secondary
+
+import (
+	pgcConstants "github.com/splunk/splunk-operator/pkg/postgresql/cluster/business/core/types/constants"
+)
+
+type Provisioner interface {
```
mploski: I think our secondary ports should reflect that we create cluster and database and we should map our interfaces around it.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+
+		CNPG cluster
+		Poolers
+		Access resources (configmap and secret)
+		And all of them needs to be set to Ready for our PostgresCluster phase to become Ready?
+
+	*/
+
+	// basically a sync logic
+	state := pgcConstants.EmptyState
```
mploski: I think the idea here was to decouple status check from cnpg status, At the same time we also check health after every stage and we move forward only if we are ok, if it is still in progress we requeue or raise error. Here we have a code we can use to check where we are with status iteratively, but I dont see yet how it solve our core problem

internal/controller/postgrescluster_controller.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	cnpgv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
 	enterprisev4 "github.com/splunk/splunk-operator/api/v4"
-	clustercore "github.com/splunk/splunk-operator/pkg/postgresql/cluster/core"
+	clustercore "github.com/splunk/splunk-operator/pkg/postgresql/cluster/business/core"
```
mploski: tbh business string is redundat here. Core itself is already a domain

pkg/postgresql/cluster/business/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+
+type clusterReadynessCheck interface {
+	Condition() (pgcConstants.State, error)
+}
+
+type provisionerHealthCheck struct {
+	cnpgCluster *cnpgv1.Cluster
+}
+
+func (c *provisionerHealthCheck) Condition() (pgcConstants.State, error) {
```
mploski: I think all of this conditions check should be part of our Ports, also how you want to map condition to phase?

pkg/postgresql/cluster/business/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	return pgcConstants.ProvisionerReady, nil
+}
+
+type poolerHealthCheck struct {
+	rwPooler *cnpgv1.Pooler
+	roPooler *cnpgv1.Pooler
+}
+
+func (p *poolerHealthCheck) Condition() (pgcConstants.State, error) {
+	return pgcConstants.PoolerReady, nil
```
mploski: how do we want to check the actual condition component has in status?

pkg/postgresql/cluster/core/types/constants/state.go
```diff
@@ -0,0 +1,41 @@
...
+
+	ClusterReady
+	ClusterPending
+	ClusterProvisioning
+	ClusterConfiguring
+	ClusterFailed
+)
+
+const (
+	ComponentsReady = PoolerReady | ProvisionerReady | SecretReady | ConfigMapReady
```
mploski: what we try to achieve here? Is it a bitmask? Since you use IOTA we endup in having just random integer?
That would be probably simpler to just use struct with keeping the state like that:
``` 
 type ClusterState struct {                                                                                                                                         
      Provisioner ComponentPhase                                                                                                                                     
      Pooler      ComponentPhase                                                                                                                                     
      ConfigMap   ComponentPhase                                                                                                                                     
      Secret      ComponentPhase                                                                                                                                     
  }  
```

pkg/postgresql/cluster/business/core/cluster.go
```diff
@@ -445,45 +451,119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
-		}, roPooler)
-		if rwErr == nil && roErr == nil && arePoolersReady(rwPooler, roPooler) {
-			logger.Info("Poolers ready, syncing status")
-			poolerOldConditions := make([]metav1.Condition, len(postgresCluster.Status.Conditions))
-			copy(poolerOldConditions, postgresCluster.Status.Conditions)
-			_ = syncPoolerStatus(ctx, c, postgresCluster)
-			rc.emitPoolerReadyTransition(postgresCluster, poolerOldConditions)
-		}
+
+	if state&pgcConstants.ComponentsReady != 0 {
```
mploski: Not sure if this logic is not broken. What happens if we set ProvisionerReady but later in stage we set failed for something. Sue to way we set state and components this would pass. I think it is because iota incrementing by 1 not by power of 2? I think we should rely not on bitmasking here, but simple struct with state for every stage and if all is good we are good

pkg/postgresql/cluster/core/cluster.go
M4KIF: Then the interface here would be implemented by the ports. Clean and nice idea

pkg/postgresql/cluster/core/cluster.go
M4KIF: we build our state here, after checking all ports for readyness/not dying there is the moment for us to decide what happened and how It happened. 
We don't set our state as state == cnpgVariable mapped to ours, we decide what do we want to do with the fact that cm's, secrets, provisioner etc. are ready.

internal/controller/postgrescluster_controller.go
M4KIF: I agree It's redundant, here It's a tradeoff for verbosity and segregation of components. And a service pattern at once, ie. the service/ is the primary port (reconciler that we provide) implementation. core/ is core, and ports/ are the contracts that we need for the core to work. They can grow large, hence the whole separate dir for ports.

pkg/postgresql/cluster/business/core/cluster.go
M4KIF: I agree, they were placed here as what I need. Solving It like you say is the thing I'm hoping for. For the provisioner/cluster etc. ports to include an interface for checking It's state. 
---
Then the adapter would be essentially mapping the dependency state to our abstraction of It's state. Ie. we have cluster ready, provisioning, failover, cupcake, coffee etc.
Mapping condition to phase would be the job of the facade, ie. the cluster.go. That would be the whole operational brain behind. Ie. lot's of individual pieces funneled into our business decisions.

M4KIF: Sth. like a stateMapper, or another objects that specialises in deciding on what phase/condition we're in could also be born. The bit mask could be used for covering the phase 1, ie. state = FinaliserNotAdded && !ClusterProvisioning etc.
Phase 2 -> state == Finaliser && ClusterReady etc.

pkg/postgresql/cluster/business/core/cluster.go
M4KIF: wdym? It's kind of the job of the adapter to test and provide that the state is actual. Ie. If we place this as a method of a port, and implement It via adapters. We actually won't work on the real state of the component in our core. Only on our understanding of It.

M4KIF: Currently It would be just to copy paste the thing that we do inside cluster.go, ie. the resource obj. of the Pooler, k8s.Get(obj, ...) and all similar. As there is no abstraction currently.

pkg/postgresql/cluster/core/types/constants/state.go
M4KIF: It's a bitmask. And It does the same job of keeping a struct with aditional field. 
And It kinda solves the case of having to create new types just for each component. 
Just adding additional states to the state machine, ie. the values in the "enum". The iota usage is an enum in go: https://yourbasic.org/golang/iota/
It's the first usage in this file, hence It's basically an enum from 0

pkg/postgresql/cluster/business/core/cluster.go
M4KIF: with unsetting the bit's at any space, this condition starts failing. As well as with not setting the bits, the values don't AND, hence if we mark ProvisionerReady and then set masks for ConfigMapFailed, it won't fire. 
And to prove how this logic would work, there would be tests that would make sure any misfires aren't possible.

pkg/postgresql/cluster/core/types/constants/state.go
mploski: > And It kinda solves the case of having to create new types just for each component.

We already create types for a new component for many reasons, so what problem it really solve? I agree it is smart way of doing this, but neverthless if new component arise you need to add it to the const and extend types. I feel like we trade go readability for a really small c-like optimisation especially with this model of bitwise comparision later:

```
state |= componentHealth.State
if state&pgcConstants.ComponentsReady == pgcConstants.ComponentsReady
```
Also, if we build state incrementally it means that the LAST successful state in state machine is an final success. With this in mind we dont need to check every other component state afterwards.

Can we do this simpler so when Im wake up at 5am in the morning I easily understand the code?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
 		}
 	}
 
-	// Final status sync.
-	var oldPhase string
-	if postgresCluster.Status.Phase != nil {
-		oldPhase = *postgresCluster.Status.Phase
+	// Aggregate component readiness from iterative health checks.
+	state := pgcConstants.EmptyState
+	conditions := []clusterReadynessCheck{
```
mploski: so after every phase that is not immediate like cluster creation we should also incorporate state check rediness. I think we discussed that we dont really need to check at the end assuming we check intermediary statuses per phase?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
-	if err := syncStatus(ctx, c, postgresCluster, cnpgCluster); err != nil {
-		logger.Error(err, "Failed to sync status")
-		if apierrors.IsConflict(err) {
-			logger.Info("Conflict during status update, will requeue")
-			return ctrl.Result{Requeue: true}, nil
+
+	for _, check := range conditions {
+		componentHealth, err := check.Condition(ctx)
+		if err != nil {
+			if statusErr := updateStatus(componentHealth.Condition, metav1.ConditionFalse, componentHealth.Reason, componentHealth.Message, componentHealth.Phase); statusErr != nil {
```
mploski: If we run this at the end of reconcillation it seems that some of the code is dead i.e if we are here, we cannot have a configmap or secret orphaned. If we do it should be discovered during this phase and requeue/err

pkg/postgresql/cluster/core/cluster.go
M4KIF: I agree with that, but Isn't then the scope == refactor the reconciler?
I've tried to stick with changing the sync logic and doing the ground prep for more changes in coming tickets and potential p&a rework.

pkg/postgresql/cluster/core/types/constants/state.go
M4KIF: I agree that this state check, after the iteration health check passes is redundant here.
And taking into consideration the potential future work. Which could include a file division.
I could expand the *healthCheck types with them returning the *(component)StateDto instead of relying on generic state bits.

M4KIF: Because later, in the very near future It seems that the project could follow this footstep of having some separation in phases and It's crucial elements. Like we've discussed on the p&a ideas brainstorm.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+				if apierrors.IsConflict(statusErr) {
+					return ctrl.Result{Requeue: true}, nil
+				}
+				return ctrl.Result{}, statusErr
+			}
+			logger.Error(err, "Component health check reported issues",
```
mploski: Please follow our logging strategy: https://splunk.atlassian.net/wiki/spaces/CCP/pages/1079831167399/PostgreSQL+Controllers+Logging+Strategy

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+				"component", check,
+				"requeueAfter", componentHealth.Result.RequeueAfter,
+				"reason", componentHealth.Reason)
+			return componentHealth.Result, err
+		}
+
+		if isPendingState(componentHealth.State) {
```
mploski: If we run this code on every phase separately, here we should requeue

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+			return componentHealth.Result, nil
 		}
-		return ctrl.Result{}, fmt.Errorf("failed to sync status: %w", err)
+		state |= componentHealth.State
 	}
-	var newPhase string
-	if postgresCluster.Status.Phase != nil {
-		newPhase = *postgresCluster.Status.Phase
+
+	if state&pgcConstants.ComponentsReady == pgcConstants.ComponentsReady {
```
mploski: thois could be potentially method on a state so it reads natually at 4 am i.e                                
  func (s State) HasAll(required State) bool {      
      return s&required == required           
  }                                                                                                        
  if state.HasAll(pgcConstants.ComponentsReady) {}

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -445,45 +447,475 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+type provisionerHealthCheck struct {
+	cluster     *enterprisev4.PostgresCluster
+	cnpgCluster *cnpgv1.Cluster
+}
+
+func newProvisionerHealthCheck(cluster *enterprisev4.PostgresCluster, cnpgCluster *cnpgv1.Cluster) *provisionerHealthCheck {
+	return &provisionerHealthCheck{cluster: cluster, cnpgCluster: cnpgCluster}
+}
+
+func (c *provisionerHealthCheck) Condition(_ context.Context) (StateInformationDto, error) {
```
mploski: I like providing interface like that! One doubt I have is about pureness and responsiblity of those condition methods. They check conditions but also fetch k8s objects. IMO k8s objects should be evaluated and the reconciller kickoff and propagated

mploski:
I like the direction we are heading two! 
Few concerns two discuss:
1. Should we aggregate the status check at the end or we should check per phase and when we pass everything in reconcile then we know we are good. Considering we build state incrementally and sequentally this make sense IMO
2. Function pureness
3. bitmask usage in that form - if 1 is true then we might not need it. If we need it mabe we can wrap it in helper functions/refactor the approach so we understand the code better in the morning when we have P0
4. I believe we should start with drowing current state machine and how we set a phase and conditions. Did you do that?

M4KIF:
> I like the direction we are heading two! Few concerns two discuss:
> 
> 1. Should we aggregate the status check at the end or we should check per phase and when we pass everything in reconcile then we know we are good. Considering we build state incrementally and sequentally this make sense IMO
> 2. Function pureness
> 3. bitmask usage in that form - if 1 is true then we might not need it. If we need it mabe we can wrap it in helper functions/refactor the approach so we understand the code better in the morning when we have P0
> 4. I believe we should start with drowing current state machine and how we set a phase and conditions. Did you do that?

1. I see and understand that, per phase check seems reasonable. Double checking is "majtki na majtki"
2. yup, health which does business.
3. simplified that a bit
4. I didn't draw the current state machine.

pkg/postgresql/cluster/core/cluster.go
M4KIF: I've proposed another approach. Which is quite similiar to the current tbh

pkg/postgresql/cluster/core/cluster.go
M4KIF: after changing the state idea, It currently isn't used that much.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -444,48 +467,493 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		ctx, configMapComponent.Name(),
+		updateStatus, configMapHealth,
+		configMapSyncErr); err != nil || result != (ctrl.Result{}) {
+		return result, err
 	}
-	if err := syncStatus(ctx, c, postgresCluster, cnpgCluster); err != nil {
-		logger.Error(err, "Failed to sync status")
+
+	logger.Info("Reconciliation complete")
+	if err := updateStatus(clusterReady, metav1.ConditionTrue, reasonCNPGClusterHealthy, msgAllComponentsReady, readyClusterPhase); err != nil {
```
vivekr-splunk: This refactor appears to drop cluster phase transition events entirely. Before the change, the controller captured the old/new phase around the final status sync and called `rc.emitClusterPhaseTransition(...)`; that helper still exists in `events.go`, but I no longer see it invoked on the happy path or on degraded transitions. From the product point of view, that means users lose the `ClusterReady` / `ClusterDegraded` event stream even though status still changes, which looks like a behavior regression rather than just an internal refactor.

vivekr-splunk (changes requested):
I found one blocking regression in the status/event flow.

pkg/postgresql/cluster/core/cluster.go
M4KIF: I agree, I've missed this when attempting the rewrite. Thank you for your input and I agree, I've split the logic and focused only on the buzzwords/keywords and totally went dark on observability/event part.

pkg/postgresql/cluster/core/types.go
mploski, line 82: I see that the condition types SecretReady, ManagedRolesReady, and ConfigMapReady have not been added.

At the moment, we only define clusterReady and poolerReady. The functions secretModel.Converge, configMapModel.Converge, and the provisioner all update clusterReady, which means the status does not reflect the more granular state described in the ticket.

According to the ticket, the expected status snapshot should be:

SecretReady       = True  (SuperUserSecretReady)
ManagedRolesReady = True  (ManagedRolesReconciled)
ConfigMapReady    = True  (ConfigMapReconciled)

The reason values also need adjustment. Currently, secretModel.Converge uses:

reasonClusterBuildSucceeded for the success case, and
reasonUserSecretFailed for the "ref not yet written" pending case.

Both are inherited from the old clusterReady vocabulary. However, the ticket specifies that the reasons should be SuperUserSecretReady and SuperUserSecretFailed, so these should be updated accordingly.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
 			}
-			return ctrl.Result{}, err
+			p.events.emitNormal(p.cluster, EventClusterUpdateStarted, "CNPG cluster spec updated, waiting for healthy state")
 		}
-		rc.emitNormal(postgresCluster, EventPoolerCreationStarted, "Connection poolers created, waiting for readiness")
-		logger.Info("Connection pooler creation started, requeueing")
-		if statusErr := updateStatus(poolerReady, metav1.ConditionFalse, reasonPoolerCreating,
-			"Connection poolers are being provisioned", provisioningClusterPhase); statusErr != nil {
-			logger.Error(statusErr, "Failed to update status")
+		if rolesErr := reconcileManagedRoles(ctx, p.client, p.cluster, p.cnpgCluster); rolesErr != nil {
```
mploski: ManagedRoles (Step 3) is currently handled inside provisionerModel.Actuate, but it would be clearer if it were implemented as a separate pipeline step.

Right now:

There is no ManagedRolesReady condition.
There is no distinct Step 3 in the pipeline.
Observers cannot distinguish between a failure in provisioning CNPG and a failure during role reconciliation.

Separating this into its own step would make the pipeline state clearer and improve observability of failures and it is also one of the acceptance criteria :-)

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+			"condition", health.Condition,
+			"reason", health.Reason,
+			"phase", health.Phase)
+		if result != (ctrl.Result{}) {
+			componentLogger.Info("Component requested explicit result",
+				"step", "converge",
+				"requeueAfter", result.RequeueAfter)
+			return result, nil
 		}
+		return result, nil
```
mploski: Thing to fix once we split phases per step.

manageComponentHealth never writes Condition=True on the success path.

The function has three branches:

error → writes False
pending state → writes False
ready → writes nothing

As a result, the pipeline only ever stamps conditions as False. The only place where Condition=True is written during reconciliation is the final call.

This happens only after all steps succeed, and it updates clusterReady only.

This breaks the audit-trail invariant described in the ticket. After a fully healthy reconcile, conditions like SecretReady, ManagedRolesReady, and ConfigMapReady never appear in etcd.

If CNPG enters a switchover during the next reconcile, the provisioner returns early with ClusterReady=False, and those provisioning conditions are still absent, not True. As a result, an operator running kubectl describe cannot distinguish between:

“the secret was provisioned successfully and is still healthy”, and
“the secret provisioning step has never run.”

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	// TODO: Port material.
+	rwExists, err := poolerExists(ctx, p.client, p.cluster, readWriteEndpoint)
+	if err != nil {
+		p.events.emitWarning(p.cluster, EventPoolerReconcileFailed, fmt.Sprintf("Failed to sync pooler status: %v", err))
+		p.health.State = pgcConstants.Failed
+		p.health.Reason = reasonPoolerReconciliationFailed
+		p.health.Message = fmt.Sprintf("Failed to check RW pooler existence: %v", err)
+		p.health.Phase = failedClusterPhase
+		p.health.Result = ctrl.Result{}
+		return p.health, err
```
mploski, line 951: previously we were setting:
```
  cluster.Status.ConnectionPoolerStatus = &ConnectionPoolerStatus{Enabled: true}                                                                         
```
when everything was fine, currently we miss it. We use it in database.go to check if we should use connection pooler. Without it we will be always using direct cnpg endpoint

mploski: we should guard this logic in unit tests/integration tests against future regressions in contract

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+			return result, err
+		}
+		if result != (ctrl.Result{}) {
+			return result, nil
 		}
 	}
 
-	// Build desired CNPG Cluster spec.
-	desiredSpec := buildCNPGClusterSpec(mergedConfig, postgresSecretName)
+	provisionerComponent := components[1].(*provisionerModel)
```
mploski: this seem to be fragile if we change order.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+
+	s.health.State = pgcConstants.Ready
+	s.health.Reason = reasonClusterBuildSucceeded
+	s.health.Message = msgSuperuserSecretReady
+	s.health.Phase = readyClusterPhase
+	s.health.Result = ctrl.Result{}
+	return s.health, nil
+}
+
+func isPendingState(state pgcConstants.State) bool {
```
mploski: The function name is bit misleading as it returns true for Pending, Provisioning, and Configuring. The name implies only the first.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+}
+
+func newProvisionerModel(c client.Client, scheme *runtime.Scheme, events eventEmitter, cluster *enterprisev4.PostgresCluster, mergedConfig *MergedConfig, secretName string) *provisionerModel {
+	return &provisionerModel{client: c, scheme: scheme, events: events, cluster: cluster, mergedConfig: mergedConfig, secretName: secretName}
+}
+
+func (p *provisionerModel) Name() string { return pgcConstants.ComponentProvisioner }
+
+func (p *provisionerModel) EvaluatePrerequisites(_ context.Context) (prerequisiteDecision, error) {
+	if p.cluster.Status.Resources == nil || p.cluster.Status.Resources.SuperUserSecretRef == nil {
```
mploski: we might still be in provision state where secrets is being created and thats why we dont have it in the status. this reason can be misleading

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		return prerequisiteDecision{
+			Allowed: false,
+			Health: componentHealth{
+				State:     pgcConstants.Pending,
+				Condition: clusterReady,
```
mploski: condition -> secretReady

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +141,846 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	health componentHealth
+}
+
+func newConfigMapModel(c client.Client, scheme *runtime.Scheme, events eventEmitter, runtime clusterRuntimeView, cluster *enterprisev4.PostgresCluster, secret string) *configMapModel {
+	return &configMapModel{client: c, scheme: scheme, events: events, runtime: runtime, cluster: cluster, secret: secret}
+}
+
+func (c *configMapModel) Name() string { return pgcConstants.ComponentConfigMap }
+
+func (c *configMapModel) EvaluatePrerequisites(_ context.Context) (prerequisiteDecision, error) {
```
mploski, line 1037: I think this gate can never actually fire, If CNPG is not healthy, provisionerModel.Converge returns a pending/failed state → phase() returns a non-zero result → the first loop returns early.

Also this should set its own condition not clusterReady

internal/controller/postgrescluster_controller_test.go
```diff
@@ -51,6 +54,18 @@ import (
 * PC-09 ignores no-op updates
  */
 
+func containsEvent(events []string, recorder *record.FakeRecorder, eventType string, event string) bool {
```
mploski: Not sure if I read it right but The helper reads exactly one event from the channel per call and returns on the first read regardless of whether it matches. If the target event is the
   second event in the channel, the Eventually loop never finds it.

M4KIF: the eventually loop calls It repeatedly until it finds or times out

pkg/postgresql/cluster/core/cluster.go
M4KIF: Yes, that's the case under usual condition. It's a double caution safety. So It can be deleted if the system is to be trusted enough. My bet would be to leave It for now. 
Condition altered.

internal/controller/postgrescluster_controller_test.go
mploski: why not to do this in one run? I.e pick all events and iterate over them. In current situation received can be also not accurate i.e recorder.Events != received if you return true early. Plus if we take one event why for :-)

pkg/postgresql/cluster/core/cluster.go
mploski: ok can you share what would be the case it fires? My understanding is that this is dead code, what do I miss? If this is dead code, we simply dont need it

internal/controller/postgrescluster_controller_test.go
mploski: Ah I just realised that the received will not be accurate as we do copy inside of the function by appending. we should pass * if you want to update this value

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	cnpgCluster = provisionerComponent.cnpgCluster
+	runtimeView := provisionerRuntimeView{model: provisionerComponent}
+	oldConditions := append([]metav1.Condition(nil), postgresCluster.Status.Conditions...)
+	runtimeComponents := []componentModel{
+		newManagedRolesModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
+		newPoolerModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, mergedConfig, cnpgCluster, poolerEnabled, poolerConfigPresent),
+		newConfigMapModel(c, rc.Scheme, rc, updateComponentHealthStatus, runtimeView, postgresCluster, postgresSecretName),
+	}
+
+	for _, component := range runtimeComponents {
```
mploski, line 267: why we are not iterating over full list of components in line 180?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
-		return ctrl.Result{}, secretErr
+	poolerEnabled = mergedConfig.Spec.ConnectionPoolerEnabled != nil && *mergedConfig.Spec.ConnectionPoolerEnabled
+	poolerConfigPresent := mergedConfig.CNPG != nil && mergedConfig.CNPG.ConnectionPooler != nil
+
+	secretComponent := newSecretModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, postgresSecretName)
+	provisionerComponent := newProvisionerModel(c, rc.Scheme, rc, updateComponentHealthStatus, postgresCluster, mergedConfig, postgresSecretName)
+
+	components := []componentModel{
+		secretComponent,
+		provisionerComponent,
```
mploski: Id rather name it clusterComponent not provisionerComponent to provide better code readability. Provisioner is notion of external postgres provider that can do many things including cluster,roles,database, pooler etc

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+type componentHealth struct {
+	State     pgcConstants.State
+	Condition conditionTypes
+	Reason    conditionReasons
+	Message   string
+	Phase     reconcileClusterPhases
+	Result    ctrl.Result
+}
+
+type componentModel interface {
```
mploski: I wonder if we cannot merge this interfaces into one i.e

```
type component interface {
Actuate(ctx context.Context) error
Converge(ctx context.Context) (componentHealth, error)
EvaluatePrerequisites(ctx context.Context) (prerequisiteDecision, error)
Namer() string
}
```

they seem to be logically bounded and I dont see a use-case to use this interfaces in separation i.e we cannot have component that doesnt have function to actuate, converge etc.We can always split it later if it grows too much, or we have different usecase

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		m.health.State = pgcConstants.Failed
+		m.health.Reason = reasonManagedRolesFailed
+		m.health.Message = fmt.Sprintf("Managed roles reconciliation failed for %d role(s)", len(status.Failed))
+		m.health.Phase = failedClusterPhase
+		m.health.Result = ctrl.Result{RequeueAfter: retryDelay}
+		return m.health, fmt.Errorf("managed roles have failed entries")
+	}
+
+	if len(status.Pending) > 0 {
+		m.health.State = pgcConstants.Failed
```
mploski: this should be  Provisioning state, and it is also not an error

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		s.health.State = pgcConstants.Failed
+		s.health.Reason = reasonSuperUserSecretFailed
+		s.health.Message = fmt.Sprintf(msgFmtSecretMissingKey, refKey)
+		s.health.Phase = failedClusterPhase
+		s.health.Result = ctrl.Result{}
+		return s.health, fmt.Errorf("secret missing key %s", refKey)
+	}
+
+	s.health.State = pgcConstants.Ready
+	s.health.Reason = reasonClusterBuildSucceeded
```
mploski: this should be probably reasonSuperUserSecretReady not reasonClusterBuildSucceeded

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -140,351 +171,1119 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+			return c.health, fmt.Errorf("configmap missing key %s", requiredKey)
 		}
 	}
 
-	// Final status sync.
-	var oldPhase string
-	if postgresCluster.Status.Phase != nil {
-		oldPhase = *postgresCluster.Status.Phase
+	c.health.State = pgcConstants.Ready
+	c.health.Reason = reasonClusterBuildSucceeded
```
mploski: this should be probably reasonConfigMapReady reason

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -755,90 +1554,6 @@ func deleteConnectionPoolers(ctx context.Context, c client.Client, cluster *ente
...
-	roPooler := &cnpgv1.Pooler{}
-	if err := c.Get(ctx, types.NamespacedName{
-		Name:      poolerResourceName(cluster.Name, readOnlyEndpoint),
-		Namespace: cluster.Namespace,
-	}, roPooler); err != nil {
-		return err
-	}
-
-	cluster.Status.ConnectionPoolerStatus = &enterprisev4.ConnectionPoolerStatus{Enabled: true}
-	rwDesired, rwScheduled := poolerInstanceCount(rwPooler)
```
mploski, line 749: i see you removed poolerInstanceCount usage, but did we delete the function?

mploski:
@M4KIF did a review + additionally passed pr through LLM. Few points brought:
  
A. managedRolesModel.EvaluatePrerequisites uses State=Failed for a blocked gate                                                                              
                                                                                                                                                                 
  func (m *managedRolesModel) EvaluatePrerequisites(_ context.Context) (prerequisiteDecision, error) {                                                           
      if m.runtime == nil || !m.runtime.IsHealthy() {                                                                                                            
          return prerequisiteDecision{                                                                                                                           
              Allowed: false,                                                                                                                                    
              Health: componentHealth{                                                                                                                           
                  State:     pgcConstants.Failed,       // ← wrong                                                                                               
                  Condition: managedRolesReady,                                                                                                                  
                  Reason:    reasonManagedRolesFailed,                                                                                                           
                  Phase:     failedClusterPhase,                                                                                                                 
                  Result:    ctrl.Result{RequeueAfter: retryDelay},                                                                                              
                                                                                                                                                                 
  The gate blocks because CNPG isn't healthy yet — that's a transient wait, not a failure. But because the phase() driver calls component.Converge(ctx) when a   
  gate blocks, and Converge also checks !m.runtime.IsHealthy() and returns State=Failed with err != nil, this path returns a non-nil error from phase() and      
  propagates up as a hard failure. The controller will log an error and return from the reconcile with an error, which triggers exponential backoff rather than  
  the RequeueAfter: retryDelay specified in the health. The state should be Provisioning with a requeue, matching the pattern every other blocked gate uses.   
                                                                                                                                                               
  B. managedRolesModel.Converge duplicates the prerequisite check and returns an error for a wait state                                                          
  
  func (m *managedRolesModel) Converge(ctx context.Context) (health componentHealth, err error) {                                                                
      ...                                                                                                                                                        
      if m.runtime == nil || !m.runtime.IsHealthy() {                                                                                                            
          m.health.State = pgcConstants.Failed                                                                                                                   
          ...                                                                                                                                                    
          return m.health, fmt.Errorf("managed roles blocked until CNPG cluster is healthy")                                                                     
      }                                                                                                                                                          
                                                                                                                                                               
This is the same condition as the prerequisite gate — structurally unreachable when called normally (gate would have blocked), but reached in the blocked-gate 
  path where phase() calls Converge directly when !gate.Allowed. The err != nil causes phase() to return an error rather than health.Result. This results in   
  exponential backoff for a completely normal "wait for CNPG to be healthy" state.                                                                               
                                                                                                                                                               
  Compare how the provisioner handles this: its blocked gate path sets State=Pending, err=nil so the driver returns the requeue result cleanly.                  
                                                                                                                                 
 C. configMapModel.EvaluatePrerequisites is structurally unreachable but still has a blocking problem                                                           
  
  The gate is now dead code (structurally can't fire since the provisioner gates the first loop). However it uses configMapsReady (correct  condition type). The dead code concern remains — if the pipeline is reorganised, the gate will incorrectly return early with State=Provisioning and then the blocked gate path in phase() calls Converge, which also checks !IsHealthy() and returns State=Provisioning, err=nil,  safely causing a requeue.

pkg/postgresql/cluster/core/cluster.go
M4KIF: It mostly dead code, shouldn't be fired. So yeah, deleted.

internal/controller/postgrescluster_controller_test.go
M4KIF: i agree with the slice pointer. I didn't understand, but yeah, the fake recorder is just a channel to which we write some strings.

M4KIF: eventually + for single exit seems like an overkill, as It's an await for a real async scenario.
But, we write to the channel, w/o async. So a for w/ contains is enough

pkg/postgresql/cluster/core/cluster.go
M4KIF: because managedRoles, Pooler, ConfigMap models need some information from previous stages and injecting It needs to happen in-between. 
The later phases should utilise ports in later iterations of the source to use the information they need to function.

pkg/postgresql/cluster/core/cluster.go
M4KIF: yeah, a good idea tbh

pkg/postgresql/cluster/core/cluster.go
M4KIF: I agree, I tend to go over the board with verbosity. I agree component containing the above methods is bound, logical and consistent.

pkg/postgresql/cluster/core/cluster.go
M4KIF: ack

pkg/postgresql/cluster/core/cluster.go
M4KIF: yeah, as we saw in the kubectl describe. Thanks! adjusted

pkg/postgresql/cluster/core/cluster.go
M4KIF: ack

pkg/postgresql/cluster/core/cluster.go
M4KIF: It was left hanging in the units. Deleted It

M4KIF:
⏳ **CLA signed** — now checking Code of Conduct status...

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -128,302 +159,1111 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	case !p.poolerConfigPresent:
+		return nil
+	case p.cnpgCluster == nil || p.cnpgCluster.Status.Phase != cnpgv1.PhaseHealthy:
+		return nil
+	default:
+		if err := createOrUpdateConnectionPoolers(ctx, p.client, p.scheme, p.cluster, p.mergedConfig, p.cnpgCluster); err != nil {
+			p.events.emitWarning(p.cluster, EventPoolerReconcileFailed, fmt.Sprintf("Failed to reconcile connection pooler: %v", err))
+			return err
+		}
+		p.events.emitNormal(p.cluster, EventPoolerCreationStarted, "Connection poolers created, waiting for readiness")
```
limak9182: correct me if I'm wrong but is this event emitted on every reconcile? Normal events should be emitted only once. In this case only when pooler is actually created

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -128,302 +159,1111 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+				return fmt.Errorf("setting controller reference: %w", setErr)
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		c.events.emitWarning(c.cluster, EventConfigMapReconcileFailed, fmt.Sprintf("Failed to reconcile ConfigMap: %v", err))
+		return err
 	}
+	if op == controllerutil.OperationResultCreated {
```
DmytroPI-dev: we can use tagged switch on op, like this: 
```
	switch op {
case controllerutil.OperationResultCreated:
		c.events.emitNormal(c.cluster, EventConfigMapReady, fmt.Sprintf("ConfigMap %s created", desiredCM.Name))
	case controllerutil.OperationResultUpdated:
		c.events.emitNormal(c.cluster, EventConfigMapReady, fmt.Sprintf("ConfigMap %s updated", desiredCM.Name))
	}
```

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -128,302 +159,1111 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		c.health.Result = ctrl.Result{}
+		return c.health, err
+	}
+
+	requiredKeys := []string{
+		configKeyClusterRWEndpoint,
+		configKeyClusterROEndpoint,
+		configKeyDefaultClusterPort,
+		configKeySuperUserSecretRef,
+	}
```
DmytroPI-dev, line 1129: We have `configKeyClusterREndpoint` in types.go ,which looks unused.

pkg/postgresql/cluster/core/types.go
```diff
@@ -52,9 +53,17 @@ const (
...
-	superUsername       string = "postgres"
-	defaultPort         string = "5432"
+	defaultDatabaseName         string = "postgres"
+	superUsername               string = "postgres"
+	defaultPort                 string = "5432"
+	configKeyClusterRWEndpoint  string = "CLUSTER_RW_ENDPOINT"
+	configKeyClusterROEndpoint  string = "CLUSTER_RO_ENDPOINT"
+	configKeyClusterREndpoint   string = "CLUSTER_R_ENDPOINT"
+	configKeyDefaultClusterPort string = "DEFAULT_CLUSTER_PORT"
+	configKeySuperUserName      string = "SUPER_USER_NAME"
```
DmytroPI-dev, line 64: Looks like this is not used anymore with your changes?

pkg/postgresql/cluster/core/types.go
```diff
@@ -52,9 +53,17 @@ const (
...
+	configKeySuperUserSecretRef string = "SUPER_USER_SECRET_REF"
+	configKeyPoolerRWEndpoint   string = "CLUSTER_POOLER_RW_ENDPOINT"
+	configKeyPoolerROEndpoint   string = "CLUSTER_POOLER_RO_ENDPOINT"
```
DmytroPI-dev, line 67: Isn't it not used anymore?

pkg/postgresql/cluster/core/types.go
```diff
@@ -77,19 +86,27 @@ const (
...
 
 	// condition reasons — clusterReady
 	reasonClusterClassNotFound  conditionReasons = "ClusterClassNotFound"
+	reasonManagedRolesReady     conditionReasons = "ManagedRolesReconciled"
+	reasonManagedRolesPending   conditionReasons = "ManagedRolesPending"
 	reasonManagedRolesFailed    conditionReasons = "ManagedRolesReconciliationFailed"
 	reasonClusterBuildFailed    conditionReasons = "ClusterBuildFailed"
 	reasonClusterBuildSucceeded conditionReasons = "ClusterBuildSucceeded"
 	reasonClusterGetFailed      conditionReasons = "ClusterGetFailed"
 	reasonClusterPatchFailed    conditionReasons = "ClusterPatchFailed"
```
DmytroPI-dev, line 103: These looks like unused, same for `reasonCNPGClusterNotHealthy  conditionReasons = "CNPGClusterNotHealthy"`

pkg/postgresql/cluster/core/cluster.go
M4KIF: yes, in this iteration It hasn't been guarded from firing every time.

pkg/postgresql/cluster/core/types.go
```diff
@@ -114,4 +131,39 @@ const (
...
+	msgCNPGPromoting             statusMessage = "Replica is being promoted to primary"
+	msgCNPGWaitingForUser        statusMessage = "Action from the user is required"
+	msgCNPGUnrecoverable         statusMessage = "Cluster failed, needs manual intervention"
+	msgCNPGCannotCreateObjects   statusMessage = "Cluster resources cannot be created"
+	msgFmtCNPGPluginError        statusMessage = "CNPG plugin error: %s"
+	msgFmtCNPGImageError         statusMessage = "CNPG image error: %s"
+	msgFmtCNPGClusterPhase       statusMessage = "CNPG cluster phase: %s"
+
+	// status messages — aggregate and component readiness checks
+	msgAllComponentsReady             statusMessage = "All components are ready"
```
DmytroPI-dev, line 153: This one is unused anymore, I guess?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1129 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		statusErr := writeComponentStatus(m.updateStatus, m.health)
+		if statusErr != nil {
+			if err != nil {
+				err = errors.Join(err, statusErr)
+			} else {
+				err = statusErr
+			}
+		}
+		health = m.health
+	}()
```
DmytroPI-dev, line 686: Are we using defer() instead of explicit error handling?

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1129 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+
+	if gateHealth, blocked := m.runtimeGateHealth(); blocked {
+		m.health = gateHealth
+		return m.health, nil
+	}
+
+	syncManagedRolesStatusFromCNPG(m.cluster, m.runtime.Cluster())
+	status := m.cluster.Status.ManagedRolesStatus
+	if status == nil {
+		m.health.State = pgcConstants.Failed
```
DmytroPI-dev, line 696: `Failed` + `RequeueAfter` is the contradiction here, I guess. `syncManagedRolesStatusFromCNPG` just ran above this and always sets a non-nil ManagedRolesStatus — so this nil branch can only be hit if syncManagedRolesStatusFromCNPG itself received a nil cnpgCluster (guarded by the early return inside it). In practice it's a transient "CNPG hasn't populated its status yet" case, which should be `Provisioning`, not `Failed`, WDYT?

M4KIF: So in theory It's a dead case, as It would have to have cnpgCluster == nil and above It is such a condition:
image
But, as a double caution, I will change the Failed to provisioning as It indeed seems more transient.

M4KIF: You know what, I would leave It at failed. As we seem guarded enough against cnpg not healthy etc. 
And having a generic failure at nil check seems more like It for me. 
If we ever execute this code (status == nil ) because of cnpg not being healthy, that's a sideeffect and a signal something was missed by us.

pkg/postgresql/cluster/core/cluster.go
M4KIF: wdym error handling?
Deffer here is to supply status write at every possible branch exit of the Converge function. It's more of a decorator pattern, ie. the method does what It wants, and we like to ensure the status write happens no matter what.

pkg/postgresql/cluster/core/types.go
M4KIF: great catch, It's true. And It should be, as I was using magic values instead of those two. Placed them back into logic.

pkg/postgresql/cluster/core/types.go
M4KIF: again, a big thing. Reverted to using them, no magic values left. Thank you!

pkg/postgresql/cluster/core/cluster.go
M4KIF: this block got changes after this comment, but your idea is good. Thanks.

DmytroPI-dev:
@M4KIF , `reasonCNPGClusterNotHealthy  conditionReasons = "CNPGClusterNotHealthy"` (line 124 in types.go can be removed

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1131 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+	_, err = controllerutil.CreateOrUpdate(ctx, c.client, cm, func() error {
+		cm.Data = desiredCM.Data
+		cm.Annotations = desiredCM.Annotations
+		cm.Labels = desiredCM.Labels
+		if !metav1.IsControlledBy(cm, c.cluster) {
+			if setErr := ctrl.SetControllerReference(c.cluster, cm, c.scheme); setErr != nil {
+				return fmt.Errorf("setting controller reference: %w", setErr)
+			}
+		}
+		return nil
```
DmytroPI-dev: The previous version emitted EventConfigMapReady on create and update. That's now gone — a configmap drift correction or first-time creation happens with no event. Converge only emits when the key check passes on the next reconcile. The gap means ops has no signal that a configmap was actually written.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1131 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+		return p.health, nil
+	}
+
+	p.cluster.Status.ConnectionPoolerStatus = &enterprisev4.ConnectionPoolerStatus{Enabled: true}
+	p.health.State = pgcConstants.Ready
+	p.health.Reason = reasonAllInstancesReady
+	p.health.Message = msgPoolersReady
+	p.health.Phase = readyClusterPhase
+	p.health.Result = ctrl.Result{}
+	p.events.emitPoolerReadyTransition(p.cluster, p.cluster.Status.Conditions)
```
DmytroPI-dev: `p.cluster.Status.Conditions` is read before the deferred `writeComponentStatus` has persisted poolerReady=True. The transition detection inside `emitPoolerReadyTransition` may fire against a stale slice. Capture `oldConditions` before the ready assignment if the intent is to detect the False→True edge.

DmytroPI-dev (approved):

pkg/postgresql/cluster/core/cluster.go
M4KIF: I suggest to fix this with a sort of "contracts as tests PR" to have the behavior locked in. It seems final and It constantly get's missed by me as I don't have the hands on results available via tests. 
It would be TDD, ie. propose tests with expected events per n reconciles/n behaviours (drift correction)/unhealthy transitions etc. -> align the code one final time -> make tests work and lock in the contract.
Now I'm constantly hitting and missing while shifting code with different intentions.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1131 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+				componentLogger.Error(err, "Component actuation transient error, requeueing", "step", "actuate")
+				return transientResult(err), nil
+			}
+			componentLogger.Error(err, "Component actuation failed", "step", "actuate")
+			return ctrl.Result{}, fmt.Errorf("%s actuate: %w", component.Name(), err)
+		}
+		componentLogger.Info("Component actuation completed", "step", "actuate")
+
+		health, err := component.Converge(ctx)
+		if err != nil && isTransientError(err) {
```
mploski, line 257: TransientError is not necessary anymore since we handle it on the Reconcile level at least for isconflict? Also isconflict shouldnt be treat the same way as toomany request the first can be retries automatically the second should go with backoff. We can clear it in the next PR though

M4KIF: so, collecting from the review in general ->
- tests for event emission to have them locked in without the possibility to cause side-effects with source changes any more
- event emission final alignment
- rollback the isTransient to simple isConflict(?)
- p.cluster.Status.Conditions is read before the deferred writeComponentStatus -> rethink the few loc so they can't cause undefined state in k8s(?)

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1131 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
-		rc.emitNormal(postgresCluster, EventSecretReady, fmt.Sprintf("Superuser secret %s created", postgresSecretName))
-		logger.Info("Superuser secret ref persisted to status")
-	}
 
-	// Re-attach ownerRef if it was stripped (e.g. by a Retain-policy deletion of a previous cluster).
-	hasOwnerRef, ownerRefErr := controllerutil.HasOwnerReference(secret.GetOwnerReferences(), postgresCluster, rc.Scheme)
-	if ownerRefErr != nil {
-		logger.Error(ownerRefErr, "Failed to check owner reference on Secret")
-		return ctrl.Result{}, fmt.Errorf("failed to check owner reference on secret: %w", ownerRefErr)
+		if err := component.Actuate(ctx); err != nil {
```
mploski: One potential gap. What if actuate logic fails? For example, lets assume CNPG is healthy and reconcileManagedRoles fails. Then we return early here with error, but failure state you set in health is never persisted. Is it on purpose?

pkg/postgresql/cluster/core/cluster.go
mploski: I agree with Dmytro here. This branch is dead, because status cannot be nil. i.e postgresCluster cannot be nil as this is why we are called in first place :-) cnpgCluster cannot be nil since we passed the clusterModel. 

Even if we constructed a hypothetical where nil was reached, the branch sets m.health and returns an error — but the  defer writeComponentStatus would then call c.Status().Update(ctx, cluster) on a nil cluster, which would panic. It's not defensive code IMO but it is a bit misleading code that implies a real production scenario.

pkg/postgresql/cluster/core/cluster.go
```diff
@@ -130,302 +159,1131 @@ func PostgresClusterService(ctx context.Context, rc *ReconcileContext, req ctrl.
...
+			if err != nil {
+				err = errors.Join(err, statusErr)
+			} else {
+				err = statusErr
+			}
+		}
+		health = c.health
+	}()
+
+	if c.runtime == nil || !c.runtime.IsHealthy() {
```
mploski, line 1147: Im a bit confused by this check as it seems again to be a dead code and in general about the Converge responsibility. IMO the health check for cnpg/cluster is done elsewhere, we should not duplicate this in next steps who depends on previous one. Otherwise we mix responsibilities of prerequisiteCheck and Converge. Converge should be responsible only for its resource state. prerequisite for dependencies. But there is a arch tension here i.e:
We would need to remove running Converge on gate not allowed and probably run directly status via writeComponentStatus(updateStatus, gate.Health) which also breaks boundry who can write. We need to pick one explicitly, currently the split is blured so we need to make sure we know what we are doing for future us

After second thought I started even questioning we need EvaluatePrerequisites everywhere. What problem it actually solve that the pipeline structure doesn't? the pipeline loop ordering is the prerequisite mechanism to the great exntent

pkg/postgresql/cluster/core/cluster_unit_test.go
```diff
@@ -1206,3 +1259,774 @@ func TestCreateOrUpdateConnectionPoolers(t *testing.T) {
...
+		},
+	}
+
+	exampleCm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pg1-config",
+			Namespace: "default",
+		},
+		Data: map[string]string{
+			"CLUSTER_RW_ENDPOINT":   "pg1-rw.default",
```
mploski: I guess we are missing CLUSTER_R_ENDPOINT?

pkg/postgresql/cluster/core/cluster.go
M4KIF: then writeComponentStatus doesn't fire and as far as I understand, k8s state is left "hanging" without an update and just with a plain/error requeue, right?

fe.
`func (m *managedRolesModel) Actuate(ctx context.Context) error {
	if rolesErr := reconcileManagedRoles(ctx, m.client, m.cluster, m.runtime.Cluster()); rolesErr != nil {
		m.events.emitWarning(m.cluster, EventManagedRolesFailed, fmt.Sprintf("Failed to reconcile managed roles: %v", rolesErr))
		m.health.State = pgcConstants.Failed
		m.health.Reason = reasonManagedRolesFailed`
So It would be best to update status everytime, before the reconcile requeue, right?

pkg/postgresql/cluster/core/cluster.go
M4KIF: I agree, the boundary seems blurred right now as well as the responsibility inside the models isn't exactly their responsibility to be fair.
In general, I would like to adjust It to the "phase locked until healthy 250% sure", ie. if we proceed further then It is a given that we don't have to worry anymore about cluster having unhealthy state 3 components later. That's a smell to be fair.
I wouldn't necessarily want to do this in this PR, as It will become even fatter and even more side-effect packed. It lacks precission rn. But If It needs to be this way I can.

pkg/postgresql/cluster/core/cluster.go
mploski: if we are raising the error I believe we should do this yes.

pkg/postgresql/cluster/core/cluster_unit_test.go
M4KIF: You're right, nice catch. Went back into the code, not into the tests. Pushing this rn

DmytroPI-dev: `configKeyClusterREndpoint   string = "CLUSTER_R_ENDPOINT"` - in correspondence with [this](https://github.com/splunk/splunk-operator/pull/1814#discussion_r3099954649), IDE says, it is not used.

pkg/postgresql/cluster/core/cluster.go
mploski, line 695: we should probably not override health otherwise we loose failure path from Actuate. This seems good on clusterModel, and override everywhere else. Shouldnt our tests catch it?

M4KIF: yup, a very serious error tbf. I'll write a test for such failure cases and align.

limak9182 (approved):

PR #1816, opened 2026-04-03
fix: propagate role removal to postgres cluster when databases are removed from postgres database spec

mploski:
### Description

Fixes a bug where removing a database from postgres database CR spec.databases (or deleting the CR with DeletionPolicy: Delete) did not cause the corresponding PostgreSQL roles to be dropped in postgres cluster. postgres cluster only drops a role from cnpg when it receives Exists: false in the cluster spec — previously, removed roles were never marked absent  but silently wiped out              
                                   
### Key Changes

  - api/v4/postgrescluster_types.go — removed omitempty from ManagedRole.Exists                                           
  - pkg/postgresql/database/core/database.go:                                                                           
    - cleanupManagedRoles now sends retained + deleted roles together in one SSA patch                                    
  - internal/controller/postgresdatabase_controller_test.go — added integration test for DB removal while CR stays alive; 
  extracted expectManagedRoleExists helper used across both role lifecycle tests 
### Testing and Verification

Added unit and integration tests. All green

### Related Issues

### PR Checklist

- [ ] All tests pass locally.

### DIFF
--- a/api/v4/postgrescluster_types.go
+++ b/api/v4/postgrescluster_types.go
@@ -36,7 +36,6 @@ type ManagedRole struct {
 
 	// Exists controls whether the role should be present (true) or absent (false) in PostgreSQL.
 	// +kubebuilder:default=true
-	// +optional
 	Exists bool `json:"exists"`
 }
 
--- a/api/v4/postgrescluster_types.go
+++ b/api/v4/postgrescluster_types.go
@@ -37,7 +37,7 @@ type ManagedRole struct {
 	// Exists controls whether the role should be present (true) or absent (false) in PostgreSQL.
 	// +kubebuilder:default=true
 	// +optional
-	Exists bool `json:"exists,omitempty"`
+	Exists bool `json:"exists"`
 }
 
 // PostgresClusterSpec defines the desired state of PostgresCluster.
--- a/internal/controller/postgresdatabase_controller_test.go
+++ b/internal/controller/postgresdatabase_controller_test.go
@@ -309,6 +309,15 @@ func seedOwnedDatabaseArtifacts(ctx context.Context, namespace, resourceName, cl
 	}
 }
 
+func expectManagedRoleExists(cluster *enterprisev4.PostgresCluster, roleName string, exists bool) {
+	rolesByName := make(map[string]enterprisev4.ManagedRole, len(cluster.Spec.ManagedRoles))
+	for _, r := range cluster.Spec.ManagedRoles {
+		rolesByName[r.Name] = r
+	}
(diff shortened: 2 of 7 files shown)


api/v4/postgrescluster_types.go
```diff
@@ -37,7 +37,7 @@ type ManagedRole struct {
 	// Exists controls whether the role should be present (true) or absent (false) in PostgreSQL.
 	// +kubebuilder:default=true
 	// +optional
-	Exists bool `json:"exists,omitempty"`
+	Exists bool `json:"exists"`
```
patrykw-splunk, line 39: if it's optional it should have `omitempty` and should be a pointer type

M4KIF: btw. is "exists" tag our custom or added by k8s api?

internal/controller/postgresdatabase_controller_test.go
```diff
@@ -493,6 +502,59 @@ var _ = Describe("PostgresDatabase Controller", func() {
...
+			postgresDB.Spec.Databases = []enterprisev4.DatabaseDefinition{{Name: "keepdb"}}
+			Expect(k8sClient.Update(ctx, postgresDB)).To(Succeed())
+
+			result, err := reconcilePostgresDatabase(ctx, requestName)
+			expectReconcileResult(result, err, 15*time.Second)
+
+			updatedCluster := &enterprisev4.PostgresCluster{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: clusterName, Namespace: namespace}, updatedCluster)).To(Succeed())
+
+			expectManagedRoleExists(updatedCluster, "keepdb_admin", true)
```
M4KIF: Here, and at line 530-533 the values could get packed into constants at the top of the file, so they are easy to change and read. Ie. those permisions are quite verbose, but by doing them with constants they can carry more meaning and less manual string pasting.

api/v4/postgrescluster_types.go
mploski: It is not optional anymore, need to remove optional flag - optionality causing problems with SSA if not set properly. @M4KIF this is our attribute so we know if we should add or remove provided role

internal/controller/postgresdatabase_controller_test.go
mploski: good point, lets do this

limak9182 (approved):

DmytroPI-dev (approved):

internal/controller/postgresdatabase_controller_test.go
mploski: done

PR #1817, opened 2026-04-03
fix: handle app download dir stat errors

vivekr-splunk:
## Summary
- return non-ErrNotExist path access errors from createAppDownloadDir
- make TestCreateAppDownloadDir use a portable ENOTDIR case instead of relying on root path permissions

## Why
GitLab CI runs the unit tests inside a root-capable container, so the old test assumption that /xyzzz.txt creation must fail is not portable. The function also swallowed non-ErrNotExist stat errors, which the portable test exposes.

## Validation
- `go test ./pkg/splunk/enterprise -run TestCreateAppDownloadDir -count=1`

### DIFF
--- a/pkg/splunk/enterprise/util.go
+++ b/pkg/splunk/enterprise/util.go
@@ -543,21 +543,12 @@ func getBundlePushState(afwPipeline *AppInstallPipeline) enterpriseApi.BundlePus
 }
 
 // createAppDownloadDir creates the app download directory on the operator pod
-func createAppDownloadDir(ctx context.Context, path string) error {
-	reqLogger := log.FromContext(ctx)
-	scopedLog := reqLogger.WithName("createAppDownloadDir").WithValues("path", path)
+func createAppDownloadDir(_ context.Context, path string) error {
 	_, err := os.Stat(path)
 	if errors.Is(err, os.ErrNotExist) {
-		errDir := os.MkdirAll(path, 0700)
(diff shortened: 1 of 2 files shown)


rlieberman-splunk (approved):

pkg/splunk/enterprise/util.go
```diff
@@ -553,6 +553,9 @@ func createAppDownloadDir(ctx context.Context, path string) error {
 			scopedLog.Error(errDir, "Unable to create directory at path")
 			return errDir
 		}
+	} else if err != nil {
+		scopedLog.Error(err, "Unable to access path")
```
kubabuczak: [nit] Why log it here? Shouldn't the calling function log it?

pkg/splunk/enterprise/util.go
```diff
@@ -553,6 +553,9 @@ func createAppDownloadDir(ctx context.Context, path string) error {
...
+		return err
 	}
 	return nil
 }
```
kubabuczak: [nit] Wouldn't this be better?
```go
_, err := os.Stat(path)
if errors.Is(err, os.ErrNotExist) {
	return os.MkdirAll(path, 0700)
}
return err
```

kubabuczak: And just leave logging to caller

kubabuczak (approved):

PR #1822, opened 2026-04-07
Feature/cspl 4378 test timeouts

kasiakoziol:
### Description

It sets timeouts for smoke and integration tests on test case, test suite and pipeline level.

### Key Changes

Test case:
- short: 15 minutes
- medium: 45 minutes
- long: 100 minutes

Test suite:
- short: 30 minutes
- medium: 90 minutes
- long: 200 minutes

Pipeline: 240 minutes

### Testing and Verification

Automated tests

### Related Issues

Jira: https://splunk.atlassian.net/browse/CSPL-4378

### PR Checklist

- [x] All tests pass locally.


.github/workflows/arm-AL2023-build-test-push-workflow-AL2023.yml
```diff
@@ -260,9 +261,7 @@ jobs:
           cp /snap/bin/kustomize ./bin/kustomize
       - name: Run smoke test
         id: smoketest
-        timeout-minutes: 240
         env:
-          TEST_TIMEOUT: 225m
           TEST_S3_ACCESS_KEY_ID: ${{ vars.TEST_S3_ACCESS_KEY_ID }}
```
kubabuczak, line 266: In general this is a bad idea. If we end test from runner then tests won't produce reports and other artefacts. Therefore timeout passed to test executor should be smaller from the timeout passed to runner. Runner timeout should in general never expire - it should be set just in case and should be significantly higher than value passed to test execution engine

kubabuczak: Also IIRC - Vivek asked us not to update workflows

Igor-splunk (approved):

test/appframework_aws/c3/appframework_aws_test.go
```diff
@@ -50,9 +49,7 @@ var _ = Describe("c3appfw test", func() {
 	var uploadedApps []string
 	var filePresentOnOperator bool
 
-	ctx := context.TODO()
-
-	BeforeEach(func() {
+	BeforeEach(NodeTimeout(testenv.TeardownTimeout), func(ctx SpecContext) {
```
kubabuczak: This is a timeout only for `BeforeEach` node? Why is it called `Teardown`?

.github/workflows/arm-AL2023-build-test-push-workflow-AL2023.yml
kasiakoziol: @kubabuczak So you suggest to as of now leave workflows unchanged and only manipulate timeouts on the test case/test suite level directly?

test/testenv/testcaseenv.go
```diff
@@ -228,16 +228,17 @@ func (testenv *TestCaseEnv) createNamespace() error {
 
 	// Cleanup the namespace when we teardown this testenv
 	testenv.pushCleanupFunc(func() error {
-		cleanupCtx := context.Background()
+		cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), TeardownTimeout-2*time.Minute)
```
kubabuczak: Why -2 minutes?

test/testenv/testenv.go
```diff
@@ -49,7 +49,7 @@ const (
 	defaultSplunkImage   = "splunk/splunk:latest"
 
 	// defaultTestTimeout is the max timeout in seconds before async test failed.
-	defaultTestTimeout = 1000000
+	defaultTestTimeout = 5400
```
kubabuczak: Shouldn't this be defined in `timeouts.go`?

test/testenv/testenv.go
```diff
@@ -64,7 +64,7 @@ const (
 	ConsistentDuration = 2000 * time.Millisecond
 
 	// DefaultTimeout is the max timeout before we failed.
-	DefaultTimeout = 2000 * time.Minute
+	DefaultTimeout = 90 * time.Minute
```
kubabuczak: Shouldn't this be defined in timeouts.go?

.github/workflows/arm-AL2023-build-test-push-workflow-AL2023.yml
kubabuczak: correct

patrykw-splunk (approved):

test/appframework_aws/c3/appframework_aws_test.go
kasiakoziol: @kubabuczak  It is the same timeout as for AfterEach execution, but I agree that this might be misleading, so let me either rename it to make it consistent or divide into two separate timeouts.

test/testenv/testenv.go
kubabuczak: Also - is't it too agresive?

test/testenv/testcaseenv.go
kasiakoziol: @kubabuczak It's a grace period so the code can fail cleanly instead of being interrupted. Without the buffer, the context deadline and the Ginkgo node timeout would race, and Ginkgo's hard kill wouldn't give the cleanup function a chance to return a meaningful error or run deferred statements.

test/testenv/deployment.go
kubabuczak: Shouldn't we migrate those too?

kubabuczak: There are also some in `testcasenv.go`

test/testenv/testenv.go
kasiakoziol: @kubabuczak Let me move it

test/testenv/testcaseenv.go
kubabuczak: Maybe it's better to use percentage? If somebody reduces `TeardownTimeout` this place may be missed

test/testenv/testenv.go
kasiakoziol: @kubabuczak It is used to wait for ns creation, operator deployment readiness and CR existence after create, so I guess we should even decrease it and will do this as a part of the commit to address your comments.

test/testenv/testcaseenv.go
kasiakoziol: @kubabuczak Agree, will do this.

test/testenv/deployment.go
kasiakoziol: @kubabuczak Is it a comment about the timeouts work or sth else? Could you please give me more context on what you mean?

kubabuczak: You migrated `PollImmediate` to `PollUntilContextTimeout` to make sure that context deadline is respected. Here we have context available, therefore we should also use new non-deprecated funciton

kasiakoziol: @kubabuczak Great, thank you. Now, it's clear. Let me address it now.

kubabuczak (approved):

PR #1827, opened 2026-04-08
Avoid monitoring console restarts for peer updates

vivekr-splunk:
## Summary
- update monitoring console reconciliation to sync peer changes at runtime instead of forcing pod restarts on MC config changes
- keep MC integration checks polling on actual config convergence and add explicit no-reset assertions for monitoring console pods
- extend the shared no-pod-reset helper so appframework suites also catch unexpected monitoring console pod resets

## Testing
- env GOCACHE=/tmp/splunk-operator-go-cache go test ./pkg/splunk/client
- env GOCACHE=/tmp/splunk-operator-go-cache go test ./pkg/splunk/enterprise -run 'TestApplyMonitoringConsole|TestGetMonitoringConsoleDesiredPeers|TestApplyMonitoringConsoleEnvConfigMap|TestGetMonitoringConsoleStatefulSet|TestAddMonitoringConsolePeer|TestRemoveMonitoringConsolePeer'
- env GOCACHE=/tmp/splunk-operator-go-cache go test -run '^$' ./test/...

## Notes
- a one-time MC StatefulSet rollout may still happen when this operator version is first deployed because the pod template stops carrying the old configmap-resourceVersion churn annotation; subsequent peer changes should not restart MC

### DIFF
--- a/pkg/splunk/client/enterprise.go
+++ b/pkg/splunk/client/enterprise.go
@@ -22,7 +22,9 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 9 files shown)


rlieberman-splunk:
It looks like the unit tests are failing currently with these changes.

PR #1835, opened 2026-04-09
CSPL-4621: add newcomer-friendly integration test writing guide

kubabuczak:
### Description

Adds a comprehensive integration test writing guide for newcomers, improves test execution UX, and fixes a Makefile bug.

### Key Changes

- New integration test guide covering test structure, writing, execution, and debugging
- Restructured setup instructions into sequential, copy-pasteable steps
- Fixed ginkgo command syntax (custom flags must come after `--`)
- Fixed `docker-buildx` Makefile target (broken builder teardown, cache preservation)
- Fixed operator log debugging example to point at cluster-wide operator namespace
- Added `tools/cleanup.sh` and `make undeploy` references to cleanup docs

### Related Issues

- [CSPL-4621](https://splunk.atlassian.net/browse/CSPL-4621)
- [CSPL-4694](https://splunk.atlassian.net/browse/CSPL-4694) (documented as limitation)

### PR Checklist

- [x] Code changes adhere to the project's coding standards
- [x] Documentation has been updated accordingly
- [x] If test framework files were changed (`test/testenv/`, `test/run-tests.sh`, `test/env.sh`), `docs/IntegrationTesting.md` has been updated
- [x] The PR description follows the project's guidelines


docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+  OP -.->|reconciles CRs in| B
+```
+
+### Key Concepts
+
+**TestEnv** (`test/testenv/testenv.go`)
+- Created once per suite in `BeforeSuite`
+- Builds a controller-runtime manager with a cached Kubernetes client
+- Configures the client to work with Splunk CRD types (v3 and v4)
+- Does **not** create namespaces or deploy the operator — that happens per-spec
```
rlieberman-splunk: Can we clarify here that deploying the operator only happens per-spec if it is not deployed at the cluster-wide scope before running the test? I actually did not know that was a feature of the tests, but it's good to know! I know it is mentioned above, but it would be helpful to have here as well.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+title: Integration Testing
+parent: Reference
+nav_order: 6
+---
+
+# Integration & E2E Test Writing Guide
+
+This guide helps newcomers understand the Splunk Operator integration test framework, write new tests, execute them, and debug failures.
+
+> **Assumed:** You have a Kubernetes cluster with the Splunk Operator deployed cluster-wide (the default). If you set `--cluster-wide=false`, the test framework will deploy a namespace-scoped operator per test case instead.
```
rlieberman-splunk: Please add that "If you set `--cluster-wide=false` in the ginkgo command to run the tests" so that it is not confused with the cluster-wide setting when deploying the operator separately. It also looks like the command only has a single `-` on cluster-wide, see https://github.com/splunk/splunk-operator/blob/main/test/trigger-tests.sh#L157.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+- Builds a controller-runtime manager with a cached Kubernetes client
+- Configures the client to work with Splunk CRD types (v3 and v4)
+- Does **not** create namespaces or deploy the operator — that happens per-spec
+
+**TestCaseEnv** (`test/testenv/testcaseenv.go`)
+- Created per `It` block via `testenv.NewDefaultTestCaseEnv(kubeClient, name)` followed by `testcaseEnvInst.NewDeployment(name)`
+- Creates a unique namespace and sets up all required resources:
+  - Namespace-scoped operator with RBAC (only when `--cluster-wide=false`; the default is `true`, which expects a pre-deployed cluster-wide operator)
+  - Cloud provider index secrets (EKS/Azure/GCP), created from environment variables
+  - License ConfigMap (only when `--license-file` is provided; without it, Splunk instances use trial license)
```
rlieberman-splunk, line 129: ~~It might be worth adding that the trial license is housed in our team's AWS account in an s3 bucket, and that is how it is used by default. If other users are not going to use our AWS account, we need to specify where to put the license file.~~

I see where that is covered in the internal documentation. Please disregard this comment.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+        Suite->>TestCaseEnv: NewDefaultTestCaseEnv(kubeClient, name)
+        TestCaseEnv->>Cluster: Create namespace
+        opt --cluster-wide=false
+            TestCaseEnv->>Cluster: Deploy operator with RBAC
+        end
+        TestCaseEnv->>Cluster: Create cloud secrets / license
+        TestCaseEnv->>Cluster: Validate prerequisites
+        Suite->>Deployment: testcaseEnvInst.NewDeployment(name)
+
+        Note over Deployment,Cluster: It: your test logic
```
rlieberman-splunk: Is it possible to add the 3 lines below in a dotted line box to show that they are all related to "your test logic"?

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+        if deployment != nil {
+            deployment.Teardown()
+        }
+        if testcaseEnvInst != nil {
+            Expect(testcaseEnvInst.Teardown()).ToNot(HaveOccurred())
+        }
+    })
+
+    Context("Standalone deployment (S1)", func() {
+        It("myfeature, integration, s1: can do something new", func() {
```
rlieberman-splunk: In case people copy and paste from here, can we update this to be something like `myfeature, mysuite, mytopology`, and then below you already explain what the options can actually be.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+
+**Build and push the operator image:**
+
+```bash
+# Single-platform build
+make docker-build IMG=<registry>/splunk-operator:latest
+make docker-push  IMG=<registry>/splunk-operator:latest
+
+# Multi-platform build (linux/amd64 + linux/arm64, pushes automatically)
+make docker-buildx IMG=<registry>/splunk-operator:latest
```
rlieberman-splunk: We can create single platform builds with `make docker-buildx` with the `PLATFORMS` variable. It is multi-platform by default but can be changed. https://github.com/splunk/splunk-operator/blob/main/Makefile#L176

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,627 @@
...
+```
+
+**Quick setup using `make` targets:**
+
+> Deploys to the cluster/context in your active kubeconfig (`~/.kube/config`).
+
+```bash
+make setup/ginkgo                    # Install Ginkgo v2 CLI + Gomega
+make deploy IMG=<registry>/splunk-operator:latest NAMESPACE=splunk-operator  # Deploy the operator
```
rlieberman-splunk: The `SPLUNK_GENERAL_TERMS` is needed here too, unless users want to manually update the value after deployment. Please do not include the actual value in this file, instead point to the README which includes all of the legal information about using the flag.

docs/IntegrationTesting.md
kubabuczak: I think that we should drop this feature - I believe that no one was aware of it and the operator is deployed from the code - another way of deployment to support. Currently we have a drift between this setup and other ways of deploying (e.g. terms are not passed)

kubabuczak: So I think that we can drop it from this guide

rlieberman-splunk: That sounds good to me!

docs/IntegrationTesting.md
kubabuczak: good catch. I'll actually remove information that we can deploy ~cluster~ operator per each test case. I think we should remove this feature. Alternatively we would need to rewrite it to helm or kustomize (to have less deployment ways to support)

docs/IntegrationTesting.md
kubabuczak: Makes sense. I will enclose it in `critical` box

docs/IntegrationTesting.md
kubabuczak: Another good catch - thank you! 

I have the plan to execute all commands from this guide tomorrow, but with such thorough review it may not be necessary 😁

.github/pull_request_template.md
```diff
@@ -19,5 +19,6 @@ _Jira tickets, GitHub issues, Support tickets..._
 - [ ] Code changes adhere to the project's coding standards.
 - [ ] Relevant unit and integration tests are included.
 - [ ] Documentation has been updated accordingly.
+- [ ] If test framework files were changed (`test/testenv/`, `test/run-tests.sh`, `test/env.sh`), `docs/IntegrationTesting.md` has been updated
```
kasiakoziol, line 22: env.sh is on the root level, not under test/ directory

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+"<mysuite>, <mytag>, <topology>: <human description>"
+```
+
+Examples:
+- `"smoke, basic, s1: can deploy a standalone instance"`
+- `"managercrcrud, integration, c3: can deploy Indexer and Search Head Cluster"`
+
+Tags used in CI filtering:
+- `smoke` — basic deployment checks (run on PRs)
+- `integration` — full integration tests (run on push to develop/main)
```
kasiakoziol: They can run on a PR if feature/ is added to a branch name

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+
+## How to Execute Tests
+
+### Prerequisites
+
+- Go installed (see `GO_VERSION` in `.env`)
+- A Kubernetes cluster with `kubectl` configured
+- Ginkgo v2 CLI — `make setup/ginkgo`
+- Operator and Splunk Enterprise images pushed to a registry your cluster can pull from (see below)
+- Operator deployed cluster-wide — `make deploy IMG=<image> NAMESPACE=splunk-operator`
```
kasiakoziol: I think we should mention Splunk General Terms here as well

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+> Deploys to the cluster/context in your active kubeconfig (`~/.kube/config`).
+
+```bash
+make setup/ginkgo                    # Install Ginkgo v2 CLI + Gomega
+make deploy IMG=<registry>/splunk-operator:latest NAMESPACE=splunk-operator SPLUNK_GENERAL_TERMS=<value>
+```
+
+> See [Splunk General Terms Acceptance](README.md#splunk-general-terms-acceptance) for the required `SPLUNK_GENERAL_TERMS` value
+
+### Run All Integration Tests via Makefile
```
kasiakoziol, line 499: I think it would be good to mention it here that they take a lot of time and are not all parallel, so it might be ineffective. Also, we do not run all of them on a single cluster, so it might exhaust the resources of a small cluster.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+| App Framework verification | `appframework_utils.go` |
+| Cloud storage (S3, Azure Blob, GCS) | `s3utils.go`, `azureutils.go`, `gcputils.go` |
+| Data ingestion and search | `ingest_utils.go`, `search_utils.go` |
+| Secrets and credentials | `secretutil.go` |
+
+---
+
+## How to Execute Tests
+
+### Prerequisites
```
kasiakoziol: There are also things like buckets for App FW tests, etc.

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+./test/trigger-tests.sh <operator-image> <enterprise-image>
+```
+
+### In CI (GitHub Actions)
+
+Tests run automatically on:
+- **PRs to main/develop:** Smoke tests (`build-test-push-workflow.yml`)
+- **Push to develop/main/feature branches:** Full integration tests (`int-test-workflow.yml`)
+- **Weekly schedule:** Nightly integration suite (`nightly-int-test-workflow.yml`)
+- **Manual trigger:** `manual-int-test-workflow.yml` with `workflow_dispatch`
```
kasiakoziol, line 569: I think this only applies to integration tests while smoke are excluded

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,668 @@
...
+| `TEST_TO_SKIP` / `SKIP_REGEX` | _(empty)_ | Regex to exclude tests |
+| `TEST_TIMEOUT` | `225m` | Ginkgo suite timeout |
+| `NUM_NODES` | `2` | Ginkgo parallel nodes |
+| `DEBUG` / `DEBUG_RUN` | `False` | If `True`, skip teardown on failure |
+
+### Cloud Provider Credentials
+
+Cloud provider variables below are used to create Kubernetes Secrets in each test namespace for SmartStore, App Framework, and index tests. In CI, these are populated from GitHub Actions secrets. For local runs, export them in your shell or source them from a `.env` file. If you're only running smoke tests without cloud storage features, these can be left unset.
+
+### AWS/EKS
```
kasiakoziol, line 742: I think it is important to mention just to be cautious not to push env.sh while trying to push another code because it will impact the remote runs as well as might disclose some confidential data such as secrets.

.github/pull_request_template.md
kubabuczak: `.env` is on the root level

kasiakoziol: Oh, yes, sorry, my bad.

kasiakoziol (approved):

docs/IntegrationTesting.md
```diff
@@ -0,0 +1,770 @@
...
+Don't test behaviors that belong to Splunk itself — it couples CI to Splunk internals and duplicates upstream coverage:
+
+- Search correctness and indexing pipeline internals
+- RF/SF replication health (reported by splunkd, not implemented by the operator)
+- App enablement and versioning inside Splunk
+- License enforcement and splunkd authentication
+
+### Gray zone
+
+Some existing tests use a Splunk-side signal as a proxy for operator correctness (e.g., `VerifyRFSFMet` after deploying an indexer cluster). This is acceptable when:
```
vivekr-splunk: Thanks for raising this. I think this case fits the "gray zone" rather than being purely out of scope. The goal of this E2E test is to verify that the CR status, emitted events, and logs accurately reflect the operand's actual runtime state. For that reason, I think it is reasonable for the test to make lightweight Splunk REST API or CLI checks, not as a replacement for operator-level assertions, but to confirm that what the operator reports matches the real Splunk behavior.

docs/IntegrationTesting.md
```diff
@@ -507,13 +605,32 @@ kubectl describe standalone -n <test-namespace>
 kubectl get events -n <test-namespace> --sort-by='.lastTimestamp'
 ```
 
+### Check for Leftovers After Tests
```
rlieberman-splunk, line 611: There is a cleanup script which is helpful for removing all of this in [tools/cleanup](https://github.com/splunk/splunk-operator/blob/main/tools/cleanup.sh).

rlieberman-splunk, line 611: It could also be good to add information on `make undeploy` to undeploy the current operator version once testing is complete. This will make sure the cluster is empty before deploying a new operator version in the future.

rlieberman-splunk (approved):

vivekr-splunk:
Thanks for putting this guide together. The documentation direction looks useful. I did notice one blocking issue in the code change mixed into this PR:

Changing `TestEnv` / `TestCaseEnv` namespaces to `sok-test-<name>` looks like it breaks existing paths that still treat `GetName()` as the namespace. For example:
- `test/testenv/testcaseenv.go` now sets `namespace: "sok-test-" + name`
- but `GetOperatorPodName()` and `DeleteOperatorPod()` still use `testcaseEnvInst.GetName()` as the namespace in non-cluster-wide mode (`test/testenv/util.go`)
- and existing tests do the same, e.g. `test/smoke/smoke_test.go` passes `testcaseEnvInst.GetName()` into `VerifyServiceAccountConfiguredOnPod()`, which shells out with `kubectl get pods -n <ns>` in `test/testenv/verificationutils.go`

Before this PR, `name == namespace`, so those call sites worked. With the new prefix, they seem to point at the wrong namespace. Could we either add/use a real `GetNamespace()` accessor and update the callers, or keep the externally used namespace semantics unchanged in this PR?

vivekr-splunk:
One small docs follow-up as well: in `docs/IntegrationTesting.md`, the debugging example currently says to inspect operator logs with `kubectl logs -n <test-namespace> deployment/splunk-op-<test-namespace>`. Earlier in the same guide, though, the assumed/default flow is a pre-deployed cluster-wide operator in the `splunk-operator` namespace. In that default case, the operator logs would come from `splunk-operator-controller-manager` in `splunk-operator`, not from each test namespace. Could we tweak that example so it matches the default execution model and avoids sending people to the wrong place while debugging?

kubabuczak:
> Thanks for putting this guide together. The documentation direction looks useful. I did notice one blocking issue in the code change mixed into this PR:
> 
> Changing `TestEnv` / `TestCaseEnv` namespaces to `sok-test-<name>` looks like it breaks existing paths that still treat `GetName()` as the namespace. For example:
> 
> * `test/testenv/testcaseenv.go` now sets `namespace: "sok-test-" + name`
> * but `GetOperatorPodName()` and `DeleteOperatorPod()` still use `testcaseEnvInst.GetName()` as the namespace in non-cluster-wide mode (`test/testenv/util.go`)
> * and existing tests do the same, e.g. `test/smoke/smoke_test.go` passes `testcaseEnvInst.GetName()` into `VerifyServiceAccountConfiguredOnPod()`, which shells out with `kubectl get pods -n <ns>` in `test/testenv/verificationutils.go`
> 
> Before this PR, `name == namespace`, so those call sites worked. With the new prefix, they seem to point at the wrong namespace. Could we either add/use a real `GetNamespace()` accessor and update the callers, or keep the externally used namespace semantics unchanged in this PR?

Oh, I did it to make kubectl commands easier - let's revert it. I'll address this issue (using GetName() instead of GetNamespace()) in separate PR

docs/IntegrationTesting.md
kubabuczak: The gray zone is specified here, since it's not testing code living in SOK, but I agree that it's reasonable to do it in our tests.

kubabuczak: Changed the wording here to be more clear on our policy

PR #1848, opened 2026-04-15
CSPL-4630: add feature gate infrastructure using k8s FeatureGate pattern

kubabuczak:
### Description

Add a feature gate registry using `k8s.io/component-base/featuregate` so new controllers can be gated behind `--feature-gates=<Gate>=true|false` with Alpha/Beta/GA lifecycle. Migrate the validation webhook toggle from the `ENABLE_VALIDATION_WEBHOOK` env var to a `ValidationWebhook` feature gate with backwards-compatible env var support.

### Key Changes

- Feature gate registry in `pkg/config/` with `ValidationWebhook` gate (Alpha, off by default) and legacy env var bridge
- Migrate `cmd/main.go` from `flag` to `pflag` and wire `--feature-gates` CLI flag
- Update webhook kustomize overlays to use container args instead of env var
- Add `docs/FeatureGates.md` with 4-step checklist for adding new gates
- Update `docs/ValidationWebhook.md` to reflect feature gate enablement

### Testing and Verification

- Unit tests cover gate registration, default-off behavior, and all three legacy env var paths (set to "true", unset, set to "false")
- Existing flags behave identically after the `flag` → `pflag` migration (1:1 rename)

### Related Issues

- [CSPL-4630](https://splunk.atlassian.net/browse/CSPL-4630)

### PR Checklist

- [x] Code changes adhere to the project's coding standards
- [x] Relevant unit and integration tests are included
- [x] Documentation has been updated accordingly
- [ ] If test framework files were changed (`test/testenv/`, `test/run-tests.sh`, `test/env.sh`), `docs/IntegrationTesting.md` has been updated
- [x] All tests pass locally
- [x] The PR description follows the project's guidelines

### DIFF
--- a/docs/FeatureGates.md
+++ b/docs/FeatureGates.md
@@ -10,14 +10,6 @@ Enable or disable feature gates at operator startup:
 /manager --feature-gates=ValidationWebhook=true
 ```
 
-Or via Helm values:
-
-```yaml
-splunkOperator:
-  featureGates:
-    ValidationWebhook: true
-```
-
 ## Maturity Lifecycle
 
 | Stage     | Default | Can Override | Next Step                               |
--- a/docs/FeatureGates.md
+++ b/docs/FeatureGates.md
@@ -26,7 +26,7 @@ Enable or disable feature gates at operator startup:
 
 ## Adding a New Feature Gate
 
-Follow these four steps:
+Follow these steps:
 
 ### 1. Register the gate in `pkg/config/featuregates.go`
 
--- a/docs/FeatureGates.md
+++ b/docs/FeatureGates.md
@@ -1,0 +1,118 @@
+# Feature Gates
+
+The Splunk Operator uses the Kubernetes [FeatureGate](https://pkg.go.dev/k8s.io/component-base/featuregate) pattern to control rollout of new functionality. Feature gates allow new code to be merged to the main branch without activating in production, giving teams a safe, per-environment opt-in mechanism.
+
+## Usage
+
+Enable or disable feature gates at operator startup:
+
+```bash
+/manager --feature-gates=ValidationWebhook=true
+```
+
+Or via Helm values:
+
+```yaml
+splunkOperator:
+  featureGates:
+    ValidationWebhook: true
+```
+
+## Maturity Lifecycle
+
+| Stage     | Default | Can Override | Next Step                               |
+|-----------|---------|-------------|-----------------------------------------|
+| **Alpha** | off     | Yes         | Promote to Beta after validation        |
+| **Beta**  | on      | Yes         | Promote to GA after sustained stability |
+| **GA**    | on      | No          | Remove gate in a future release         |
+
+## Current Feature Gates
+
+| Gate                  | Default | Stage | Since   | Description                                              |
+|-----------------------|---------|-------|---------|----------------------------------------------------------|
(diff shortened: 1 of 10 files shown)


kasiakoziol (approved):

docs/FeatureGates.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Current Feature Gates
+
+| Gate                  | Default | Stage | Since   | Description                                              |
+|-----------------------|---------|-------|---------|----------------------------------------------------------|
+| `ValidationWebhook`  | `false` | Alpha | v3.2.0  | Centralized validation webhook server for CR admission   |
+
+## Adding a New Feature Gate
+
+Follow these four steps:
```
rlieberman-splunk: It looks like there are only steps 1-2 and then some examples or optional steps for new controllers. Can we update this to reflect that?

kubabuczak: Good catch. Fixed!

rlieberman-splunk (approved):

docs/ValidationWebhook.md
```diff
@@ -85,23 +85,23 @@ If you prefer not to use cert-manager, you can provide your own TLS certificates
 
 #### Option 1: Use the Webhook-Enabled Kustomize Overlay
 
-Deploy using the `config/default-with-webhook` overlay which includes all necessary webhook components:
+Deploy using the `config/default-with-webhook` overlay which includes all necessary webhook components and enables the `ValidationWebhook` feature gate automatically:
 
 ```bash
 # Build and apply the webhook-enabled configuration
 kustomize build config/default-with-webhook | kubectl apply -f -
```
vivekr-splunk: This example still renders placeholder values like `WATCH_NAMESPACE_VALUE`, `SPLUNK_ENTERPRISE_IMAGE`, and `SPLUNK_GENERAL_TERMS_VALUE` directly into the manifest. Unlike the `config/default` flow, there is not a `make` target here doing the `sed` substitutions first, so `kustomize build config/default-with-webhook` is not actually ready to apply as-is. Can we either document the required replacements here or switch the example to a supported path that produces a fully rendered manifest?

docs/ValidationWebhook.md
```diff
@@ -290,7 +298,7 @@ kubectl get validatingwebhookconfiguration splunk-operator-validating-webhook-co
 
 ```bash
 kubectl logs -n splunk-operator deployment/splunk-operator-controller-manager | grep -i webhook
-# Look for: "Validation webhook enabled via ENABLE_VALIDATION_WEBHOOK=true"
+# Look for: "Validation webhook enabled via feature gate"
```
vivekr-splunk: The troubleshooting text here says to grep for `"Validation webhook enabled via feature gate"`, but `cmd/main.go` now logs `"Validation webhook enabled"`. Can we update one side or the other so users are not searching for a string that never appears?

vivekr-splunk (commented):
A few follow-up doc issues after the feature-gate refactor.

docs/ValidationWebhook.md
kubabuczak: While I agree with the premise I think it is outside of scope of this PR

docs/ValidationWebhook.md
kubabuczak: Good catch - will fix it

docs/ValidationWebhook.md
kubabuczak: I fixed it

patrykw-splunk (approved):

PR #1850, opened 2026-04-16
Fix CEL validation error when provider field is missing in VolumeSpec

gabrielm-splunk:
## Problem
After upgrading to v3.1.0, customers encountered validation errors when updating existing Standalone CRs that were created with previous versions:

```
The Standalone "standalone" is invalid: status.smartstore.volumes[0]: 
Invalid value: "object": no such key: provider evaluating rule: 
region is required when provider is aws
```

This error occurred because the CEL validation rule added in #1740 attempted to access `self.provider` without first checking if the field exists. In upgrade scenarios, status fields populated by older operator versions may not include the `provider` field, causing the validation to fail with "no such key: provider".

## Root Cause
The original CEL validation rule was:
```
self.provider != 'aws' || size(self.region) > 0
```

This rule evaluates `self.provider` even when the field doesn't exist, violating CEL's requirement to check field existence with `has()` before accessing optional fields.

## Solution
Updated the CEL validation rule in `api/v4/common_types.go` to:
```
!has(self.provider) || self.provider != 'aws' || (has(self.region) && size(self.region) > 0)
```

This change:
1. First checks if `provider` field exists with `!has(self.provider)`
2. If it doesn't exist, validation passes (allows backward compatibility)
3. Only validates the AWS region requirement when provider is explicitly set to 'aws'
4. Also checks `has(self.region)` before accessing the region field

## Testing
Reproduced the issue by:
1. Deployed v3.1.0 operator and CRDs to an EKS cluster
2. Created a Standalone CR with the buggy v3.1.0 CRDs
3. Attempted to patch the status with a volume object missing the `provider` field (simulating an upgrade scenario)
4. Confirmed the error: `The Standalone "standalone" is invalid: status.smartstore.volumes[0]: Invalid value: "object": no such key: provider evaluating rule: region is required when provider is aws`
5. Applied the fix (regenerated CRDs with the corrected validation rule)
6. Verified the same patch operation succeeds without errors

## Impact
- **Fixes**: Upgrade path from pre-3.1.0 versions where status fields were populated without the provider field
- **Maintains**: The validation requirement that AWS volumes must have a region
- **No impact**: On new deployments as they will include the provider field

## Checklist
- [x] Changes have been tested and validated
- [x] CRDs regenerated using `make manifests`
- [x] Commit includes Co-Authored-By for compliance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api/v4/common_types.go
+++ b/api/v4/common_types.go
@@ -327,7 +327,7 @@ type IndexConfDefaultsSpec struct {
 }
 
 // VolumeSpec defines remote volume config
-// +kubebuilder:validation:XValidation:rule="self.provider != 'aws' || size(self.region) > 0",message="region is required when provider is aws"
+// +kubebuilder:validation:XValidation:rule="!has(self.provider) || self.provider != 'aws' || (has(self.region) && size(self.region) > 0)",message="region is required when provider is aws"
 type VolumeSpec struct {
 	// Remote volume name
 	// +kubebuilder:validation:Required
--- a/config/crd/bases/enterprise.splunk.com_clustermanagers.yaml
+++ b/config/crd/bases/enterprise.splunk.com_clustermanagers.yaml
@@ -1164,7 +1164,8 @@ spec:
                       type: object
                       x-kubernetes-validations:
                       - message: region is required when provider is aws
-                        rule: self.provider != 'aws' || size(self.region) > 0
+                        rule: '!has(self.provider) || self.provider != ''aws'' ||
+                          (has(self.region) && size(self.region) > 0)'
                     type: array
                 type: object
               clusterManagerRef:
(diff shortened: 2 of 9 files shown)


docs/index.yaml
```diff
@@ -55,6 +55,34 @@ entries:
     urls:
     - https://splunk.github.io/splunk-operator/splunk-enterprise-3.1.0.tgz
     version: 3.1.0
+  - apiVersion: v2
```
kasiakoziol, line 58: Why do you try to push these changes as a part of multiple PRs?

gabrielm-splunk: @kasiakoziol Yeah not sure why the changes were showing up here as well. Older versions of these commits before they were merged in the other PR were still existing here, so just needed to rebase off of latest and they went away

rlieberman-splunk (approved):

PR #1858, opened 2026-04-20
Validate cross-referenced parameters between PostgresCluster and PostgresClusterClass

limak9182:
### Description

  - Adds admission-time cross-resource validation for PostgresCluster against its referenced PostgresClusterClass
  - Uses `mgr.GetAPIReader()` (live API read, not cached) to avoid false rejections during cache propagation

### Key Changes

  | Rule | Field | Condition |
  |------|-------|-----------|
  | Class existence | `spec.class` | Referenced PostgresClusterClass must exist |
  | Storage floor | `spec.storage` | Cannot be lower than class default |
  | Version floor | `spec.postgresVersion` | Major version cannot be lower than class default. If minor version specified in class cluster's minor cannot be lower. |
  | Pooler enablement | `spec.connectionPoolerEnabled` | Cannot be `true` when class has it disabled |
  | Merge completeness | `spec.instances`, `spec.postgresVersion`, `spec.storage` | After overlay, all three must be resolvable from cluster or class |

### Testing and Verification

  - unit test cases with fake `client.Reader` covering all rules and edge cases
  - Existing pg_hba validation tests updated to pass `nil` reader (no behavioral change)
  - `go build ./...` passes
  - `go test ./pkg/splunk/enterprise/validation/...` passes

### Related Issues

### PR Checklist

- [ ] All tests pass locally.


pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
```diff
@@ -557,3 +559,166 @@ func TestPostgresClusterClassPgHBAUpdateIntegration(t *testing.T) {
...
+			obj: &enterpriseApi.PostgresCluster{
+				TypeMeta:   metav1.TypeMeta{APIVersion: "enterprise.splunk.com/v4", Kind: "PostgresCluster"},
+				ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: "default"},
+				Spec:       enterpriseApi.PostgresClusterSpec{Class: "nonexistent"},
+			},
+			wantAllowed: false,
+			wantMessage: "PostgresClusterClass not found",
+		},
+		{
+			name: "rejected - storage below class floor",
```
DmytroPI-dev: `storage below class floor` - I assume we can override this value on a cluster lever even with lower ones, am I wrong?

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
```diff
@@ -557,3 +559,166 @@ func TestPostgresClusterClassPgHBAUpdateIntegration(t *testing.T) {
...
+				Spec: enterpriseApi.PostgresClusterSpec{
+					Class:   "prod",
+					Storage: ptrQuantity("10Gi"),
+				},
+			},
+			wantAllowed: false,
+			wantMessage: "storage cannot be lower than class default",
+		},
+		{
+			name: "rejected - version below class floor",
```
DmytroPI-dev, line 636: `version below class floor` - same as above - can't we set lower versions on cluster level now?

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
```diff
@@ -557,3 +559,166 @@ func TestPostgresClusterClassPgHBAUpdateIntegration(t *testing.T) {
...
+			obj: &enterpriseApi.PostgresCluster{
+				TypeMeta:   metav1.TypeMeta{APIVersion: "enterprise.splunk.com/v4", Kind: "PostgresCluster"},
+				ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: "default"},
+				Spec: enterpriseApi.PostgresClusterSpec{
+					Class:                   "prod",
+					ConnectionPoolerEnabled: ptrBool(true),
+				},
+			},
+			wantAllowed: false,
+			wantMessage: "connectionPoolerEnabled cannot be enabled",
```
DmytroPI-dev: "connectionPoolerEnabled cannot be enabled" - sounds strange (enabled cannot be enabled 😸 )

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
```diff
@@ -557,3 +559,166 @@ func TestPostgresClusterClassPgHBAUpdateIntegration(t *testing.T) {
...
+		},
+		{
+			name: "allowed - storage equal to class",
```
DmytroPI-dev: "allowed - storage equal to class" - and if higher?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+		allErrs = append(allErrs, validateAgainstClass(obj, reader)...)
+	}
+
 	return allErrs
 }
 
 // ValidatePostgresClusterUpdate validates a PostgresCluster on UPDATE.
-func ValidatePostgresClusterUpdate(obj, oldObj *enterpriseApi.PostgresCluster) field.ErrorList {
-	return ValidatePostgresClusterCreate(obj)
+func ValidatePostgresClusterUpdate(obj, oldObj *enterpriseApi.PostgresCluster, reader client.Reader) field.ErrorList {
```
DmytroPI-dev, line 54: `oldObj` is ignored on purpose?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("class"),
+			obj.Spec.Class,
+			"referenced PostgresClusterClass not found"))
+		return allErrs
+	}
+
+	classConfig := class.Spec.Config
+
+	mergedInstances := obj.Spec.Instances
```
DmytroPI-dev: do we need separate vars for each value? why not struct?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+	if classConfig == nil {
+		return allErrs
+	}
+
+	if obj.Spec.Storage != nil && classConfig.Storage != nil {
+		if obj.Spec.Storage.Cmp(*classConfig.Storage) < 0 {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec").Child("storage"),
+				obj.Spec.Storage.String(),
+				"storage cannot be lower than class default ("+classConfig.Storage.String()+")"))
```
DmytroPI-dev: As above: why we can't override storage on cluster level? I assume it is better to check whether it was changed/removed from spec later on and/or set to lower values. WDYT?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+		clusterMajor, clusterMinor := parseVersion(*obj.Spec.PostgresVersion)
+		classMajor, classMinor := parseVersion(*classConfig.PostgresVersion)
+		if clusterMajor > 0 && classMajor > 0 {
+			versionTooLow := clusterMajor < classMajor ||
+				(clusterMajor == classMajor && classMinor >= 0 && clusterMinor >= 0 && clusterMinor < classMinor)
+			if versionTooLow {
+				allErrs = append(allErrs, field.Invalid(
+					field.NewPath("spec").Child("postgresVersion"),
+					*obj.Spec.PostgresVersion,
+					"postgresVersion cannot be lower than class default ("+*classConfig.PostgresVersion+")"))
```
DmytroPI-dev: Same as for storage size, why can't we set lower version on cluster level? And not checking removal/downgrading later on?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+func parseVersion(version string) (major, minor int) {
+	for i, ch := range version {
+		if ch == '.' {
+			major, _ = strconv.Atoi(version[:i])
+			minor, _ = strconv.Atoi(version[i+1:])
+			return major, minor
+		}
+	}
+	major, _ = strconv.Atoi(version)
+	return major, -1
```
DmytroPI-dev, line 132: Do we need to check for minor at all? I mean, we are allowed to downgrade minor version, don't we?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
```diff
@@ -190,3 +196,344 @@ func TestGetPostgresClusterWarningsOnUpdate(t *testing.T) {
...
+	for _, obj := range objects {
+		b = b.WithRuntimeObjects(obj)
+	}
+	return b
+}
+
+func ptrBool(b bool) *bool       { return &b }
+func ptrString(s string) *string { return &s }
+func ptrInt32(i int32) *int32    { return &i }
+
```
DmytroPI-dev: why don't use `	"k8s.io/utils/ptr"` instead of custom methods?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
```diff
@@ -190,3 +196,344 @@ func TestGetPostgresClusterWarningsOnUpdate(t *testing.T) {
...
+			obj: &enterpriseApi.PostgresCluster{
+				Spec: enterpriseApi.PostgresClusterSpec{
+					Class:           "prod-pinned",
+					PostgresVersion: ptrString("17.5"),
+				},
+			},
+			wantErrCount: 0,
+		},
+		{
+			name:  "invalid - cluster minor lower than class minor",
```
DmytroPI-dev, line 359: We are allowed to downgrade minors, aren't we?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+			field.NewPath("spec").Child("class"),
+			obj.Spec.Class,
+			"referenced PostgresClusterClass not found"))
+		return allErrs
+	}
+
+	classConfig := class.Spec.Config
+
+	mergedInstances := obj.Spec.Instances
+	mergedVersion := obj.Spec.PostgresVersion
```
DmytroPI-dev: Also, we are re-inventing the wheel here 😸 , as we have class/cluster merge logic (instances, storage, postgresVersion, pooler defaults, et.) separately from the reconcile path. That’s easy to drift over time. I’d extract a shared helper if possible so admission and reconciliation enforce the same rules from one place.

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+		if classDisabled {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec").Child("connectionPoolerEnabled"),
+				true,
+				"connectionPoolerEnabled cannot be enabled when disabled in PostgresClusterClass"))
+		}
+	}
+
+	return allErrs
+}
```
DmytroPI-dev: This changes `connectionPoolerEnabled` from an override into a hard class policy. The API comments currently say the cluster value “takes precedence” / “can be overridden”, so rejecting `cluster=true` when the class is `false` looks like a semantic change that should either be documented explicitly or revisited.

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+	if mergedStorage == nil {
+		allErrs = append(allErrs, field.Required(specPath.Child("storage"),
+			"must be set in PostgresCluster or PostgresClusterClass"))
+	}
+
+	if classConfig == nil {
+		return allErrs
+	}
+
+	if obj.Spec.Storage != nil && classConfig.Storage != nil {
```
DmytroPI-dev: Same point for `storage` and `postgresVersion`: the merge code treats class values as defaults, but this webhook now treats them as minimum allowed values. If that policy shift is intentional, I’d update the API docs/comments because “overrides” reads differently today.

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
...
+	var allErrs field.ErrorList
+
+	class := &enterpriseApi.PostgresClusterClass{}
+	if err := reader.Get(context.Background(), client.ObjectKey{Name: obj.Spec.Class}, class); err != nil {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("class"),
+			obj.Spec.Class,
+			"referenced PostgresClusterClass not found"))
+		return allErrs
+	}
```
DmytroPI-dev, line 72: `reader.Get()` collapses every error into “referenced PostgresClusterClass not found”. That will hide real issues like API/RBAC failures and return a misleading validation error. I’d special-case IsNotFound and surface other errors separately.

pkg/splunk/enterprise/validation/validate.go
```diff
@@ -84,10 +85,10 @@ func ValidateWithClient(ar *admissionv1.AdmissionReview, validators map[schema.G
...
 	}
 
-	// Create validation context if client is available
+	// Create validation context if client or reader is available
 	var vc *ValidationContext
-	if k8sClient != nil {
-		vc = NewValidationContext(k8sClient, req.Namespace)
+	if k8sClient != nil || reader != nil {
+		vc = NewValidationContext(k8sClient, reader, req.Namespace)
 	}
```
DmytroPI-dev: Cross-resource validation now depends entirely on `vc.Reader.` If a caller provides only a cached client, or calls the validator directly, the checks silently disappear. I’d prefer a Reader -> Client fallback so the behavior is less wiring-sensitive.

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
```diff
@@ -190,3 +196,344 @@ func TestGetPostgresClusterWarningsOnUpdate(t *testing.T) {
...
+					Instances:       ptrInt32(1),
+					PostgresVersion: ptrString("17"),
+					Storage:         ptrQuantity("10Gi"),
+				},
+			},
+			wantErrCount: 0,
+		},
+	}
+
+	for _, tt := range tests {
```
DmytroPI-dev, line 514: The new tests cover the happy path well, but they mostly assert error count/field or Allowed. I’d add message assertions for the unit tests and one update-path integration case for the cross-resource checks, so semantic regressions in the comparison logic are easier to catch.

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
```diff
@@ -557,3 +559,166 @@ func TestPostgresClusterClassPgHBAUpdateIntegration(t *testing.T) {
...
+	s := runtime.NewScheme()
+	enterpriseApi.AddToScheme(s)
+	b := fake.NewClientBuilder().WithScheme(s)
+	for _, obj := range objects {
+		b = b.WithRuntimeObjects(obj)
+	}
+	return b
+}
+
+func TestCrossResourceValidationIntegration(t *testing.T) {
```
DmytroPI-dev, line 588: The new tests cover the happy path well, but they mostly assert error count/field or Allowed. I’d add message assertions for the unit tests and one update-path integration case for the cross-resource checks, so semantic regressions in the comparison logic are easier to catch.

cmd/main.go
```diff
@@ -348,6 +348,7 @@ func main() {
 			ReadTimeout:  readTimeout,
 			WriteTimeout: writeTimeout,
 			Client:       mgr.GetClient(),
+			Reader:       mgr.GetAPIReader(),
```
mploski: could you elab more, why we need this reader? What problem do we solve?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
```diff
@@ -36,12 +40,111 @@ func ValidatePostgresClusterCreate(obj *enterpriseApi.PostgresCluster) field.Err
 		}
 	}
 
+	if reader != nil {
```
mploski, line 46: could you elab why we expect some reader to validate against class logic?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
mploski: +1

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
mploski: I think there is no contradiction here i.e our default logic is there and it is used in the runtime reconciler but business rules validation is done here in the validation webhook. So we have "overrides" with guardrails in place. I agree we need to document this logic properly though

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
mploski: I agree with @DmytroPI-dev our idea here is not to say: if class has 1000GB's of disc our cluster cannot use 999. What we really want to say once we DEPLOYED cluster with 1000 we cannot change cluster config to 999

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
mploski: In this case we probably should enforce minimum version of postgres coming from class and not allow lowering it down from the compliance perspective

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
mploski: +1 from connection pooler perspective I think we SHOULD allow override it from false to positive BUT if user make it enabled in cluster we need to enforce that cnpg config for pooler is in place otherwise reject

pkg/splunk/enterprise/validation/context.go
```diff
@@ -27,9 +27,14 @@ import (
...
 // to access Kubernetes resources
 type ValidationContext struct {
-	// Client is the Kubernetes client for resource lookups
+	// Client is the Kubernetes client for resource lookups (cached)
 	Client client.Client
 
+	// Reader is a live API reader that bypasses the cache.
+	// Use this for cross-resource lookups where eventual consistency
+	// could cause false rejections.
+	Reader client.Reader
```
mploski: Is this really a problem in our case? if we for example deploy new cluster CR and we need to check latest deployed clusterClass isnt it already populated within the cache? Why we need to bypass it? Why this isnt a problem for other CR's checked by this?

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: make sense, in that case we can rely on our already implemented CEL validation:
```go
  // +kubebuilder:validation:XValidation:rule="!has(oldSelf.storage) || (has(self.storage) && quantity(self.storage).compareTo(quantity(oldSelf.storage)) >= 0)",messageExpression="..."
```

I removed this block. Now, I'm checking only if we have storage defined at least onece in Class or Cluster

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
limak9182: yes, correct

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
limak9182: For the compliance we prohibit setting lower version than in class

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
limak9182: fixed

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
limak9182: removed

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: yes, we don't care about oldObj

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: refactored

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: done

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: resolved, https://github.com/splunk/splunk-operator/pull/1858#discussion_r3130597351, downgrading while being alligned with version set in class is allowed.

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
limak9182: fixed

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
limak9182: downgrade is allowed, only setting lower than in class is banned

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: refactored

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: refactored

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: added comment for clarity

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: removed API reader

pkg/splunk/enterprise/validation/validate.go
limak9182: removed API reader

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation_test.go
limak9182: done

pkg/postgresql/cluster/adapter/webhook/postgres_webhook_integration_test.go
limak9182: done

cmd/main.go
limak9182: removed

pkg/postgresql/cluster/adapter/webhook/postgrescluster_validation.go
limak9182: removed

pkg/splunk/enterprise/validation/context.go
limak9182: removed

